PR #5, opened 2016-01-20
Vector-based JoinedTupleCollector for HashJoin

craig-chasseur:
This should hopefully fix one of the performance pain points for hash join.


pateljm:
@craig-chasseur Looks good so am merging. Also ran a quick test on my Mac with 10GB SSB and query 1 (in release build). The time with and without the join tuple collector was very close. So this is a good starting point and at least give us one more dimension to think about for join performance.

PR #6, opened 2016-01-20
SimpleScalarSeparateChainingHashTable for fast-path join & grouping

craig-chasseur:
This PR introduces `SimpleScalarSeparateChainingHashTable`, which is a special, simplified version of `SeparateChainingHashTable` for scalar (non-composite) keys that have a reversible hash function. It eliminates much of the complexity of key management (including the need to store a key separately at all) and hashing. Using the simplified version in hash-join and aggregation-grouping operations is expected to be faster than using the original `SeparateChainingHashTable` (although the original, along with `LinearOpenAddressingHashTable`, is still available as fallbacks for cases where we require their full generality, like string keys or composite keys).

Besides the introduction of the new `HashTable` class itself, this PR also introduces a couple of other changes that the new implementation depends on:
1. Hashing logic in `TypedValue` is changed so that we use our own trivially reversible hash function for simple scalar types instead of relying on std::hash.
2. Functionality in the base `HashTable` class that involves management of key storage has been moved to a new helper class `HashTableKeyManager`. The preexisting `HashTable` implementations require this functionality and automatically create a `HashTableKeyManager` object when they are created. The new `SimpleScalarSeparateChainingHashTable` doesn't require a `HashTableKeyManager` and thus doesn't create one.

There is probably some additional fast-path tuning that we can do for hash joins generally and for the new `SimpleScalarSeparateChainingHashTable` specifically, but this is a good first step towards optimizing the common case for joins and grouping operations with simple scalar keys.

### DIFF
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -1,6 +1,6 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
- *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2015-2016 Pivotal Software, Inc.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -497,6 +497,8 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
     }
   }
 
+  // Remember key types for call to SimplifyHashTableImplTypeProto() below.
+  std::vector<const Type*> key_types;
   for (std::vector<E::AttributeReferencePtr>::size_type attr_idx = 0;
        attr_idx < left_join_attributes.size();
        ++attr_idx) {
@@ -506,6 +508,7 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
       THROW_SQL_ERROR() << "Equality join predicate between two attributes of different types "
                            "is not allowed in HashJoin";
     }
+    key_types.push_back(&left_attribute_type);
   }
 
   // Choose the smaller table as the inner build table,
@@ -551,7 +554,9 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   S::HashTable *hash_table_proto = query_context_proto_->add_join_hash_tables();
 
   hash_table_proto->set_hash_table_impl_type(
-      HashTableImplTypeProtoFromString(FLAGS_join_hashtable_type));
+      SimplifyHashTableImplTypeProto(
+          HashTableImplTypeProtoFromString(FLAGS_join_hashtable_type),
+          key_types));
 
   const CatalogRelationSchema *build_relation = build_relation_info->relation;
   for (const attribute_id build_attribute : build_attribute_ids) {
@@ -1005,6 +1010,7 @@ void ExecutionGenerator::convertAggregate(
     }
   }
 
+  std::vector<const Type*> group_by_types;
   for (const E::NamedExpressionPtr &grouping_expression : physical_plan->grouping_expressions()) {
     unique_ptr<const Scalar> execution_group_by_expression;
     E::AliasPtr alias;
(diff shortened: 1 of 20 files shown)


query_optimizer/ExecutionGenerator.cpp
```diff
@@ -551,7 +554,9 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   S::HashTable *hash_table_proto = query_context_proto_->add_join_hash_tables();
 
   hash_table_proto->set_hash_table_impl_type(
-      HashTableImplTypeProtoFromString(FLAGS_join_hashtable_type));
+      SimplifyHashTableImplTypeProto(
```
pateljm: Can we add a comment here to clearly state when the simplified hash table is selected?

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -1029,8 +1036,12 @@ void ExecutionGenerator::convertAggregate(
...
   aggr_state_proto->set_estimated_num_entries(cost_model_->estimateCardinality(physical_plan));
 
-  aggr_state_proto->set_hash_table_impl_type(
-      HashTableImplTypeProtoFromString(FLAGS_aggregate_hashtable_type));
+  if (!group_by_types.empty()) {
+    aggr_state_proto->set_hash_table_impl_type(
+        SimplifyHashTableImplTypeProto(
+            HashTableImplTypeProtoFromString(FLAGS_aggregate_hashtable_type),
+            group_by_types));
+  }
```
pateljm: Same comments as above (a comment here would be very useful).

storage/SimpleScalarSeparateChainingHashTable.hpp
```diff
@@ -0,0 +1,1062 @@
...
+#include "glog/logging.h"
+
+namespace quickstep {
+
+/** \addtogroup Storage
+ *  @{
+ */
+
+/**
+ * @brief A lean simplified version of SeparateChainingHashTable that only
```
pateljm: A little more detail here about what "simple" means would be great to help the reader.

types/TypedValue.hpp
```diff
@@ -271,6 +287,30 @@ class TypedValue {
   }
 
   /**
+   * @brief Determine if the hash function getHash() is reversible for
```
pateljm: Can we expand the comments to indicate why reversible is important and how it is used?

storage/HashTable.hpp
```diff
@@ -127,14 +116,18 @@ class HashTable : public HashTableBase<resizable,
...
 
-  // A surrogate hash value for empty buckets. Keys which actually hash to this
-  // value should have their hashes mutated (e.g. by adding 1). We use zero,
-  // since we will often be using memory which is already zeroed-out and this
-  // saves us the trouble of a memset. This has some downside, as the std::hash
-  // function we use is the identity hash for integers, and the integer 0 is
-  // common in many data sets and must be adjusted (and will then spuriously
-  // collide with 1). Nevertheless, this expense is outweighed by no longer
-  // having to memset large regions of memory when initializing a HashTable.
+  // Some HashTable implementations (notably LinearOpenAddressingHashTable)
```
pateljm: Since this is a key "gateway" file for anyone looking at hashtables for operators, could we add a more detailed comment field here describing briefly the hash tables, and specifically the cases when the SimpleScalarSeparateChainingHashTable is used. Thanks!

pateljm:
Looks quite good @craig-chasseur. Thanks! A few comments to make it easier for someone to browse the code. 

Also, did you carry out any performance experiments (say with SSB Q1?). If yes, would be nice to document that here.

craig-chasseur:
I actually don't have a benchmarking setup for SSB. Could we get one of the
students who worked on benchmarking to write a wiki page with step-by-step
instructions for getting benchmark numbers for us? That would be quite
helpful.

pateljm:
@craig-chasseur @cramja has started a benchmark repo at https://github.com/UWQuickstep/benchmarks 

An easier way to run SSB is pull the files from: http://pages.cs.wisc.edu/~jignesh/data/. The tarball there (its large) has raw files for 1 and 10 SF SSB, and the SQL files.

craig-chasseur:
Hey @pateljm, added some explanatory comments as requested. Please take another look.

pateljm:
LTGM. Thanks! Merging now.

PR #7, opened 2016-01-21
Auto detect the hardware concurrency level, when possible, and use that.

pateljm:
Use std::thread::hardware_concurrency mechanism to detect the degree of hardware parallelism and use that as the default.

### DIFF
--- a/cli/QuickstepCli.cpp
+++ b/cli/QuickstepCli.cpp
@@ -118,10 +118,11 @@ static constexpr char kPathSeparator = '/';
 static constexpr char kDefaultStoragePath[] = "qsstor/";
 #endif
 
-DEFINE_int32(num_workers, 1, "Number of worker threads. The default is picked "
-                             "by examining the reported hardware concurrency "
-                             "level. If that can't be computed, "
-                             "then set to 1.");
+DEFINE_int32(num_workers, 0, "Number of worker threads. If this values is "
+                             "specified and is greater than 0, then this "
+                             "user-supplied value is used. Else (i.e. the"
+                             "default case), we examine the reported "
+                             "hardware concurrency level, and use that.");
 DEFINE_bool(preload_buffer_pool, false,
             "If true, pre-load all known blocks into buffer pool before "
             "accepting queries (should also set --buffer_pool_slots to be "
@@ -140,19 +141,22 @@ int main(int argc, char* argv[]) {
   gflags::ParseCommandLineFlags(&argc, &argv, true);
   google::InitGoogleLogging(argv[0]);
 
-  // Detect the hardware concurrency level.
-  unsigned int num_hw_threads = std::thread::hardware_concurrency();
+  // Detect the hardware concurrency level. Note this call will return 0
+  // if it fails (which it may on some machines/environments).
+  const unsigned int num_hw_threads = std::thread::hardware_concurrency();
 
-  // It is possible that the call above fails and returns 0, in which case
-  // the default should be 1.
-  quickstep::FLAGS_num_workers = (num_hw_threads == 0) ? 1 : num_hw_threads;
-  LOG(INFO) << "Number of threads set to "
-            << quickstep::FLAGS_num_workers
-            << "\n";
+  // Use the command-line value if that was supplied, else use the value
+  // that we computed above, provided it did return a valid value.
+  // TODO(jmp): May need to change this at some point to keep one thread
+  //            available for the OS if the hardware concurrency level is high.
+  const unsigned int real_num_workers = quickstep::FLAGS_num_workers != 0
+                                      ? quickstep::FLAGS_num_workers
+                                      : (num_hw_threads != 0 ?
+                                         num_hw_threads
+                                         : 1);
 
-  if (quickstep::FLAGS_num_workers > 0) {
-    printf("Starting Quickstep with %d worker thread(s)\n",
-           quickstep::FLAGS_num_workers);
+  if (real_num_workers > 0) {
+    printf("Starting Quickstep with %d worker thread(s)\n", real_num_workers);
   } else {
     LOG(FATAL) << "Quickstep needs at least one worker thread";
   }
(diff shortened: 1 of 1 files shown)


cli/QuickstepCli.cpp
```diff
@@ -135,6 +140,16 @@ int main(int argc, char* argv[]) {
   gflags::ParseCommandLineFlags(&argc, &argv, true);
   google::InitGoogleLogging(argv[0]);
 
+  // Detect the hardware concurrency level.
+  unsigned int num_hw_threads = std::thread::hardware_concurrency();
+
+  // It is possible that the call above fails and returns 0, in which case
+  // the default should be 1.
+  quickstep::FLAGS_num_workers = (num_hw_threads == 0) ? 1 : num_hw_threads;
```
craig-chasseur: This *always* overrides the user's choice for the number of threads, even if `std::thread::hardware_concurrency()` returns 0.

I think it makes sense to have a special (but still user-specifiable) value indicate that we should attempt to auto-detect the number of threads. That should also be the default for the command-line option. But, if the user explicitly specifies something else, that should take precedence over auto detection. So to summarize:

1. We change the default for `num_workers` from 1 to 0, and change the message to say that 0 (the default) means that we will attempt to auto-detect the number of threads.
2. If `FLAGS_num_workers` is nonzero, we use it as-is and do NOT attempt auto-detection.
3. If `FLAGS_num_workers` is zero, we call `std::thread::hardware_concurrency()` to attempt auto detection.
4. If `std::thread::hardware_concurrency()` returns 0, fallback on 1 thread as the default of last resort.

Or in code terms:
```
const unsigned real_num_workers
    = quickstep::FLAGS_num_workers != 0
        ? quickstep::FLAGS_num_workers
        : (std::thread::hardware_concurrency() != 0
            ? std::thread::hardware_concurrency()
            : 1);

File: cli/QuickstepCli.cpp [hunk 2]
```diff
@@ -135,6 +140,16 @@ int main(int argc, char* argv[]) {
...
+  LOG(INFO) << "Number of threads set to "
```
craig-chasseur: No need to log this here. Just a couple of lines below, we print out the number of worker threads on startup already.

craig-chasseur:
Hey @pateljm, just finished a review pass. Very cool idea, but there's a bug that needs fixing in the implementation. Should be a quick and easy change and then we can merge.

cli/QuickstepCli.cpp
pateljm: Will do!

cli/QuickstepCli.cpp
pateljm: Good point! Fixed.

cli/QuickstepCli.cpp
```diff
@@ -135,9 +141,22 @@ int main(int argc, char* argv[]) {
...
-    printf("Starting Quickstep with %d worker thread(s)\n",
-           quickstep::FLAGS_num_workers);
+  // Detect the hardware concurrency level. Note this call will return 0
+  // if it fails (which it may on some machines/environments).
+  const unsigned int num_hw_threads = std::thread::hardware_concurrency();
+
+  // Use the command-line value if that was supplied, else use the value
+  // that we computed above, provided it did return a valid value.
+  // TODO(jmp): May need to change this at some point to keep one thread
+  //            available for the OS if the hardware concurrency level is high.
```
zuyu: Do we need to reserve one hardware thread for `Foreman`? And one for `Shiftboss` in near future?

FYI, `Foreman` will be changed as a long-running service to process concurrent query executions, instead of current one-query-per-Foreman model.

cli/QuickstepCli.cpp
```diff
@@ -116,7 +118,11 @@ static constexpr char kPathSeparator = '/';
 static constexpr char kDefaultStoragePath[] = "qsstor/";
 #endif
 
-DEFINE_int32(num_workers, 1, "Number of worker threads");
+DEFINE_int32(num_workers, 0, "Number of worker threads. If this values is "
```
craig-chasseur: Grammar: should be singular "value" instead of plural "values".

cli/QuickstepCli.cpp
craig-chasseur: std::thread::hardware_concurrency() returns the number of (virtual) CPU
cores on the host machine, not the maximum number of concurrent threads.
Right now, Foreman's CPU use is much lower than that of workers, so it's
not that much of a problem to create an extra foreman thread that can be
freely scheduled on any CPU. In the future, if we wind up running many
concurrent queries, it might make sense to reserve one or two cores
exclusively for foremen threads, but I don't think that's an issue for now.

zuyu: Cool. So if a CPU supports simultaneous multithreading (SMT), the max number of concurrent threads would be 2x of the number of CPU cores returned by `std::thread::hardware_concurrency`?

craig-chasseur: No, it would actually be the number of SMT threads. So, on my laptop with
hyperthreading (4 cores, 8 hyperthreads), it returns 8.

Hardware SMT threads appear as physical CPU cores to the operating system
(sort of... Linux actually does detect which of the virtual "cores" are SMT
siblings and has some scheduling optimizations based on that... for
instance it assumes that the cost of migrating a thread to its sibling is
negligible, but that's not visible to userspace programs).

We should also be clear about what is meant by "maximum concurrency". Linux
and other modern OSes have no problem with scheduling thousands of threads
on a CPU core (although if the threads in question are high CPU use or
contend for other resources, you can expect the system to chug).
std::thread::hardware_concurrency() just gives us a hint about the actual
number of parallel CPU cores (or with SMT, thread contexts) *in hardware*.
Since our workers are mostly either CPU or memory bound, it makes good
sense to use that number of workers. Old performance experiments showed
that when the workload was mostly CPU bound with few cache misses, using
the number of CPU cores vs. doubling to use hyperthreads made almost no
difference in performance. On the other hand, when memory access was an
issue and cache misses were higher, we got a decent boost (10-30% IIRC)
from using the hyperthreads (mainly because while one of the virtual cores
in a pair was stalling during a cache miss, the other one could
almost-instantly take over the physical core and get some useful work done).

TLDR: std::thread::hardware_concurrency() is just a "hint" about
concurrency. It's certainly safe to go over what it says, but the hint is
actually quite good for our purposes, whether or not hyperthreading is
available.

pateljm: Zuyu - for the foreman, tmb, ... as we talked about we will likely end up with a common binary to deploy. The deployment framework can explicitly specify the number of workers if it needs to.

craig-chasseur:
LGTM. Merging.

PR #11, opened 2016-01-22
Blockproperties staging

cramja:
**Parser support for BLOCKPROPERTIES**
This PR contains parser support for the BLOCKPROPERTIES clause for the CREATE TABLE statement. Currently the parser recognizes the syntax:
```SQL
CREATE TABLE name (attr1 type, attr2 type ... ) WITH BLOCKPROPERTIES (TYPE columnstore, sort attr1, COMPRESS ALL); 
```
and several variants. Users can specify storage types, sort columns where supported, compressed columns where supported. Sort types will include the 5 currently supported types. COMPRESS may be followed with the keyword ALL or a comma list of attribute names enclosed in parentheses like COMPRESS (attr1, attr2)

I plan to make a separate pull request for the optimizer implementation.


zuyu:
@cramja Please sync with the upstream master.

parser/tests/Create.test
```diff
@@ -137,3 +137,38 @@ create table test (attr int) check(attr>0)
 ERROR: Table Constraints (CHECK) is not supported yet (1 : 30)
 create table test (attr int) check(attr>0)
                              ^
+==
+
+create table test (attr0 int, attr1 int) with blockproperties (type rowstore);
```
pateljm: Could we use the standard SQL conventions and capitalize all tokens? Thanks!

parser/ParseBlockProperties.cpp
```diff
@@ -0,0 +1,43 @@
+/**
+ *   Copyright 2011-2015 Quickstep Technologies LLC.
```
pateljm: Since this is a new file, we can drop these two old copyright lines.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,191 @@
+/**
+ *   Copyright 2011-2015 Quickstep Technologies LLC.
+ *   Copyright 2015 Pivotal Software, Inc.
```
pateljm: Same comment here for copyright as above.

storage/BasicColumnStoreTupleStorageSubBlock.cpp
```diff
@@ -173,45 +173,45 @@ BasicColumnStoreTupleStorageSubBlock::BasicColumnStoreTupleStorageSubBlock(
...
 }
 
-bool BasicColumnStoreTupleStorageSubBlock::DescriptionIsValid(
+int BasicColumnStoreTupleStorageSubBlock::DescriptionIsValid(
     const CatalogRelationSchema &relation,
     const TupleStorageSubBlockDescription &description) {
   // Make sure description is initialized and specifies BasicColumnStore.
   if (!description.IsInitialized()) {
-    return false;
+    return -1;
```
pateljm: Could we create an enum type and return that (within this class) and use that instead of hard-coded constants? Thanks!

storage/CompressedColumnStoreTupleStorageSubBlock.cpp
```diff
@@ -91,15 +91,15 @@ CompressedColumnStoreTupleStorageSubBlock::CompressedColumnStoreTupleStorageSubB
...
 }
 
-bool CompressedColumnStoreTupleStorageSubBlock::DescriptionIsValid(
+int CompressedColumnStoreTupleStorageSubBlock::DescriptionIsValid(
     const CatalogRelationSchema &relation,
     const TupleStorageSubBlockDescription &description) {
   // Make sure description is initialized and specifies CompressedColumnStore.
   if (!description.IsInitialized()) {
-    return false;
+    return -1;
```
pateljm: Same request as above to use enums.

storage/CompressedPackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -81,16 +81,16 @@ CompressedPackedRowStoreTupleStorageSubBlock::CompressedPackedRowStoreTupleStora
...
 
-bool CompressedPackedRowStoreTupleStorageSubBlock::DescriptionIsValid(
+int CompressedPackedRowStoreTupleStorageSubBlock::DescriptionIsValid(
     const CatalogRelationSchema &relation,
     const TupleStorageSubBlockDescription &description) {
   // Make sure description is initialized and specifies
   // CompressedPackedRowStore.
   if (!description.IsInitialized()) {
-    return false;
+    return -1;
```
pateljm: Again can we avoid using hard-coded constants? Thanks.

pateljm:
Excellent and sorely needed work. Thanks @cramja! 

See comments from the first round. @craig-chasseur may have more feedback. 

Also, would be nice if somewhere there was an easy to use "manual" that described what block property combinations are allowed, when, and with some examples. One way would be to create a large comment field somewhere in these files. ParseStatement.hpp seems like a good candidate right before the ParseStatementCreateTable function. I know the comments there will be non-uniform, but we are trying to be better about commenting the code.

storage/BasicColumnStoreTupleStorageSubBlock.cpp
zuyu: IMHO, we don't need to change the return value. Instead, for the debug purpose, we could print out the proto content:

```C++
DCHECK(BasicColumnStoreTupleStorageSubBlock::ProtoIsValid(relation_, description_))
    << "Attempted to create BasicColumnStoreTupleStorageSubBlock from an invalid proto description:\n"
    << proto.DebugString();
```

craig-chasseur:
The Travis build is showing many lint errors. Please fix those.

craig-chasseur:
There are also errors preventing the build from completing with either clang or GCC in Travis. Those need to be fixed, too.

.gitignore
```diff
@@ -8,3 +8,4 @@ Makefile.in
 autom4te.cache
 .DS_Store
 *~
+/doc
```
craig-chasseur: Missing newline at end of file. Also, this change appears to be unrelated to the purpose of this PR. Could you give some context as to what tool is producing this `/doc` directory and open a separate PR to change it if needed?

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
+/**
+ *   Copyright 2011-2015 Quickstep Technologies LLC.
```
craig-chasseur: Since this is a new file, it shouldn't have the Quickstep Technologies copyright, only the new copyright for Pivotal (also, the year should by 2016, or 2015-2016 if you also worked on it last year).

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * @brief Parsed BlockPropertyItem which simply a key (property) and the corresponding value.
+ * @details The value may actually be a list of Parse strings. This is to cover the case where a 
+ *    user specifies a subset of columns to compress rather than using the keyword ALL 
+ */
+class ParseBlockPropertyItem : public ParseTreeNode {
+ public:
+
+  // string representation of the ALL SQL keyword
+  static const std::string AllString;
```
craig-chasseur: Constants are named starting with a k like so: `kAllString`. Also, prefer using a C-style character array (e.g. `static const char kAllString[];`) for string constants rather than a `std::string` object where possible, as that does not require object construction and destruction as the program is loaded and unloaded respectively.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * @brief Constructor.
+   *
+   * @param line_number Line number of the first token of this node in the SQL statement.
+   * @param column_number Column number of the first token of this node in the SQL statement.
+   * @param property describes which type of block property this is. Corresponds to a field of the 
+   *          StorageBlockLayoutDescription message
+   * @param value a parse string representing the assignment of the property
+   **/
+  ParseBlockPropertyItem( int line_number,
```
craig-chasseur: Style nit: remove extra space between opening paren `(` and parameter.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+   * @param column_number Column number of the first token of this node in the SQL statement.
+   * @param property describes which type of block property this is. Corresponds to a field of the 
+   *          StorageBlockLayoutDescription message
+   * @param value a parse string representing the assignment of the property
+   **/
+  ParseBlockPropertyItem( int line_number, 
+                          int column_number,
+                          Property property,
+                          ParseString* value)
+                                  : ParseTreeNode(line_number, column_number),
```
craig-chasseur: Style nit: usually only indent 4 extra spaces for a new level.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+                          PtrList<ParseString>* values)
+                                  : ParseTreeNode(line_number, column_number),
+                                    property_(property),
+                                    values_(values) { }
+
+  std::string getName() const override {
+    return "BlockPropertyItem";
+  }
+
+  Property property() const {
```
craig-chasseur: Please doxygen all public methods. For very simple accessors like this, it can be something as simple as:
```
@return The type of property set by this item.
```

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+ * @details The value may actually be a list of Parse strings. This is to cover the case where a 
+ *    user specifies a subset of columns to compress rather than using the keyword ALL 
+ */
+class ParseBlockPropertyItem : public ParseTreeNode {
+ public:
+
+  // string representation of the ALL SQL keyword
+  static const std::string AllString;
+
+  enum Property {
```
craig-chasseur: Consider using a C++11 strongly-typed enum (`enum class`) for additional safety.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+  std::string getName() const override {
+    return "BlockPropertyItem";
+  }
+
+  Property property() const {
+    return property_;
+  }
+
+  const ParseString* value() const {
```
craig-chasseur: This version that returns a single value will segfault if there are zero values. A user may also mistakenly call it when there is more than one value and not deal with the other values. Recommend dropping it and just using the `values()` accessor below.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  const PtrList<ParseString>& values() const {
+    return *(values_.get());
+  }
+
+  /**
+   * @brief Gets a string version of the property which this pair describes
+   * 
+   * @return a string description of the property
```
craig-chasseur: Could you expand this a bit and describe the format of the string that is returned?

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+      std::vector<std::string> *inline_field_values,
+      std::vector<std::string> *non_container_child_field_names,
+      std::vector<const ParseTreeNode*> *non_container_child_fields,
+      std::vector<std::string> *container_child_field_names,
+      std::vector<std::vector<const ParseTreeNode*>> *container_child_fields) const override {
+
+    inline_field_names->push_back("property");
+    inline_field_values->push_back(getPropertyString());
+
+    inline_field_names->push_back("value");
```
craig-chasseur: Here you consider values as an inline field and manually stitch a string together to represent it. I think it makes more sense to view the values as a container child field and use the logic we already have for printing those out.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+      str.erase(str.size() - 1, 1);
+      str += ")";
+      inline_field_values->push_back(str);
+    } else {
+      inline_field_values->push_back(values_->begin()->value());
+    }
+  }
+
+ private:
+
```
craig-chasseur: Get rid of extra newline here.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+ */
+
+/**
+ * @brief Parsed BlockPropertyItem which simply a key (property) and the corresponding value.
+ * @details The value may actually be a list of Parse strings. This is to cover the case where a 
+ *    user specifies a subset of columns to compress rather than using the keyword ALL 
+ */
+class ParseBlockPropertyItem : public ParseTreeNode {
+ public:
+
```
craig-chasseur: Get rid of extra newline.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * @brief Returns if the user specified to compress all attributes using a special keyword
+   * 
+   * @return true if this is a compression property and the value is all
+   */
+  bool compressAll() const;
+
+ protected:
+
```
craig-chasseur: Get rid of extra newline.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+  DISALLOW_COPY_AND_ASSIGN(ParseBlockPropertyItem);
+};
+
+/**
+ * @brief Contains user-specified physical properties of a table's blocks as BlockPropertyItem's
+ */
+class ParseBlockProperties : public ParseTreeNode {
+ public:
+
```
craig-chasseur: Get rid of extra newline.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * @brief Constructor.
+   *
+   * @param line_number Line number of the first token of this node in the SQL statement.
+   * @param column_number Column number of the first token of this node in the SQL statement.
+   * @param property describes which type of block property this is. Corresponds to a field of the 
+   *          StorageBlockLayoutDescription message
+   * @param values list of parse strings representing the assignment of the property
+   **/
+  ParseBlockPropertyItem( int line_number,
```
craig-chasseur: Remove extra space between opening paren and parameter.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+class ParseBlockProperties : public ParseTreeNode {
+ public:
+
+  /**
+   * @brief Constructor.
+   *
+   * @param line_number Line number of the first token of this node in the SQL statement.
+   * @param column_number Column number of the first token of this node in the SQL statement.
+   **/
+  ParseBlockProperties( int line_number,
```
craig-chasseur: Get rid of extra space between paren and parameter.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * @brief Constructor.
+   *
+   * @param line_number Line number of the first token of this node in the SQL statement.
+   * @param column_number Column number of the first token of this node in the SQL statement.
+   **/
+  ParseBlockProperties( int line_number, 
+                        int column_number, 
+                        PtrList<ParseBlockPropertyItem>* properties)
+                                : ParseTreeNode(line_number, column_number),
```
craig-chasseur: Usually only indent 4 extra spaces for a new level.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /**
+   * @brief Gets the specified BlockPropertyItem.
+   * 
+   * @param property The property, one of kSort, kCompress, or kType.
+   * @return Pointer to the property or nullptr if not found.
+   */
+  const ParseBlockPropertyItem* getPropertyItem(ParseBlockPropertyItem::Property property) const {
+    for(auto propertyItem = properties_->begin(); propertyItem != properties_->end(); ++propertyItem) {
```
craig-chasseur: We generally prefer to avoid using `auto` unless it is actually necessary in a template context, or it significantly improves readability. Here, it would probably be clearer to just write a range-for loop with an explicit type like so:
```
for (const ParseBlockPropertyItem &item : *properties_) {
  if (item.property() == property) {
    return &item;
  }
}
return nullptr;
```

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+  /**
+   * @brief Constructor.
+   *
+   * @param line_number Line number of the first token of this node in the SQL statement.
+   * @param column_number Column number of the first token of this node in the SQL statement.
+   **/
+  ParseBlockProperties( int line_number, 
+                        int column_number, 
+                        PtrList<ParseBlockPropertyItem>* properties)
```
craig-chasseur: Missing doxygen for this parameter (should also make it clear that this object takes ownership of `properties`).

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,218 @@
...
+ protected:
+  void getFieldStringItems(
+      std::vector<std::string> *inline_field_names,
+      std::vector<std::string> *inline_field_values,
+      std::vector<std::string> *non_container_child_field_names,
+      std::vector<const ParseTreeNode*> *non_container_child_fields,
+      std::vector<std::string> *container_child_field_names,
+      std::vector<std::vector<const ParseTreeNode*>> *container_child_fields) const override {
+
+    for(auto propertyItem = properties_->begin(); propertyItem != properties_->end(); ++propertyItem) {
```
craig-chasseur: See above about use of `auto`.

craig-chasseur: Really, a `ParseBlockProperties` object is a container of `ParseBlockPropertyItem`s. So, you would have a single container child field name "properties" and all the contained properties as a single entry in the fields underneath it.

parser/ParseBlockProperties.cpp
craig-chasseur: Don't need legacy Quickstep Technologies copyright for new files.

parser/ParseBlockProperties.cpp
```diff
@@ -0,0 +1,43 @@
...
+      return "sort";
+    case kType:
+      return "type";
+    default:
+      return "unknown";
+  }
+}
+
+bool ParseBlockPropertyItem::compressAll() const {
+  return (property_ == kCompress && values_->begin()->value().compare(AllString) == 0);
```
craig-chasseur: What if there are multiple `values_` and one of them (not the first) is ALL? What if the user specifies a column named "ALL" with double-quotes (so that the lexer/parser treat it as a name literal and not the SQL ALL keyword)?

Rather than doing late string-comparison here, I think it makes more sense to have a special boolean flag indicating "all" that is set by a special constructor for ParseBlockPropertyItem. Then, the parser could create the ParseBlockPropertyItem with that constructor when it encounters the SQL keyword ALL by itself, but it will not become confused in other circumstances (e.g. if double-quoted "ALL" is used as a column name).

parser/SqlLexer.lpp
```diff
@@ -39,6 +39,8 @@ class Comparison;
 class ParseAssignment;
 class ParseAttribute;
 class ParseAttributeDefinition;
+class ParseBlockProperties;
```
craig-chasseur: When changing the Lexer or Parser sources, make sure to also regenerate the preprocessed source files (cd into `parser/preprocessed` and run the `genfiles.sh` shell script). Our build system uses flex and bison to make the lexer and parser when they are available, but it has the preprocessed files as a fallback if flex and/or bison aren't installed.

parser/SqlParser.ypp
```diff
@@ -759,6 +777,40 @@ opt_table_constraint_def_commalist:
     /* $$ = $1; */
   };
 
+opt_block_properties:
+  TOKEN_WITH TOKEN_BLOCKPROPERTIES '(' block_property_item_list ')' {
+    $$ = new quickstep::ParseBlockProperties(@1.first_line, @1.first_column, $4);
+  } | {
```
craig-chasseur: For consistency with the style of cases in the rest of this file, put the `|` beginning the next case on its own line after the closing `}` bracket.

parser/SqlParser.ypp
```diff
@@ -759,6 +777,40 @@ opt_table_constraint_def_commalist:
...
+  | TOKEN_SORT any_name {
+    $$ = new quickstep::ParseBlockPropertyItem(@1.first_line, @1.first_column, quickstep::ParseBlockPropertyItem::kSort, $2);
+  }
+  | TOKEN_COMPRESS '(' name_commalist ')' {
+    $$ = new quickstep::ParseBlockPropertyItem(@1.first_line, @1.first_column, quickstep::ParseBlockPropertyItem::kCompress, $3);
+  } 
+  | TOKEN_COMPRESS TOKEN_ALL {
+    $$ = new quickstep::ParseBlockPropertyItem(@1.first_line, @1.first_column, 
+                    quickstep::ParseBlockPropertyItem::kCompress, 
+                    new quickstep::ParseString(0, 0, quickstep::ParseBlockPropertyItem::AllString));
```
craig-chasseur: `0, 0` is not the accurate location for this string (but see notes in `ParseBlockProperties.cpp` about not representing "ALL" as a string).

parser/SqlParser.ypp
```diff
@@ -759,6 +777,40 @@ opt_table_constraint_def_commalist:
...
+  }
+  | TOKEN_SORT any_name {
+    $$ = new quickstep::ParseBlockPropertyItem(@1.first_line, @1.first_column, quickstep::ParseBlockPropertyItem::kSort, $2);
+  }
+  | TOKEN_COMPRESS '(' name_commalist ')' {
+    $$ = new quickstep::ParseBlockPropertyItem(@1.first_line, @1.first_column, quickstep::ParseBlockPropertyItem::kCompress, $3);
+  } 
+  | TOKEN_COMPRESS TOKEN_ALL {
+    $$ = new quickstep::ParseBlockPropertyItem(@1.first_line, @1.first_column, 
+                    quickstep::ParseBlockPropertyItem::kCompress,
```
craig-chasseur: See this section of the style guide for rules on indenting parameters for function calls: https://google.github.io/styleguide/cppguide.html#Function_Calls

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -761,6 +762,25 @@ void ExecutionGenerator::convertCreateTable(
     ++aid;
   }
 
+  // if specified, set the physical block type
```
craig-chasseur: For the most part, comments are written with regular English grammar. So, capitalize the start of sentences and end them with periods.

query_optimizer/resolver/Resolver.hpp
```diff
@@ -160,6 +163,18 @@ class Resolver {
...
   /**
+   * @brief Helper method for resolving physical block properties.
+   *
+   * @param block_properties The user's parsed block properties
+   * @param attributes The attributes for the table (needed for SORT and COMPRESS
+   *          attribute name to id mappings)
+   * @return A valid TupleStorageSubBlockDescription. Caller takes ownership
+   **/
+  TupleStorageSubBlockDescription* resolveBlockProperties(
+                        const ParseBlockProperties& block_properties,
```
craig-chasseur: See style guide rules about indenting function parameters: https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions

craig-chasseur:
Still working my way through this PR. I have some comments on the parser parts, though. For now, one of the most important things to do is fix everything that's coming up red in Travis. That makes it easier to review knowing that the mechanically-checked stuff isn't broken.

storage/BasicColumnStoreTupleStorageSubBlock.cpp
cramja: @zuyu Although it's not obvious, changing the return value is preparation for the next PR which completes the BLOCK PROPERTIES feature. The system will create a proto message and then send it to the SubBlockTypeRegistery to get a return code and generate an SQL error explaining the return code. This saves the repeating of all the checks which blocks already do with their validation functions.

cramja:
@craig-chasseur @pateljm Thank you for going through thoroughly. I will run with these comments and try to address them in individual commits where possible. A couple of questions:

* Travis is using cpplint. Is this the google python script which checks against google's style guide? Should we run it before all commits, or just get more familiar with the Google Style guide?
* Before submitting a PR, is there a way to run through the build/test sequence which Travis goes through?
* How to do correct printing of the ParseTree isn't obvious, even from looking at the test cases. Is there someone at UW who could give a quick explanation of what the logic behind this is (including the Logical and Physical variants)?
* About Error Codes vs Returned enums. I agree, the current change to returning error codes in the form of integers doesn't seem to be a good solution because it is not self-documenting. @craig-chasseur what are your thoughts on returning an enum, and if this is a better way to do it, should the declaration of this enum be in the base class for tuple stores, TupleStorageSubBlock?

cramja:
Also, I was doing work on the implementation of the BLOCK PROPERTIES on the logical and physical levels yesterday and did not realize that commits I pushed to the branch would be added to the PR. If it makes sense to break this PR up into two separate PRs I can do that, though I will need someone who is familiar with these git features to make sure that I do it correctly. Or I don't mind doing the PR in a monolith, except, obviously it's hard to comment on specifics.

craig-chasseur:
To answer your questions:

1. We have a script that runs our modified cpplint over all the sources. I
highly recommend running it before you commit. The script is in
third_party/cpplint/lint_everything.py. You should also run
validate_cmakelists.py (in the repo root) to check that your CMakeLists
dependencies are properly set, and cyclic_dependency.py to check that there
are no circular linking dependencies.
2. You can read the file ".travis.yml" to see what the Travis CI setup is
doing. Basically, it runs the 3 python lint/validation scripts mentioned
above, then builds Quickstep for a particular entry in our build matrix
(the dimensions are RELEASE vs. DEBUG, GCC vs. Clang, and
VECTOR_COPY_ELISION_LEVEL set to "none" or "joinwithbinaryexpressions"). We
always compile with -Werror, so any compiler warnings will be treated as
errors and cause the build to fail. Finally, ctest is invoked to run all
the unit tests. If there is a problem with any of these tests, then the
build is considered failed.
3. Qiang (unfortunately no longer at UW) is the person who originally wrote
the TreeStringSerializable code and most of the current optimizer. The
parser itself is a mix of his work and mine. Basically, there are three
kinds of "fields" in a TreeStringSerializable: inline fields (arbitrary
"flat" members of a node), non-container child fields (where there is a
*single* child node nested under a node), and container child fields (where
there is some number of child fields that the root node acts as a
"container" for).
4. Generally an enum is preferable to integer error codes, however in this
case I think it is preferable to keep using bool to preserve the simplicity
of the API. Most client code only cares about whether the description is
valid, not why it might be (and the proto debug string can be dumped in
those cases where we really want to know what's up).

Also note that I haven't completed a full pass on all the code yet, only
the parser part so far.

craig-chasseur:
Definitely do a separate PR. We don't want to pollute an open PR with a
bunch of new code and WIP features. You can use any existing branch as the
base for a new branch in git, not just master. So, from within a branch
called, say "existing-feature", you would do:

git checkout -b my-new-feature

cramja:
@craig-chasseur @zuyu @saketj Saket, I've included you on this comment because it relates to your CREATE INDEX task. 

We want to give users a descriptive error message when they make an invalid layout via BLOCKPROPERTIES or CREATE INDEX. There's already code to check for validity in the blocks themselves. If we change the return values of these methods from bools to int or enums, we can reuse the validity code in the parser. I see a few pros and cons:
Pros: 
+ Reuse checking code in DescriptionIsValid
+ Better user feedback for SQL errors
Cons
- Touches several files related to storage
- enum return values are slightly less intuitive than true/false

Originally, I had opted to re-write much of the error-checking code in the Resolver and the ExecutionGenerator but the amount of repetition led me to changing the return value of DescriptionIsValid.

Final thoughts?

cramja:
@craig-chasseur I misunderstood how git works and committed a lot of new code on top of the PR changes. I'm thinking the easiest way forward might be to branch off an earlier commit with the changes I originally committed to the PR- with the parser code- and create a new PR on this separate branch. I can then later merge the implementation changes in an submit a new PR. This would mean closing this PR and creating a new, smaller one essentially.

storage/BasicColumnStoreTupleStorageSubBlock.cpp
zuyu: No. If a proto is invalid, we just stop the program and print the proto for debug purpose.

pateljm:
@cramja I think it is ok if you want to simply open a new PR. We can simply delete this PR. Lots of good stuff about typical issues that new committers might face, so before we kill this PR we should list these perhaps on a wiki page.

cramja:
@pateljm I will do this. Yes, I think especially Saket will benefit from the information on this thread. So, how about keeping this thread around for a week or so while I get another PR up and accepted, and from all this, I'll start a wiki "Pull Request Guide" or something other... I could see it merging with the "Code Review Guidelines" page.

saketj:
@cramja thanks Marc for including in this thread. Couple of useful information related to PR and coding guidelines that are really helpful.

parser/ParseBlockProperties.hpp
cramja: Note! Admittedly, I do not fully understand the instanciation process in C++ (like why can you not assign kAllString within the class?)

parser/ParseBlockProperties.hpp
cramja: Noted!

parser/ParseBlockProperties.hpp
cramja: Noted!

zuyu:
@cramja Closing a PR does NOT mean to delete all info here. You could always find this PR using the same URL.

cramja:
@zuyu That's good news.

@zuyu @craig-chasseur What is the convention for commenting multiple constructors in a row with many shared arguments but the tailing arguments being unique per constructor? 

For example:
```c++
  /**
   * @brief Constructor.
   *
   * @param line_number Line number of the first token of this node in the SQL statement.
   * @param column_number Column number of the first token of this node in the SQL statement.
   * @param property describes which type of block property this is. Corresponds to a field of the 
   *          StorageBlockLayoutDescription message
   * @param value a parse string representing the assignment of the property
   **/
  ParseBlockPropertyItem(int line_number, 
                         int column_number,
                         Property property,
                         ParseString* value)
                             : ParseTreeNode(line_number, column_number),
                               property_(property) {
    values_.reset(new PtrList<ParseString>);
    values_->push_back(value);
  }

  /**
   * @brief Constructor.
   *
   * @param line_number Line number of the first token of this node in the SQL statement.
   * @param column_number Column number of the first token of this node in the SQL statement.
   * @param property describes which type of block property this is. Corresponds to a field of the 
   *          StorageBlockLayoutDescription message
   * @param values list of parse strings representing the assignment of the property
   **/
  ParseBlockPropertyItem(int line_number, 
                         int column_number,
                         Property property,
                         PtrList<ParseString>* values)
                             : ParseTreeNode(line_number, column_number),
                               property_(property),
                               values_(values) { }
```

In this example, can we delete the repeated line_number explanation?

cramja:
@craig-chasseur @zuyu Another style question. Can we use lambdas?
For example, I have a helper function in resolver.cpp:
```C++
// helper method which is used multiple times in resolveBlockProperties
// returns -1 if given name was not found among the attribute names
// it might be possible to change this to a lambda statment within the method resolveBlockProperties
attribute_id attributeIdFromAttributeName(const std::string& name, const PtrList<ParseAttributeDefinition>& attributes) {
  std::string lower_name = ToLower(name);
  attribute_id attr_id = 0;
  for (const ParseAttributeDefinition &attribute_definition : attributes) {
    if(lower_name.compare(ToLower(attribute_definition.name()->value())) == 0){
      return attr_id;
    }
    ++attr_id;
  }
  return -1;
}
```

Can I instead declare this as a lambda in the beginning of the only function which I am using it in?

cramja:
Also, some areas of code seem to use a 100 space limit while others use 80. Is there a consensus on this?

zuyu:
@cramja For the code length, we set to `120` chars. But for comments `80` chars per line.

cramja:
I was getting the feeling a lot of these questions were answered already. For future reference, here's something one should read before submitting a pr [style guide](https://github.com/pivotalsoftware/quickstep/wiki/Quickstep-Style-Guide). Also, maybe it should be updated to include mention of the lint script. Either way, I will mention that in the yet-to-be-written newbie pr guide.

PR #13, opened 2016-01-26
Removed redundant args and mark args const in Foreman.

zuyu:
This pr improved `Foreman` APIs in two folds:
* Removed the redundant `operator pointer` argument, as it could be referenced by another passing argument called `operator index`.
* Marked arguments const, wherever possible.


hbdeshmukh:
Looks good to me, merging.

PR #17, opened 2016-01-27
Block properties parser

cramja:
**NEW PR** (in other words, you've probably seen most of this code before)

**Parser support for BLOCKPROPERTIES**
This PR contains parser support for the BLOCKPROPERTIES clause for the CREATE TABLE statement. Currently the parser recognizes the syntax:
```SQL
CREATE TABLE name (attr1 type, attr2 type ... ) 
   WITH BLOCKPROPERTIES (TYPE columnstore, sort attr1, COMPRESS ALL); 
```
and several variants. Users can specify storage types, sort columns where supported, compressed columns where supported. Sort types will include the 5 currently supported types. COMPRESS may be followed with the keyword ALL or a comma list of attribute names enclosed in parentheses like COMPRESS (attr1, attr2)

A seperate pull request will be made for the Optimizer pieces.


parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,247 @@
...
+   * @param property Describes which type of block property this is. Corresponds
+   *  to a field of the StorageBlockLayoutDescription message.
+   * @param value A parse string representing the assignment of the property. 
+   *  ParseBlockPropertyItem takes ownership of the pointer.
+   **/
+  ParseBlockPropertyItem(int line_number,
+                         int column_number,
+                         Property property,
+                         ParseString* value)
+                             : ParseTreeNode(line_number, column_number),
```
zuyu: This is not the correct way to indent. Please read the similar files to get familiar with the code style.

```
  ParseBlockPropertyItem(int line_number, typeX xxx)
      : ParseTreeNode(line_number),
         xxx_(xxx) {
  }
```

cramja:
@pateljm @craig-chasseur I have almost completed the PR with the optimizer pieces for BLOCKPROPERTY support. If this is ready to be merged, can we merge/close? Then I will create another PR, first one to refactor the parser unit tests (capitalize SQL keywords) then one with the optimizer bits.

parser/ParseBlockProperties.cpp
```diff
@@ -0,0 +1,40 @@
...
+      return "sort";
+    case Property::kType:
+      return "type";
+    default:
+      return "unknown";
+  }
+}
+
+bool ParseBlockPropertyItem::compressAll() const {
+  return property_ == Property::kCompress && compress_all_;
```
craig-chasseur: A simple accessor like this can go in the header (helps the compiler to inline it).

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,245 @@
...
+    inline_field_names->push_back("property");
+    inline_field_values->push_back(getPropertyString());
+
+    if (values_->size() > 1) {
+      container_child_field_names->push_back(std::string("values"));
+      container_child_fields->emplace_back();
+      for (const ParseString& item_value : *values_) {
+        container_child_fields->back().push_back(&item_value);
+      }
+    } else if (values_->size() == 1) {
```
craig-chasseur: I think a single value should also be treated as a (single element) list using `container_child_fields` instead of this special case.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * @brief Parsed BlockPropertyItem which simply a key (property) and the 
+ *        corresponding value.
+ * @details The value may actually be a list of Parse strings. This is to cover 
+ *          the case where a user specifies a subset of columns to compress 
+ *          rather than using the keyword 'ALL'. 
+ */
+class ParseBlockPropertyItem : public ParseTreeNode {
+ public:
+  enum class Property {
```
craig-chasseur: This doesn't handle describing IndexSubBlocks (although I understand another student is working on that)... what about the number of slots for a StorageBlockLayout, though? How would the user specify that they want more than 1 slot?

parser/ParseStatement.hpp
```diff
@@ -153,6 +156,7 @@ class ParseStatementCreateTable : public ParseStatement {
  private:
   std::unique_ptr<ParseString> relation_name_;
   std::unique_ptr<PtrList<ParseAttributeDefinition> > attribute_definition_list_;
+  std::unique_ptr<ParseBlockProperties> opt_block_properties_;
```
craig-chasseur: You should probably add a const accessor method so that this field can be read after the object is created.

craig-chasseur:
Just finished a pass. Mostly looks good, although please see the comment about how to represent that a layout should have >1 slot.

cramja:
Thanks for the feedback. Hmm, slots, to be honest I didn't think this was important so I was planning on always setting it to one. However, it's not hard at all to account for this and I can add this to the syntax.

Maybe
```sql
WITH BLOCKPROPERTIES (TYPE ~, SORT ~, COMPRESS ~, SLOTS n);
```

where n is between 1 and something reasonable but not excessively large (42).

What do you think?

parser/ParseBlockProperties.cpp
cramja: If both of these methods are moved to the header and there is nothing left in this file but the namespace declaration, we can delete it right?

parser/ParseBlockProperties.hpp
cramja: I like that idea.

parser/ParseBlockProperties.cpp
pateljm: Yup -- you can delete the file if all this moves to the header.

pateljm:
The # slots will be determined by other statements like create index and should be handled automatically by the SM. So we should leave it out from the user specified syntax here. Comments @craig-chasseur?

craig-chasseur:
Well, the block-based architecture encapsulates both storage layout and
indices. I think the number of slots should be configurable.

pateljm:
@craig-chasseur Ok -- just wanted to make sure we are referring to the same thing. The slot here then would control the block size for the relation? If so, should we let the user specify the slot size, or the storage block size, which we can then round to the closest multiple of the smallest default slot size?

cramja:
@pateljm I am going to add test code for this feature today. I know we wanted to shift over to using all caps for Sql keywords- maybe we should do this in a separate PR, capitalizing all the parser tests in one go?

cramja:
I am working on adding the support for `BLOCKSIZE`. To support the numeric value, I am templatizing the ParseBlockPropertyItem which is a large amount of changes. Shall I keep the same PR or open a new one?

craig-chasseur:
Are you sure you need to templatize it? Could you make it so that the value
is optionally an integer instead of a list of strings (essentially
tagged-union semantics, although not necessarily with an actual C++ union)?

cramja:
I'm not familiar with tagged union sematics. Would that mean something along the lines of adding the NumericParseLiteral to the ParseBlockPropertyItem and then accounting for the case where we use the numeric instead of strings in the class methods?

If that's the case, it does seem simpler and I will do that.

craig-chasseur:
That's the idea. :)

pateljm:
@cramja Good idea about a separate PR to fix the SQL!

cramja:
@craig-chasseur I updated BlockProperties based on your suggestion to support parsed Numerics. 

I also made the constructor for BlockPropertyItem private and now the Parser code must call static methods in BlockPropertyItem to create specific types of BlockPropertyItems. I did this because I wanted to restrict possible usage of BlockProperties to make it harder to create an invalid combination of properties.

Also, trying to CREATE a block with BLOCKPROPERTIES now throws an SQL error saying that the feature is not supported yet.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,331 @@
...
+    kSort,
+    kType
+  };
+
+  /**
+   * @brief Creates a BlockSize block property item.
+   *
+   * @param line_number Line number of the first token of this node in the SQL statement.
+   * @param column_number Column number of the first token of this node in the SQL statement.
+   * @param numeric_value The size value of the block in Mb.
```
craig-chasseur: "Mb" (with lowercase "b") actually means megabits, i.e. 128 KB. Better to write out megabytes (or mebibytes) for clarity.

Also, just for clarity, are we measuring the size in megabytes or in slots (currently 2MB each)? If the former, how do we deal with a requested size that's not an exact multiple of the slot size?

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,331 @@
...
+   *
+   * @return The specific block property which this instance is describing.
+   */
+  Property property() const {
+    return property_;
+  }
+
+  /**
+   * @return A list of ParseString values which the parser has captured to
+   *         describe this property. This may be empty in the case of 'COMPRESS ALL'.
```
craig-chasseur: This comment is no longer accurate. On the one hand, it may be empty in the case of a numeric block size property. On the other hand, an "ALL" string is now added to the values when `GetCompressAllItem()` is called.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,331 @@
...
+      std::vector<std::string> *inline_field_values,
+      std::vector<std::string> *non_container_child_field_names,
+      std::vector<const ParseTreeNode*> *non_container_child_fields,
+      std::vector<std::string> *container_child_field_names,
+      std::vector<std::vector<const ParseTreeNode*>> *container_child_fields) const override {
+    inline_field_names->push_back("property");
+    inline_field_values->push_back(getPropertyString());
+
+    if (hasStringValues()) {
+      container_child_field_names->push_back((values_->size() == 1) ? "value" : "values");
```
craig-chasseur: It's probably easier (if not 100% grammatically correct from a human perspective) to always use the plural "values" (that way it's easier to write tools that parse the output of a serialized tree).

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,331 @@
...
+      container_child_fields->emplace_back();
+      for (const ParseString& item_value : *values_) {
+        container_child_fields->back().push_back(&item_value);
+      }
+    } else {
+      container_child_field_names->push_back("value");
```
craig-chasseur: See above about always using the plural "values".

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,331 @@
...
+   * @return Name of the parser construct represented by this class.
+   */
+  std::string getName() const override {
+    return "BlockProperties";
+  }
+
+  /**
+   * @brief Gets the specified BlockPropertyItem.
+   *
+   * @param property The property, one of kSort, kCompress, or kType.
```
craig-chasseur: ... or `kBlockSize`

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,331 @@
...
+   * @return Pointer to the property or nullptr if not found.
+   */
+  const ParseBlockPropertyItem* getPropertyItem(ParseBlockPropertyItem::Property property) const {
+    for (const ParseBlockPropertyItem &item : *properties_) {
+      if (item.property() == property) {
```
craig-chasseur: Is there ever any validation to make sure that the same property is not specified twice? What about `COMPRESS`, where it is legal to specify more than one column that should be compressed (it seems like we only allow a single `COMPRESS` clause that optionally has comma-separated columns, which is OK, but we should probably have some validation to make sure that multiple clauses are superseding each other).

parser/SqlParser.ypp
```diff
@@ -759,6 +780,48 @@ opt_table_constraint_def_commalist:
...
+    $$ = quickstep::ParseBlockPropertyItem::GetCompressAllItem(@1.first_line, @1.first_column,
+                                                               @2.first_line, @2.first_column);
+  }
+  | TOKEN_BLOCKSIZE TOKEN_UNSIGNED_NUMVAL TOKEN_MB {
+    $$ = quickstep::ParseBlockPropertyItem::GetBlockSizeItem(@1.first_line, @1.first_column, $2);
+  }
+  | TOKEN_BLOCKSIZE TOKEN_UNSIGNED_NUMVAL TOKEN_GB {
+    delete $2;
+    $$ = nullptr;
+    NotSupported(&@3, yyscanner, "Gb unit not supported");
```
craig-chasseur: Why not? If you only want to allow units of megabytes, we might as well drop the `TOKEN_GB` entirely and only have the above case.

parser/SqlParser.ypp
```diff
@@ -759,6 +780,48 @@ opt_table_constraint_def_commalist:
...
+    $$ = quickstep::ParseBlockPropertyItem::GetSortItem(@1.first_line, @1.first_column, $2);
+  }
+  | TOKEN_COMPRESS '(' name_commalist ')' {
+    $$ = quickstep::ParseBlockPropertyItem::GetCompressAttributesItem(@1.first_line, @1.first_column, $3);
+  } 
+  | TOKEN_COMPRESS TOKEN_ALL {
+    $$ = quickstep::ParseBlockPropertyItem::GetCompressAllItem(@1.first_line, @1.first_column,
+                                                               @2.first_line, @2.first_column);
+  }
+  | TOKEN_BLOCKSIZE TOKEN_UNSIGNED_NUMVAL TOKEN_MB {
```
craig-chasseur: `TOKEN_UNSIGNED_NUMVAL` can have a fractional part and/or an arbitrary exponent. Any thoughts for how we would validate it as an integer?

craig-chasseur:
Hi @cramja. This is looking very good. I just have a couple of small suggestions and questions about input validation in the latest review pass. Those should be easy to address, then we can get this merged. Thank you for your patience and diligence in this process!

cramja:
@craig-chasseur You're welcome! Thank you for the comments on this process, it is helpful in the learning process for C++/QS.

Today during the group meeting, @saketj and I came up with a way to share code between the CREATE INDEX and CREATE TABLE statements. We're going to create a general key-value parse syntax to handle all the special properties of the index and storage blocks.

Basically the parser construct will look something like, ( ParseString ParseLiteral, ... ParseString ParseLiteral ) where the string is the property name and the Literal is either the numeric or string value.

As a consequence everything in this pull request is sort of void and I will close it. Then @saketj  and I will create a new pull request for the key-value syntax and Parse classes.

craig-chasseur:
OK, that sounds good to me. Code reuse is definitely a good thing. I'm
giving @saketj some comments on his PR right now just to think about things
for the future.

PR #19, opened 2016-01-28
CreateTable inserts the table into Catalog without using WorkOrder.

zuyu:
`CreateTableOperator` does not need `WorkOrder`-based execution model. Instead, the table will get inserted immediately into the master `Catalog` once `Foreman` processes it.

This is consistent with distributed Quickstep, and avoids the unnecessary `WorkOrder` serialization.

### DIFF
--- a/relational_operators/CreateTableOperator.cpp
+++ b/relational_operators/CreateTableOperator.cpp
@@ -21,8 +21,6 @@
 
 #include "catalog/CatalogDatabase.hpp"
 
-#include "glog/logging.h"
-
 namespace quickstep {
 
 bool CreateTableOperator::getAllWorkOrders(WorkOrdersContainer *container) {
--- a/relational_operators/CreateTableOperator.cpp
+++ b/relational_operators/CreateTableOperator.cpp
@@ -1,6 +1,6 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
- *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2015-2016 Pivotal Software, Inc.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -20,7 +20,6 @@
 #include <memory>
 
 #include "catalog/CatalogDatabase.hpp"
-#include "query_execution/WorkOrdersContainer.hpp"
 
 #include "glog/logging.h"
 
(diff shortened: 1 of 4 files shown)


pateljm:
@zuyu Note there is work to split the parser/optimizer from the rest of the system, so this change may need to be undone in the near future.

Can you state the goal for this change: Is it to speed up the CREATE TABLE command? Thanks!

zuyu:
@pateljm No, the split of the parser and the optimizer does not affect here. This PR is related to the `Foreman` based query execution and the master `Catalog`.

The goal of this PR is to avoid a `Work Order` to access the master `Catalog`, which co-locates with `Foreman` in the `Conductor`. Note that the `Work Order` may execute on a remote `Worker` node. Therefore, we do not need to serialize a `Work Order` for this operator.

pateljm:
@jianqiao @hbdeshmukh Hi: May I bug both of you to chime in to let @zuyu know if this looks good. 

@jianqiao I want to make sure this does not cause trouble for you as you work on the splitting of the frontend from there backend. 

If all looks good, let me know and I'm happy to close this PR.

jianqiao:
It does not affect my work. I think this is a design choice at the backend.

But one thing to note is that this modification will have Foreman's thread (i.e. the thread that runs Foreman::processMessage) executing the create table operation. If the create table operation becomes non-light-weight in the future (say catalog accessing is provided as a service). We might need to undo this design since it stalls foreman.

hbdeshmukh:
In terms of execution engine, I think it looks alright. 

@zuyu Have you tested the changes in the command line mode (apart from the unit tests)?

zuyu:
@hbdeshmukh Yes, it works in the shell CLI.

@jianqiao Yes, you are right. If the master `Catalog` is remote (as a service), we need to serialize `CatalogRelation`.

@pateljm Please close the PR. Thanks! And we need to add an end-to-end test for the shell CLI in the Travis CI, in addition to the existing unit tests.

relational_operators/CreateTableOperator.cpp
```diff
@@ -20,27 +20,15 @@
...
 
 namespace quickstep {
 
 bool CreateTableOperator::getAllWorkOrders(WorkOrdersContainer *container) {
   if (!work_generated_) {
     work_generated_ = true;
-    container->addNormalWorkOrder(
-        new CreateTableWorkOrder(relation_.release()),
-        op_index_);
+    database_->addRelation(relation_.release());
```
craig-chasseur: This is a somewhat hacky abuse of `getAllWorkOrders()`, in that it does actual work in a method that is only supposed to *produce* work.

I do agree that there is a need for "foreman only" operators, but we should probably make that explicit rather than hiding it here.

craig-chasseur:
Sorry to chime in late on an already-merged PR, but we may want to consider making the mechanism for doing work at the foreman more explicit and general, as noted in my above comment.

zuyu:
@craig-chasseur I'll revert this change later, as someone in the team will make the catalog as a service.

PR #20, opened 2016-01-28
Work orders container refactor

zuyu:
Use `unique_ptr` to hold `WorkOrder`s in `WorkOrdersContainer`.

### DIFF
--- a/query_execution/WorkOrdersContainer.hpp
+++ b/query_execution/WorkOrdersContainer.hpp
@@ -139,7 +139,8 @@ class WorkOrdersContainer {
    * @param numa_node_id The NUMA node.
    *
    * @return A WorkOrder which prefers numa_node_id. If no such WorkOrder is
-   *         available, return nullptr.
+   *         available, return nullptr. The caller is responsible for taking the
+   *         ownership.
    **/
   WorkOrder* getNormalWorkOrderForNUMANode(const std::size_t operator_index,
                                            const int numa_node_id) {
@@ -158,7 +159,8 @@ class WorkOrdersContainer {
    *        prefer exactly one NUMA node over workorders which list more than
    *        one NUMA node as their preference.
    *
-   * @return A WorkOrder. If no WorkOrder is available, returns nullptr.
+   * @return A WorkOrder. If no WorkOrder is available, returns nullptr. The
+   *         caller is responsible for taking the ownership.
    **/
   WorkOrder* getNormalWorkOrder(const std::size_t operator_index,
                                 const bool prefer_single_NUMA_node = true) {
@@ -175,7 +177,8 @@ class WorkOrdersContainer {
    * @param numa_node_id The NUMA node.
    *
    * @return A WorkOrder that prefers numa_node_id. If no such WorkOrder is
-   *         available, return nullptr.
+   *         available, return nullptr. The caller is responsible for taking the
+   *         ownership.
    **/
   WorkOrder* getRebuildWorkOrderForNUMANode(const std::size_t operator_index,
                                             const int numa_node_id) {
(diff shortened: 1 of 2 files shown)


hbdeshmukh:
Why do you need a unique pointer to store the WorkOrders? There are two possibilities - first is the normal execution. The container holds the pointer to a WorkOrder until it is scheduled. Once it is scheduled, the pointer is removed from the container and passed on to the worker that executes it. After the execution is over the worker frees the WorkOrder memory pointed by the pointer. 

In the second case when failure occurs, WorkOrderContainer has some un-executed WorkOrders which couldn't be scheduled. The container deletes them in its destructor. 

Therefore, I don't see a reason why the WorkOrders should have a unique pointer to the stored WorkOrders. If I am missing something, let me know.

zuyu:
You're right that for both cases. But for the failure case, it is much cleaner to use `unique_ptr` to hold pointers.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -311,16 +311,17 @@ class WorkOrdersContainer {
     }
 
     inline void addWorkOrder(WorkOrder *workorder) {
-      workorders_.push(workorder);
+      workorders_.emplace(workorder);
```
hbdeshmukh: Are you sure that this after this line, there will exist only a single pointer to ``workorder``? What happens to the
original pointer ``workorder``? For your reference, check out the discussion on this link - http://stackoverflow.com/questions/26094274/cannot-understand-how-to-add-new-object-to-stdliststdunique-ptrclassname

query_execution/WorkOrdersContainer.hpp
```diff
@@ -350,16 +351,17 @@ class WorkOrdersContainer {
     }
 
     inline void addWorkOrder(WorkOrder *workorder) {
-      workorders_.push_back(workorder);
+      workorders_.emplace_back(workorder);
```
hbdeshmukh: Same comments as line 314.

query_execution/WorkOrdersContainer.hpp
zuyu: Yes, either `emplace` or `emplace_back` creates a `unique_ptr` to contain `workorder`, which is created by `new`.

This is the same as the first example from the best answer by `Angew` in your given link.

@craig-chasseur Any comments?

hbdeshmukh: No @zuyu, that's not what the answer says. It says that ``emplace_back`` is not the correct choice when you are inserting a raw pointer that owns the resource, which is exactly our scenario. 

You can either construct the WorkOrder object (and thereby its pointer) right when you make the emplace call. That's not feasible in our case.

You need to call the ``make_unique`` function that's available in C++14. But considering that most of our code is C++11 compliant, I won't be a fan of using it here.

zuyu: Oh, I got the point: if an exception happens during the `emplace`, there will be memory leak for newly created `WorkOrder`. (Btw, the original version with `push_back` should have the same memory leak issue).

And as suggested by the same post, we could avoid the problem using the following code:

```c++
WorkOrder *workorder = new WorkOrder;

std::list<std::unique_ptr<WorkOrder>> workorders_;
workorders_.emplace_back(std::unique_ptr<WorkOrder>(workorder));
```

query_execution/WorkOrdersContainer.hpp
```diff
@@ -142,7 +142,7 @@ class WorkOrdersContainer {
    *         available, return nullptr.
```
hbdeshmukh: We need to add comments about the ownership-semantics here. It will be good if we explain the flow of ownership : Relational Operators produce the WorkOrders, the container owns it until it is scheduled. Once the WorkOrder is scheduled, the ownership is passed to the Worker that executes it and the executing worker deletes it.

zuyu:
@hbdeshmukh I've added the requested comments. Please take another look! Thanks!

hbdeshmukh:
Looks good, merging.

PR #21, opened 2016-01-30
Add support for CREATE INDEX syntax rules to SQL Parser

saketj:
This PR adds support for CREATE INDEX syntax rules to Quickstep SQL. This PR does not implement the CREATE INDEX functionality, but rather provides specific rules to support future implementations. Therefore as such, it touches only the lexical and syntactical SQL parser code. Following syntax for CREATE INDEX statement is proposed that will allow different types of indexes to be defined on some (or all) attributes of a table, along with the ability to customize and fine-tune index properties:

> CREATE INDEX **_index_name_** ON _**tbl_name**_ ( column_name, ... ) 
USING **_index_type_** (  _**index_properties**_ )

If a list of columns is not specified, then the index will be created on all the attributes of the table. Index properties are also optional and can be used to fine-tune the default properties of an index. Since Quickstep supports a multitude of indexes, USING clause is mandatory to select a specific index type (this can be changed though, if we want to support a default indexing mechanism). 

Unit tests for CREATE INDEX syntax rules are defined in parser/tests/Index.test.


cramja:
@saketj Parser code looks good. I like that we simply added the syntax and not the implementation. 
I have a couple of questions.
*  Is the index name significant?
*  Actually, index name points to a more general question: how will index creation/destruction be handled in the block-based architecture? 

How we handled the issue with Storage-properties was to force any physical declarations at table creation time meaning that all blocks of the relation will have the same storage properties. If an index is created after storage blocks already exist (contain tuples) do we want indices to be retro-actively created over the blocks or do we want heterogeneous storage? There's no problem with heterogeneous structure from a query-answering perspective, but maybe from a systems perspective there's something I'm missing.

It might ease your implementation to reuse some of the BlockProperties code to describe the index. As in the syntax might look like:
```SQL
CREATE INDEX btree_foo ON foo WITH PROPERTIES (
   TYPE btree, INDEX (attr1, attr2)
);
```

This way you could simply extend the ParseBlockProperties classes which are being added in the pr I'm [working on](https://github.com/pivotalsoftware/quickstep/pull/17/files#diff-393458487a2da2329d59fa8bfe81856dR60).

saketj:
@cramja  Thanks Marc for these comments. Yeah, the idea was to invite comments & feedback on the syntax rules for 'create index' in this PR and once the syntax is finalized, do the implementation part in the next PR. 

To begin with, I looked up at the syntax defined by various other SQL vendors and wanted to propose something that would be similar in structure to what they have, so it may seem familiar to the users. You are right in pointing out the significance of the index name itself and I guess, it was decided that we will not support building indexes on blocks that were already created before. However, the purpose of index name that I had in mind was for a different reason- Quickstep allows one to build different kinds of indexes on the storage blocks. An index name can be later used to identify these indexes separately, once they are defined. So, if a user decides to change the properties of a bloom filter index, he has a way to reference that index differently from a BitWeaving index that may be built on the same table. But again, if such use-cases are not something that will be expected, index names can be done away with. 

Also, I guess the index property parsing will have much in common with the BlockProperties parser and definitely I plan to reuse some portions of it. In fact, I was looking for a more general properties parser that can be defined for such kinds of syntax and something that can be abstracted and reused across multiple statements, like 'create table', 'create index', 'alter table', 'alter index' etc.

cramja:
@saketj I agree about generalizing the parse property class for use with other SQL statements. Maybe we could talk about this in person tomorrow if you are free at some point in the afternoon.

One thing about about naming indices is that the Catalog message structure doesn't provision for a named indices, so you'd probably need to extend the message and create a mapping between index and name in the `message CatalogRelation`.

saketj:
@cramja Sure Marc, great! I will catch up with you tomorrow to discuss on this. It seems now that naming indexes will need some changes to CatalogRelation. I haven't looked that part yet thoroughly, so I will probably wait to get views from everyone whether it will be useful or not in the future before I make changes to those.

pateljm:
@saketj This is a good start. I do like the idea of an index name, amongst other things it allows any errors that occur -- e.g. a tuple insertion failing because of an index property (unique) violation -- to be made more precise. Of course, we don't have unique/keys right now, but good to put in an index name right away.

saketj:
@pateljm Sure thing, Prof. Patel I will then keep on working on adding the index_name feature. Any comments on the other structures of the statement, like should we make USING clause mandatory to specify index type (like MySQL) or should type of the index be specified as part of the properties? 

USING is a new token that I have introduced in the Lex parser and have considered using the optional parentheses to specify the properties. So, the simplest index will look something like:
`CREATE INDEX any_index ON any_table USING any_index_type` 
and complicated ones can take the form
`CREATE INDEX any_index ON any_table (any_attributes,...) USING any_index_type (any_properties)`

parser/SqlLexer.lpp
```diff
@@ -120,6 +120,7 @@ unsigned_numeric_literal {exact_numeric_literal}|{approximate_numeric_literal}
 "between"          return TOKEN_BETWEEN;
 "bigint"           return TOKEN_BIGINT;
 "bit"              return TOKEN_BIT;
+"bloom_filter"     return TOKEN_BLOOM_FILTER_INDEX;
```
craig-chasseur: SQL keywords usually don't have underscores, so maybe call this "bloomfilter" instead.

craig-chasseur: We don't have a bloom filter index in master yet (although if one is in the works, that's awesome). We do, however, have a CSBTree index that we want to be able to use, so it would be good if we had a keyword for that so that we can specify such an index.

parser/SqlParser.ypp
```diff
@@ -440,6 +446,9 @@ void NotSupported(const YYLTYPE *location, yyscan_t yyscanner, const std::string
   table_constraint_def  // unimplemented
   table_constraint_def_commalist        // unimplemented
   opt_table_constraint_def_commalist    // unimplemented
+  opt_column_list       // unimplemented
```
craig-chasseur: Sounds like this will be a version of `name_commalist` that is optionally empty. You might want to just make a general-purpose `opt_name_commalist`.

parser/SqlParser.ypp
```diff
@@ -543,6 +555,15 @@ create_table_statement:
     $$ = new quickstep::ParseStatementCreateTable(@1.first_line, @1.first_column, $3, $5);
   };
 
+create_index_statement:
```
craig-chasseur: Having `CREATE INDEX` as a separate statement might be a bit tricky because of the way that index sub-blocks are handled in Quickstep. Since indices are internal to blocks and not external and global, it is difficult to add indices in an ad-hoc fashion if a relation already has blocks in it. We could modify the default StorageBlockLayout so that newly-created blocks have the appropriate indices, but dealing with pre-existing blocks is a trickier issue. We could either:

1. Leave existing blocks as-is without the new indices. This is lightweight and easy to implement but may confound users' expectations.
2. Reorganize the entirety of the existing relation with the new layout (essentially bulk-copying blocks into blocks of the new layout with a trivial SELECT operation). This is potentially quite heavyweight but ensures that all existing data is indexed. Another wrinkle is how we should deal with blocks that may not have the default layout for the relation (it's always been a goal for Quickstep, and something that is supported in the storage engine, to have different blocks in a relation have different layouts, although we don't have good support for that in the default SQL query path yet).

One option that sidesteps this issue entirely is to handle creating indices using a sub-clause of the `CREATE TABLE` statement. That way, we're always establishing indices as part of the default layout for a table up-front.

parser/SqlParser.ypp
```diff
@@ -759,6 +780,51 @@ opt_table_constraint_def_commalist:
     /* $$ = $1; */
   };
 
+opt_column_list:
+  /* empty */
```
craig-chasseur: The empty clause should return `NULL`.

parser/SqlParser.ypp
```diff
@@ -440,6 +446,9 @@ void NotSupported(const YYLTYPE *location, yyscan_t yyscanner, const std::string
   table_constraint_def  // unimplemented
   table_constraint_def_commalist        // unimplemented
   opt_table_constraint_def_commalist    // unimplemented
+  opt_column_list       // unimplemented
+  index_type            // unimplemented
```
craig-chasseur: I think it makes sense for the type of this to be the enum `IndexSubBlockType` from `storage/StorageBlockInfo.hpp`

parser/SqlParser.ypp
```diff
@@ -759,6 +780,51 @@ opt_table_constraint_def_commalist:
...
+opt_column_list:
+  /* empty */
+  | '(' name_commalist ')' {
+    delete $2;
+    NotSupported(&@1, yyscanner, "list of column names in CREATE INDEX statement");
+    YYERROR;
+  };
+
+index_type:
+  TOKEN_BLOOM_FILTER_INDEX {
```
craig-chasseur: As noted above, we ought to support CSBTree indices, too.

parser/SqlParser.ypp
```diff
@@ -759,6 +780,51 @@ opt_table_constraint_def_commalist:
...
+    NotSupported(&@1, yyscanner, "Bloom Filter Index");
+    YYERROR;
+  };
+
+opt_index_properties:
+  /* empty */
```
craig-chasseur: Empty case should return NULL.

parser/SqlParser.ypp
```diff
@@ -759,6 +780,51 @@ opt_table_constraint_def_commalist:
...
+    NotSupported(&@1, yyscanner, "index property in CREATE INDEX statement");
+    YYERROR;
+  }
+  | index_properties_list ',' index_property {
+    NotSupported(&@1, yyscanner, "list of index properties in CREATE INDEX statement");
+    YYERROR;
+  };
+
+index_property:
+  any_name literal_value {
```
craig-chasseur: Using the general `literal_value` might be a little too permissive, but I understand that you and @cramja are working on a general-purpose key-value configuration object that should handle these issues.

parser/SqlLexer.lpp
saketj: Thanks Craig for the reviews. I will update the token for bloom_filter and add a keyword for CSBTree as well. Yes, there is a working bloom filter implementation in one of the branches at University repo, I plan to integrate that with the Pivotal branch in coming weeks.

parser/SqlParser.ypp
saketj: Noted.

parser/SqlParser.ypp
saketj: Right, yeah I will change this to be of type string_list_

parser/SqlParser.ypp
saketj: Sure Craig, your concerns for supporting indices on pre-existing blocks is absolutely right and something that I also discussed here with Jignesh and Marc. So for now, the way we have decided is to just build indices on the new ones and not worry about the pre-existing blocks.

parser/SqlParser.ypp
saketj: Noted.

parser/SqlParser.ypp
saketj: Sure, will modify these to return NULL, though I was also tempted to use the %empty directive to specify empty rules, but this feature is not backward-compatible with bison 2.5 ( the one that is run on Travis builds)

parser/SqlParser.ypp
saketj: Yeah, we will soon replace this with a more generic key-value configuration that is sort of reusable across multiple statements.

cramja:
@saketj  and I spoke about the create index changes so that we are on the same page about what changes can be shared with the block property changes. The issue we ran into was how to eventually handle Drop Index for which we decided that at the current time the user can dump the database into a file and then reload it. Obviously a hack but would work. A better solution that can be baked into the current system is a new operator which copies an existing block into a new block with a different layout and then deletes the original block.

With that it looks good, going to accept.

parser/SqlLexer.lpp
```diff
@@ -120,6 +120,8 @@ unsigned_numeric_literal {exact_numeric_literal}|{approximate_numeric_literal}
 "between"          return TOKEN_BETWEEN;
 "bigint"           return TOKEN_BIGINT;
 "bit"              return TOKEN_BIT;
+"bloomfilter"      return TOKEN_BLOOM_FILTER_INDEX;
+"csbtreeindex"     return TOKEN_CSB_TREE_INDEX;
```
craig-chasseur: Keyword should be just "csbtree" (the extra trailing "index" is redundant).

parser/SqlParser.ypp
```diff
@@ -130,7 +130,7 @@ typedef void* yyscan_t;
   quickstep::PtrList<quickstep::ParseTableReference> *table_reference_list_;
   quickstep::ParseTableReferenceSignature *table_reference_signature_;
 
-  quickstep::ParseDataType *data_type_;
+  quickstep::ParseDataType *data_type_;
```
craig-chasseur: Extra trailing whitespace. Please remove.

parser/SqlLexer.lpp
```diff
@@ -120,6 +120,8 @@ unsigned_numeric_literal {exact_numeric_literal}|{approximate_numeric_literal}
 "between"          return TOKEN_BETWEEN;
 "bigint"           return TOKEN_BIGINT;
 "bit"              return TOKEN_BIT;
+"bloomfilter"      return TOKEN_BLOOM_FILTER_INDEX;
```
craig-chasseur: In most cases (except for keywords that are synonyms), the token name should follow from the actual keyword, e.g. `TOKEN_BLOOM_FILTER` without the trailing "_INDEX"

craig-chasseur:
Pinging @saketj and @cramja. Sorry I have been so busy and haven't taken a close look at PRs lately, but I noticed a couple of small issues in this PR that could use fixing. They are very minor and should be quick fixes.

saketj:
@craig-chasseur thanks for the comments. I will quickly fix these, although we have already merged this PR. What would you recommend- should a separate quick-fix branch be made to fix just these or can this be a part of the next PR that adds support for parsing key value pairs, (that PR will be opened soon) The new PR, nevertheless, touches the parser code.

craig-chasseur:
Hi Saket. I recommend a new, small PR just to do these little fixes.

saketj:
@craig-chasseur Thanks Craig, I have opened a separate PR that just does these fixes. Please have a look and I will wait for all the Travis checks to complete before merging.

PR #23, opened 2016-02-01
capitalized SQL keywords in parser tests

cramja:
Before adding the new CREATE INDEX and BLOCKPROPERTIES features and associated unit tests, the current parser tests needed to be refactored to capitalize SQL keywords. This should make reading the tests slightly easier and also stays true to the SQL standard of capitalizing SQL keywords.


zuyu:
LGTM.

cramja:
TY

PR #25, opened 2016-02-03
Serialized ForemanMessage as proto.

zuyu:
This PR, as a part of moving to distributed database,
 1. Replaced `ForemanMessage` as proto.
 2. Refactored how Foreman processes incoming messages.


query_execution/Foreman.cpp
```diff
@@ -192,15 +182,56 @@ void Foreman::run() {
...
+        // Dispatch the WorkerMessages to the workers. We prefer to start the search
+        // for the schedulable WorkOrders beginning from 'op_index'. The first
+        // candidate worker to receive the next WorkOrder is the one that sent the
+        // response message to Foreman.
+        // TODO(zuyu): Improve the data locality for the next WorkOrder.
+        dispatchWorkerMessages(0, op_index);
+        break;
+      }
+      case kWorkOrderFeedbackMessage: {
+        WorkOrder::FeedbackMessage msg(const_cast<void *>(tagged_message.message()),
```
hbdeshmukh: Why don't you need a proto version of the Feedback Message?

query_execution/QueryExecutionMessages.proto
```diff
@@ -0,0 +1,33 @@
...
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto2";
+
+package quickstep.serialization;
+
+// Used for both Normal WorkOrders and RebuildWorkOrders.
+message WorkOrderCompletionMessage {
```
hbdeshmukh: I agree that for now, we can live with single message type for both of them but separating messages for WorkOrder completion and Rebuild WorkOrder completion is better. 
By design, RebuildWorkOrders are always going to be about rebuilding a single block. Where as a WorkOrder itself could mean various things with respect to the input - e.g. hash tables, multiple blocks (e.g. sort). If we need to make any choices in the scheduler about the recently executed workorder and its input and output data, we will need to provide the information about the data locations through the completion messages. That's where separating the two message types will come handy.

query_execution/QueryExecutionMessages.proto
```diff
@@ -0,0 +1,33 @@
...
+  required uint64 operator_index = 1;
+  required uint64 worker_id = 2;
+}
```
hbdeshmukh: A basic question - if I have to add more fields in the message class, I should only add them here, right?

query_execution/QueryExecutionMessages.proto
```diff
@@ -0,0 +1,33 @@
...
+
+message DataPipelineMessage {
+  required uint64 operator_index = 1;
+  required fixed64 block_id = 2;
+  required int32 relation_id = 3;
+}
+
+message WorkOrdersAvailableMessage {
+  required uint64 operator_index = 1;
+}
```
hbdeshmukh: I think these message classes are neat and help us to get rid of the C++ classes for the message types, which were very lightweight anyway. Thanks!

query_execution/Worker.hpp
```diff
@@ -103,6 +104,18 @@ class Worker : public Thread {
...
+   * @brief Send the response WorkOrder completion message.
+   *
+   * @param receiver The id of the TMB client which should receive the response.
+   * @param op_index The index of the operator to which the WorkOrder belongs.
+   * @param isRebuildWorkOrder True if it is a RebuildWorkOrder. Otherwise
+   *        false.
+   **/
+  void sendWorkOrderCompleteMessage(const tmb::client_id receiver,
+                                    const std::size_t op_index,
+                                    const bool isRebuildWorkOrder);
```
hbdeshmukh: The style of variable naming is different than our usual standard which is - ``is_rebuild_workorder``

query_execution/Worker.cpp
```diff
@@ -94,4 +75,27 @@ void Worker::run() {
   }
 }
 
+void Worker::sendWorkOrderCompleteMessage(const tmb::client_id receiver,
+                                          const size_t op_index,
+                                          const bool isRebuildWorkOrder) {
```
hbdeshmukh: The style of variable naming is different than our usual standard which is - ``is_rebuild_workorder``

relational_operators/DeleteOperator.cpp
```diff
@@ -81,19 +85,25 @@ void DeleteWorkOrder::execute(QueryContext *query_context,
...
-  ForemanMessage message(ForemanMessage::DataPipelineMessage(delete_operator_index_,
-                                                             input_block_id_,
-                                                             rel_id_));
+  serialization::DataPipelineMessage proto;
+  proto.set_operator_index(delete_operator_index_);
+  proto.set_block_id(input_block_id_);
+  proto.set_relation_id(rel_id_);
+
+  const std::size_t proto_length = proto.ByteSize();
+  char *proto_bytes = static_cast<char*>(std::malloc(proto_length));
```
hbdeshmukh: Basic question - why are the characters allocated on heap?

hbdeshmukh:
Hi @zuyu 
Finished first round of review. Let me know once you finish addressing them.

query_execution/Foreman.cpp
zuyu: We could do so for sure, and it is better to refactor it in another separate PR, right?

query_execution/QueryExecutionMessages.proto
zuyu: Sorry, I did not get your point about why need to separate the completion message for `WorkOrder` and `RebuildWorkOrder`.

If I understand correctly, what you mentioned here are all about how to serialize different `WorkOrder`s.

query_execution/QueryExecutionMessages.proto
zuyu: Depend on what kind of info to add. If it is related to this message, add the field here. Otherwise, create a new message and then a new message type in `QueryExecutionTypedefs.hpp`.

query_execution/QueryExecutionMessages.proto
zuyu: Thanks!

relational_operators/DeleteOperator.cpp
zuyu: A couple of reasons:

 1. The length of `proto` is not fixed,
 2. This implementation is used in other places in the codebase as well.

zuyu:
@hbdeshmukh I've addressed your comments. Please take another look. Thanks!

query_execution/Foreman.cpp
hbdeshmukh: I am not saying that we need it, I am trying to understand if we can do without a proto for this message type. My guess was that since such messages are only sent to and from the director node, we might not need it.

zuyu: The answer is yes and no.

For yes part, so far we do not create a proto for the `feedback message`, and it works fine.

On the other side, for now every discrete `WorkOrder feedback message` is a deserialized proto (e.g., in `TextScanOp` and `Sort`-related operators), and it is up to the operators to interpret the underlying meanings of the feedback message in `receiveFeedbackMessage`.

hbdeshmukh: Are you saying it works fine without the proto even on a multi-process scenario? That's the part I am trying to understand.

query_execution/QueryExecutionMessages.proto
hbdeshmukh: Let us assume that the scheduler wants to know about where a particular hash table is "hot" (i.e. in cache) right now. The feedback from ``BuildHash`` or ``HashJoin`` work order is a very good source for such information. Such kind of information is not conveyed through RebuildWorkOrder completion messages. Therefore it makes sense to treat the two message types separately.

query_execution/Foreman.cpp
zuyu: Yes, because the message content is a serialized proto, not the message itself.

query_execution/QueryExecutionMessages.proto
zuyu: I get that the scheduler may want to know some run-time info, and the `worker_id` (or the `shiftboss_id` in near future) in `WorkOrderCompletionMessage `, is good enough to know where the hash table lives.

Even if I miss some info requested by the scheduler here, we could create another PR to improve, right? Note that the goal of this PR is just to replace `ForemanMessage`s with `proto`.

hbdeshmukh: Alright, you can add a TODO about it.

relational_operators/DeleteOperator.cpp
pateljm: @zuyu I think in general, we shouldn't have a response that the reason is because it was done elsewhere in the code, as going forward we want to identify issues and not only not repeat them, but also fix existing issues. 

BTW, just right after this we do malloc for proto_bytes, so there is an inconsistency in what we allocated on the stack vs the heap.

In general, its a good idea to not use too much stack memory. But, have to be careful to delete allocations on the heap, especially when there are errors/early return.

query_execution/QueryExecutionMessages.proto
zuyu: Could you please give me more info about TODOs? I think it already provides the enough info for the scheduler for now.

relational_operators/DeleteOperator.cpp
zuyu: @pateljm Well, I should have pointed out that the Google protobuf example used to have the following code, until using `SerializeToOstream` instead. But `SerializeToArray` fits our case, as `SerializeToOstream` introduces an extra overhead for converting `std::string` to a c-like string.

```
int size = address_book.ByteSize(); 
void *buffer = malloc(size);
address_book.SerializeToArray(buffer, size);
```

Using both the heap and stack does not mean we are not **consistent** for memory allocation, because we use the heap for a variable-length, out-of-band data (there is no safe and efficient way to allocate them on the stack in c++), while the stack for the rest.

And yes, I release the heap memory immediately after use in the same function.

query_execution/QueryExecutionMessages.proto
hbdeshmukh: The TODO should mention that the message class will be replaced by two separate message classes - ``NormalWorkOrderCompletionMessage`` and ``RebuildWorkOrderCompletionMessage``.

relational_operators/DeleteOperator.cpp
hbdeshmukh: I agree with @zuyu's explanation.  It might be better if we add this explanation as a comment (in one of the several places where the pattern is used), so that the rationale for using heap memory is clear to the reader.

hbdeshmukh:
The PR is good to go except a couple of remarks about adding comments. Unless @pateljm has any more concerns, we can merge the PR after the comments are addressed.

pateljm:
LTGM. Please close after the remarks about the comments have been addressed. Thanks!

hbdeshmukh:
Merging.

PR #26, opened 2016-02-09
Change relation_id to CatalogRelation& in relational operators

jianqiao:
This modification is the first step of isolating temporary relations.

Note that in a bigger plan, we will decouple the Quickstep frontend from its backend, where the frontend only acquires a readonly copy of the catalog. So temporary relations will not be allowed to be directly  inserted into the catalog by the frontend query planner. Instead, we will create a temporary CatalogDatabase that stores the temporary relations, then send the temporary database to the backend to be merged into the stored database. 

So for a relational operator, at the time point when serialization/deserialization happens, a relation may either from the stored database, or from the temporary database. One single "relation_id" won't be sufficient to differentiate these two situations. And a quick solution is to use the reference CatalogRelation&.


query_optimizer/ExecutionGenerator.cpp
```diff
@@ -714,7 +714,7 @@ void ExecutionGenerator::convertCopyFrom(
     const P::CopyFromPtr &physical_plan) {
   // CopyFrom is converted to a TextScan and a SaveBlocks.
 
-  const relation_id output_relation_id = physical_plan->catalog_relation()->getID();
+  const CatalogRelation *output_relation = physical_plan->catalog_relation();
```
pateljm: Why return a pointer, instead of an object by reference?

jianqiao: The original design of many classes have their getter methods return pointers other than references, e.g. the "physical::CopyFrom" class here. Perhaps at this moment we can just stick to the original design, and put an issue on uniforming the styles.

pateljm: Sounds good! LTGM - Mereging.

zuyu:
@pateljm Sorry for the late comment after merging.

If we are moving the master `catalog` as a service, this PR does not make sense, as operators now have to keep a copy of `CatalogRelation`s, instead of the simple `relation id`s.

relational_operators/DropTableOperator.hpp
```diff
@@ -48,9 +48,9 @@ class DropTableOperator : public RelationalOperator {
    * @param only_drop_blocks If true, only drop the blocks belonging to
    *        relation, but leave relation in the database.
    **/
-  explicit DropTableOperator(const relation_id rel_id,
+  explicit DropTableOperator(const CatalogRelation &relation,
```
zuyu: No need to pass `CatalogRelation` to this operator. The original argument `relation_id` is good enough to generate the `work order`.

jianqiao: A quick answer is that frontend can't generate the relation_id - that happens in the backend.

To give more details, here I explain a bit more about the big picture. In our plan, the frontend will receive a readonly copy of the catalog, then performs query planning / optimization and generates the query plan. The query plan will be executed by the backend.

So, the frontend's output will contain four messages that are encoded into protobuf:
(1) The query context.
(2) The DAG of relational operators.
(3) The temporary catalog relations.
(4) A reference to the result relation.

The backend, upon receiving the query plan, will do the follows:
(1) First, reconstruct query context, DAG of relational operators, temporary relations and the reference to the result relation.
(2) Insert temporary relations into the catalog database.
(3) Execute the DAG of relational operators. Note that temporary relations are dropped during execution.
(4) Return results, then drop the (temp) result relation.

About why relation_id is not sufficient. Here is one example.

Suppose currently the catalog database has two stored relations R and S, which have relation_id 0 and 1. Now we are to execute a complex query Q1 so three temporary relations A, B, C are generated. The question is, what is the relation_id of A, B, C? At first glance, we might use 2, 3, 4. But it is actually *resolving relation_id at the frontend*, which endures the loss of isolation.

That is, suppose just after Q1 is planned but its query plan has not yet arrived at the backend, some query Q2 is also planned that occupies catalog database slot 2, 3 and gets executed first. Then when Q1 arrives at the backend, it will see a confliction, and it becomes nasty to modify the relation_id's at this time point - you have to dive into every corner of query context and relational operators to inject code to adjust the relation_id's.

One thing is true is that from the pure backend perspective, relation_id is sufficient. But currently the data structures (especially relational operators) are shared by the both the frontend and backend. So use CatalogRelation& to replace relation_id is the most convenient way. Note that the pure backend part -- WorkOrders of relational operators -- are not affected. They'e still using relation_id's.

PR #29, opened 2016-02-10
Update CMakeLists.txt

pateljm:
Fix the version check as it breaks with flex 2.5.37. This may have to change again in the future.


hbdeshmukh:
Looks good.

PR #30, opened 2016-02-10
Add support for generic key-value based specifiers for SQL statements

saketj:
This PR adds support for generic key-value based parsing in the SQL Parser that will allow SQL statements to have a wide range of property specifiers. Currently, these key-value based specifiers support three generic formats for key-values:
* string_key , string_value
* string_key ,  list_of_string_values
* string_key , literal_value

As an illustration, consider a use-case for `CREATE INDEX` statement. We want to provide for a `CREATE INDEX` syntax that will allow users to preferentially tune index parameters based on the index type. A verbose `CREATE INDEX` statement might look like:

> CREATE INDEX _index_name_ ON _table_name_ 
USING BLOOMFILTER ( size_in_bytes 1000, false_positivity_range (0.4, 0.6), num_hashes 20 )

The optional parameters specified in the parentheses are instances of generic key value pairs. What constitutes a valid or invalid key-value pair is left on the statement resolver to decide. 
Hence, now individual SQL statements like `CREATE INDEX`, `CREATE TABLE`, etc.  can decide on the number and the kind of parameters/options they want to support, without having to create a whole bunch of lexical tokens for all possible configurations.


saketj:
@cramja Hey Marc, I have finished adding the grammar rules for the key value pairs and have implemented it for `CREATE INDEX` statement now. Feel free to edit/comment on any part and see if the grammar rules work for block properties, as well.

cramja:
Hi Saket, I looked over the changes to the parser, and I think the grammar part looks good-like what we had on the board, though I have not tested it. One thing I would suggest is that it'll make your life easier if you define the index type as ParseString and then do string comparison in the resolver to match it to the correct enum for entry into the proto message. 

Also, the cmake lists are a little off but I hesitated to fix them because your changes might make that irrelevant. We can work it out this afternoon and get this thing merged!

saketj:
@cramja Thanks Marc for the comments. I will modify the Index type to be a ParseString. Sure, lets work this afternoon to put this thing a good shape that it can be merged. I wonder why all the Travis tests have failed. I looked up the logs and seems to me that the process failed, even after passing all the tests. This seems little strange, I will have a detailed look on it.

cramja:
It failed with, `The command "./validate_cmakelists.py" exited with 1.` because some of the cmake files in parser were missing dependencies.

cramja:
As we discussed yesterday, the changes look good. I will now merge and start changing BlockProperties to use this PR.

PR #31, opened 2016-02-10
Change default storage layout to column store whenever the schema permits

pateljm:
Change the default layout to column store. This will likely slow down updates but likely speed up analytic queries. How much? Here are some numbers (Release build) on the SSB 10GB benchmark on a single socket with 10 cores. 

Query | Before  | After 
:--------|--------:|--------:|
Load the big `lineitem` table | 123.7 seconds | 207.6 seconds 
Q1 | 3.0 seconds | 0.96 seconds  

Need to work on speeding up insertion into (compressed) column stores. A future PR. 

PS: The following command was use to start quickstep: `./quickstep_cli_shell -printing_enabled=false -buffer_pool_slots=10240 -parallelize_load -num_workers=10 -worker_affinities="0,4,8,12,16,20,24,28,32,36"`


hbdeshmukh:
Hi @pateljm This is very impressive!

pateljm:
Thanks @hbdeshmukh for all your help this morning on getting the right parameters. That was crucial in getting these numbers!

zuyu:
@pateljm This is awesome!

One minor grammar issue: capitalize the first letter in all the comments, as we treat them as English sentences.

cramja:
The code looks good.

There are a few corner cases to consider.

* If the user creates a table where the first attribute is an attribute which does not have a standard comparator, then setting the SORT_ID to that attribute will make an invalid description.
* Similarly, there are some attribute types which cannot be compressed in the relation, the resulting description will be invalid.

For more information on this, the [DescriptionIsValid](https://github.com/pivotalsoftware/quickstep/blob/master/storage/CompressedColumnStoreTupleStorageSubBlock.cpp#L94) method of CompressedColumnStore will show you what conditions must be met while creating a block.

* Finally, setting the default block to column store removes the ability to ad-hoc add tuples to the store. (What if we could keep a rowstore block around for ad-hoc insertions? Is the this even a useful feature?)

cramja:
I should add that a quick fix to the invalid description corner case if to create the description. We can then test if it is valid, and if not, use the old method of using a rowstore.

pateljm:
Excellent points @cramja on catching the case when the first attribute does not have a comparator. Let me mull over the other point you make about ad hoc tuples. 

On a different note, I'm so glad to see you ramping up so quickly and your attention to detail. Thanks!

pateljm:
Ok - had a nice chat with @cramja and would make sense to not merge this PR even after reworking. Better idea is to wait for the block properties work to be completed, and then build on top of that. Deleting this PR next. Thanks Marc!

PR #33, opened 2016-02-12
Overrode the doneFeedingInputBlocks method for HashJoinOperator

hbdeshmukh:
This PR fixes the correctness issue of some of the SSB query results. It seems that the base class method ``doneFeedingInputBlocks()`` of ``RelationalOperators``, was not overridden in ``HashJoinOperator``. The doxygen of ``doneFeedingInputBlocks()`` does mention that it needs to be overridden for exactly such scenarios, but the overriding was not done for ``HashJoinOperator``. 

In addition to that, I also added another condition before fetching normal WorkOrders for an operator (thanks to Qiang for discovering these issues).

I tried all the SSB queries with 10 threads and they produce consistent results now.

### DIFF
--- a/relational_operators/HashJoinOperator.hpp
+++ b/relational_operators/HashJoinOperator.hpp
@@ -142,6 +142,9 @@ class HashJoinOperator : public RelationalOperator {
   }
 
   void doneFeedingInputBlocks(const relation_id rel_id) override {
+    // The HashJoinOperator depends on BuildHashOperator too, but it
+    // should ignore a doneFeedingInputBlocks() message that comes
+    // after completion of BuildHashOperator. Therefore we need this check.
     if (probe_relation_.getID() == rel_id) {
       done_feeding_input_relation_ = true;
     }
--- a/relational_operators/HashJoinOperator.hpp
+++ b/relational_operators/HashJoinOperator.hpp
@@ -141,6 +141,12 @@ class HashJoinOperator : public RelationalOperator {
     return output_relation_.getID();
   }
 
+  void doneFeedingInputBlocks(const relation_id rel_id) override {
+    if (probe_relation_.getID() == rel_id) {
+      done_feeding_input_relation_ = true;
+    }
+  }
+
  private:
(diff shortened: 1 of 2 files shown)


relational_operators/HashJoinOperator.hpp
```diff
@@ -141,6 +141,12 @@ class HashJoinOperator : public RelationalOperator {
     return output_relation_.getID();
   }
 
+  void doneFeedingInputBlocks(const relation_id rel_id) override {
```
pateljm: Nice work catching this! Could you add a comment here to let others know why the override is needed. Thanks!

hbdeshmukh: Thanks @pateljm! I added the clarifying comment. Please take another look.

zuyu: If I understand correctly: this fix ensures that `HashJoinOperator` is available for scheduling unless `BuildHashOperator` is done.

More details:
`Foreman::markOperatorFinished` calls `doneFeedingInputBlocks` for all dependent operators. `HashJoinOperator` depends on two operators: `BuildHashOperator` on the build relation and some operator (called `ProbeOp`) on the probe one. Thus, without this fix, once `ProbeOp` has finished execution before `BuildHashOperator` is done, it also calls `doneFeedingInputBlocks`, but incorrectly triggers the semantics that `HashJoinOperator` has all blobs of hash table on the build relation.

hbdeshmukh: Not exactly. This fix is relevant when there's a HashJoin is a consumer of the output of some dependent operator (most likely a select).  The Foreman needs to notify HashJoin::doneFeedingInputBlocks only when the select stops pipelining blocks. 

As you know that BuildHash is also a dependent operator of HashJoin, the HashJoin is also notified of doneFeedingInputBlocks when BuildHash is completed. But HashJoin should ignore that message.

pateljm:
LTGM. Merging.

PR #36, opened 2016-02-16
Serialize Work Order as Proto.

zuyu:
This is a single large PR that serialize all `work order`s as proto. It is a part of moving to the distributed database.

`Foreman` creates `work order` protos in `WorkOrdersContainer`s, and then reconstructs `work order` from protos before dispatching to `worker`s. Note that in the distributed `Quickstep`, the `work order` reconstruction part happens in `Shiftboss`.

@jianqiao Please take a look for my changes on operators, as it might affect your work on separating the  front-end and the back-end. Thanks!


pateljm:
@zuyu Can we add an abstraction around to short-circuit the creation of WO using protobufs? In a single-node case, we should bypass. 

Also, in the distributed case, wouldn't the Foreman in the Director simply create the WOs rather than go through the protobuf way? protobufs are only needed when sending across processes (Director to the Worker). We should not add the protobuf overhead unnecessarily.

zuyu:
@pateljm Yes, I could use macros (e.g., `QUICKSTEP_DISTRIBUTED`) to differentiate the ways to generate `work order`s in the single-node and the distributed cases. But I don't have a better approach to do so, as both cases have only minor differences in some implementations in the function level (e.g., `getAllWorkOrders` for each operator).

In the distributed case, serializing `work order`s is required, as components used to create actual `work order`s (e.g., `StorageManager` and `QueryContext` in `worker`s) are not accessible to `Foreman` in `Director`.

On the other hand, I could compare the overhead of w/ and w/o an`work order` serializations. If it is minor, we may keep the serializations.

And as a follow-up, we should better to use `Shiftboss` in the distributed case only.

Btw, I am debugging a test failure related to serializing `SortMergeWorkOrder`. I did not encounter this issue when using `protobuf 3` beta version.

PR #40, opened 2016-02-19
Fixed a bug in message sending check in TextScanOperator.

hbdeshmukh:
Changed a comparison in GLOG CHECK macro that was left unattended in the previous merge. Thank @zuyu for pointing it out.


zuyu:
LGTM.

Issue #32, opened 2016-02-11
How to display proto message in an optimizer output?


cramja:
**Background**
In an update to the parser/optimizer, we are supporting setting physical properties of blocks. This is done by taking parsed SQL and translating it into a proto message in the the resolver. The proto message is passed down to the operator which finally creates the relation. 

**?**
This question is about the the `getFieldStringItems` method. How should one display a proto message in the field string?

1) One solution I thought of was simply to have it be an attribute with a boolean value, so that the print out would look something like
```
CreateTableStatement[name=foo]
+-HasProto[True]

or
CreateTableStatement[name=foo]
+-HasProto[False]
```

This has the downside of not displaying the details of the proto message.

2) Another solution is to use [Google's protobuf dumping program](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.text_format) to dump the entire message in a field. However, this might make the message look messy.

3) Another solution is to not include any extra output for block properties.

___
I prefer solution (2) because it tells all so to speak. But we might be in favor of the simplicity of (1).

Thoughts?
Marc

zuyu:
Before printing the `proto`, I have a few basic questions:
 1. why the `proto` is the grand truth for the block properties? Do you directly construct the `proto` from the parser results, rather than saving the block property info in some local variables and serializing as `proto` later?
 1. what fields contain in the `proto`?
 1. why need to dump the entire message, rather than display each field one-by-one?

Thanks!

cramja:
To answer your questions:

1) The proto is a convenient and lightweight way to take the 'messy' user input and place it in a machine format. Saving it to local variables would likely mean making logical/physical classes to encapsulate the BlockProperties, which seems cumbersome and redundant when the proto message can hold the needed information.

2) The proto is a [TupleStorageSubBlockDescription](https://github.com/pivotalsoftware/quickstep/blob/master/storage/StorageBlockLayout.proto#L21), with 4 fields describing the physical properties of the block (type-rowstore/columnstore, sort column, etc)

3) Dumping the entire message is fine.

I will go ahead and use the dump everything solution. @zuyu if you agree, go ahead and close this request or we can keep the conversation going.

pateljm:
Hi @cramja: I like solution #2. Go for it. Closing this issue too.

zuyu:
I agree that we need to print out the `proto`, but from the implementation's perspective, not in the way you mentioned to use proto APIs to dump the entire message (the output does not follow our print format).

[...]

PR #41, opened 2016-02-19
Block properties

cramja:
# Changes for Addition of BLOCKPROPERTIES clause

## What is it?
This PR builds off of the recent Parse-key-value PR. Now we are using these additions to the parser to parse a different construct, the BLOCKPROPERTIES clause. BLOCKPROPERTIES is a set of key value pairs describing the physical layout of a block.

## High level changes

1) Parse constructs in Parser files.
2) Resolver additions - the resolution of block properties is challenging. This method covers many corner cases.
3) Tests for resolver. Each of the error cases caught by the resolver are tested.
4) Proto representation in Optimizer. I created a new generic key-value Optimizer node type as well as function for loading a TupleStorageBlockDescription into this representation.

## How does it work?
To test it, run the test suite. 

To use the SQL syntax:

```SQL

CREATE TABLE foo (attr int, attr2 int) WITH BLOCKPROPERTIES
(TYPE compressed_columnstore, SORT attr, COMPRESS ALL, BLOCKSIZE 10);

```

`TYPE` values: \[rowstore, split_rowstore, compressed_rowstore, columnstore, compressed_columnstore\]
`SORT` value can be an attribute name specified in the attribute list.
`COMPRESS` values are attribute names in a list format, `(attr1, attr2)` or the keyword `ALL`.
`BLOCKSIZE` is a mega-byte sized slot size for block. Note that this is converted in the resolver to slots. A LOG message is printed out describing the conversion (10 megabytes translates into 4 2mb-sized slots).

## Question for reviewers:
* What is a good limit for BLOCKSIZE. Certainly 1 is a good lower bound, is there a reasonable upper bound for megabyte size?
* 

## Follow changes.
This feature is incomplete. Still needed are

* Printing of the physical level optimizer objects containing proto messages.
* Changes for worker to add properties to 

These will come as a separate PR, or if the reviewers think they should be appended to this on, I can add them.

## EDIT 2/23
Completed this feature. Everything in the 'next PR' comment has been added.


cramja:
@saketj Could you comment on the method of [printing proto messages in the Optimizer](https://github.com/pivotalsoftware/quickstep/pull/41/files#diff-d0f00d4aa0ac6f9db0f8ab5390f02df2R133)? I think you could reuse this functionality in your Index feature and so it would be good to have that looked over by both of us.

parser/ParseStatement.hpp
```diff
@@ -148,11 +162,20 @@ class ParseStatementCreateTable : public ParseStatement {
     for (const ParseAttributeDefinition& attribute_definition : *attribute_definition_list_) {
       container_child_fields->back().push_back(&attribute_definition);
     }
+
+    if (opt_block_properties_) {
```
saketj: Shouldn't the value pointed by opt_block_properties_ be checked for nullptr, instead of opt_block_properties_ pointer itself?

parser/SqlParser.ypp
```diff
@@ -239,6 +240,7 @@ void NotSupported(const YYLTYPE *location, yyscan_t yyscanner, const std::string
 %token TOKEN_FROM;
 %token TOKEN_FULL;
 %token TOKEN_GROUP;
+%token TOKEN_GB;
```
saketj: Is this TOKEN_GB still being used in the Parser grammar? If not, then maybe we can remove it.

parser/SqlParser.ypp
```diff
@@ -252,6 +254,7 @@ void NotSupported(const YYLTYPE *location, yyscan_t yyscanner, const std::string
 %token TOKEN_LEFT;
 %token TOKEN_LIMIT;
 %token TOKEN_LONG;
+%token TOKEN_MB;
```
saketj: Same as above.

parser/SqlParser.ypp
```diff
@@ -834,23 +846,34 @@ key_value:
...
   any_name '(' name_commalist ')' {
     $$ = new quickstep::ParseKeyStringList(@1.first_line, @1.first_column, $1, $3);
   };
 
-key_literal_value:
-  any_name literal_value {
-    $$ = new quickstep::ParseKeyLiteralValue(@1.first_line, @1.first_column, $1, $2);
+key_integer_value:
+  any_name TOKEN_UNSIGNED_NUMVAL {
+    if ($2->float_like()) {
```
saketj: I was wondering whether if we can specify the requirement for integer values using the grammar rules itself. Then, we will not need an if-clause statement altogether.

query_optimizer/LogicalGenerator.cpp
```diff
@@ -43,6 +45,7 @@ LogicalGenerator::~LogicalGenerator() {}
 L::LogicalPtr LogicalGenerator::generatePlan(
     const ParseStatement &parse_statement) {
   resolver::Resolver resolver(optimizer_context_);
+  DVLOG(4) << "Parse tree:\n" << parse_statement.toString();
```
saketj: I guess, this maybe a residual print statement from debugging. Wasn't earlier, unless we want to print the parse tree now too.

query_optimizer/OptimizerTree.hpp
```diff
@@ -118,6 +122,169 @@ class OptimizerTree : public OptimizerTreeBase {
...
 
+/**
+ * @brief A helper class for printing Protobuf messages in a simple flat-format.
+ * @details Holds key-value pairs of a proto message. Key values are added as strings.
+ *          To use, create a representation and then add properties:
+ *             auto foo = new OptimizerProtoRepresentation<OptimizerPtrType>();
+ *             foo.addProperty("friendly_name", proto.GetExtension(ProtoType::friendly_name));
+ */
+template<class TreeNodeType>
+class OptimizerProtoRepresentation : public OptimizerTreeBase {
```
saketj: @cramja Hey Marc, I like this idea of helper class to generate string values corresponding to the Protobuf messages. I think this will be useful for when passing the index block properties as well.

query_optimizer/OptimizerTree.hpp
```diff
@@ -118,6 +122,169 @@ class OptimizerTree : public OptimizerTreeBase {
...
+                           std::vector<TreeNodeType> *non_container_child_fields,
+                           std::vector<std::string> *container_child_field_names,
+                           std::vector<std::vector<TreeNodeType>> *container_child_fields) const {
+    for (auto property : properties_) {
+      non_container_child_field_names->push_back("Property");
+      non_container_child_fields->push_back(property);
+    }
+  }
+  // A list of managed properties.
+  std::vector<std::shared_ptr<const OptimizerProtoPropertyRepresentation> > properties_;
```
saketj: I am little confused between the use of std::vector and std::unique_ptr<PtrList> to represent list of pointers. Marc, could you please explain when to use one over the other?

cramja: Yes, I will add a comment. In a nutshell, when the vector goes out of scope, so do its contained unique ptrs. All of the unique pointers will call the dtors of their managed objects:

```c++
#include <iostream>
#include <string>
#include <memory>
#include <vector>

using namespace std;

class B {
 public:
  B(const char* s) : name(s) {}
  ~B() {
      cout << name << " has been deleted.\n";
  }
  string name; 
};

int main() {
	{
	    vector<unique_ptr<B> > vec;
	    unique_ptr<B> pb1(new B("bob"));
	    vec.push_back(std::move(pb1));
	    unique_ptr<B> pb2(new B("fred"));
	    vec.push_back(std::move(pb2));
	    unique_ptr<B> pb3(new B("chuck"));
	    vec.push_back(std::move(pb3));
	}
	return 0;
}
```

Will print:
```
bob has been deleted.
fred has been deleted.
chuck has been deleted.
```

parser/SqlParser.ypp
cramja: I think that would require coming up with new tokens, one for ints and one for floats. Maybe this could be addressed in another PR.

query_optimizer/OptimizerTree.hpp
saketj: Ah, I see. The vector calls the destructors on all of its stored objects, which are unique_ptr in this case. And the unique pointer calls the destructors to the pointed objects.

zuyu: No, not at all.

* `PtrList` is a wrapper for storing pointers in `std::list` (a linked list) with a scope deleter.
* `vector` is an array-based container.

saketj: @zuyu Correct, thanks Zuyu. So, apart from the difference in the underlying container, I guess that both provide similar guarantee for the lifecycle management for the objects that they contain. So, the usage depends on the choice between an array vs linked list mostly.

jianqiao: @saketj I think one difference is that PtrList (as well as PtrVector) can do ownership transitions. You will see that many classes in Quickstep have this macro at the bottom:
`DISALLOW_COPY_AND_ASSIGN(ClassName)`

This will disallow copy assignment (to avoid potential expensive copy assignment that we are not aware of), so you cannot push an existing object into a vector, say
```
A a; // Suppose A has the DISALLOW_COPY_AND_ASSIGN(A) macro
vector<A> a_vec;
a_vec.push_back(a); // Compile error
a_vec.push_back(std::move(a)); // Okay if the move constructor A(A &&other) is
                               // implemented, but that's somehow tedious.
```

So what we usually do is:
```
A *a = new A();
PtrVector<A> a_vec;
a_vec.push_back(a); // Take ownership of a.
```

saketj: @jianqiao Thanks Jianqiao for the detailed comment. This explains a lot of stuff now regarding their usage.

cramja: After reading these comments and thinking about how the `OptimizerProtoRepresentation` class will be used, I realize that the vector of shared pointers in the class:
```
// A list of managed properties.
std::vector<std::shared_ptr<const OptimizerProtoPropertyRepresentation> > properties_;
```
should actually be a vector of `unique_ptrs`. This is because  `OptimizerProtoRepresentation`, the parent node in the tree representation of the Proto message will be shared across the Logical/physical CreateTable instances. Its children, the `OptimizerProtoPropertyRepresentation` will be managed by their parent, so, as long as there is a shared_ptr handling the shared parent, the children can be safely handled in unique ptrs owned by the parent.

Even better, I'll reuse the PtrVector class. Thank you @jianqiao for pointing this out.

cramja:
@pateljm Made BLOCKSIZE -> BLOCKSIZEMB change.

As for the `help` syntax, I am imagining a manpages-style format for the messages that would be returned:

> Finally, can we find a way to print a help message to explain the details of these parameters. Ideally, we should be able to type in something like:

``` SQL
> help CREATE TABLE
Creates new relation in the database. Optionally parameterized with physical properties.

CREATE TABLE (attribute_name TYPE, ...) 
   [WITH BLOCKPROPERTIES (PROPERTY value, ...)];

attribute_name    A name for a column in the relation.
TYPE              A valid attribute type. (see help TYPE)
[BLOCKPROPERTIES] Optional storage parameters. (see help BLOCKPROPERTIES)
```

Is this sort of what you had in mind?

pateljm:
@cramja The help is exactly what I had in mind!

cramja:
@saketj I finished this PR (top to bottom). Can we do a code review sometime soon so we can merge it?

cramja:
@zuyu By the way, to followed up with your comment in the [issue](https://github.com/pivotalsoftware/quickstep/issues/32#issuecomment-183580886) I filed a few days ago, the solution I implemented is here. [It's a class](https://github.com/pivotalsoftware/quickstep/pull/41/files#diff-d0f00d4aa0ac6f9db0f8ab5390f02df2R137) specifically for displaying proto messages. Making a class was necessary because the proto key-value pairs (not actually how the proto is structured but close enough for us) needed to extend an OptimizerTree base class.

cramja:
Instead of using inline fields, I opted to use a subtree so that it was clear that this was a proto property. For example, the physical tree output for a CREATE statement is something like:
```
+-plan=CreateTable[relation=foo]
| +-block_properties=ProtoDescription
| | +-Property=ProtoProperty[Property=blocktype,Value=compressed_columnstore]
| | +-Property=ProtoProperty[Property=sort,Value=0]
| | +-Property=ProtoProperty[Property=compress,Value=0]
| | +-Property=ProtoProperty[Property=slots,Value=2]
| +-attributes=
|   +-AttributeReference[id=0,name=col1,relation=foo,type=Int]
+-output_attributes=
  +-AttributeReference[id=0,name=col1,relation=foo,type=Int]
```

parser/SqlParser.ypp
pateljm: @cramja We can file an issue for the new tokens, so that we can consider it as a separate PR.

pateljm:
LTGM. Merging.

parser/ParseBlockProperties.hpp
```diff
@@ -0,0 +1,254 @@
...
+      seen_keys.insert(lower_key);
+    }
+    return nullptr;
+  }
+
+  /**
+   * @brief Returns the first invalid key contained in the key-value list.
+   * @details An invalid key-value is a key-value with a key whose case-insensitive
+   *          string does not match one of the valid key names.
+   *
```
zuyu: @cramja There are trailing spaces in this line, and other places in this file. Please fix in another PR.

PR #42, opened 2016-02-22
CREATE INDEX resolver and operator implementation

saketj:
This is the second PR in continuation with the series of PRs that aim to add `CREATE INDEX` functionality in Quickstep. It builds on the previously added support for `CREATE INDEX` syntax in the parser and implements the resolver and operator for the `CREATE INDEX` statement. 

> With this PR, it is now possible to create placeholder indices on a relation that get saved in the protobuf representation of the catalog. Of course, using a duplicate index name will throw an error. 

This PR touches a wide stretch of codebase from query resolver to a new operator implementation. Specifically, it achieves the following:

- The resolver now understands how to use a parsed CreateIndex statement to generate a logical plan.
- The PhysicalGenerator subsequently uses the logical plan to generate a physical plan.
- The ExecutionGenerator uses the physical plan to create a work order that relies on a new Create Index operator. For now, this operator is supposed to simply modify the CatalogRelation to store the name of the index. This operator will be extended in the upcoming PR. 
- In the work order that is generated and finally executed by the Foreman, the CatalogRelation's protobuf representation is modified to store the index names as repeated strings. 

Goals for the next PR:
The placeholder index representations will be replaced in the upcoming PR by their rightful data structure representation. The Create Index operator will become more dense to represent the index properties in  protobuf format and use them to modify the StorageBlockLayout. The modified StorageBlockLayout can then be used to create blocks with indices.


cramja:
Looks like a good outline. There was a simple grammar concept which you added that I liked which was to create tokens for index type and then have a grammar to convert them into the correct string which saves the hassle of checking in the Resolver.

I am curious how you plan to create the mapping between index name and storage block which they will reside on. Maybe we could chat before the meeting?

saketj:
Thanks Marc for the comments. I will fix the indentation issues. Sure, lets chat before the meeting. I have some ideas, maybe I can run them with you.

cramja:
@saketj and I met to discuss briefly the development process for this PR. We decided we will merge this one and open a second PR to complete the feature's 'stubs' present in this PR.

cramja:
So Saket, looks good to merge if the tests pass.

cramja:
I should also add that one of the things I learned from the BlockProperties feature is that the taking in a list of key-values in your parser construct is sort of cumbersome to handle in the resolver.

You may consider handling some of the validation code for the properties inside of the ParseCreateStatement class or even in the parser.

PR #43, opened 2016-02-23
Refactored DropTableOperator.

zuyu:
@pateljm @jianqiao @hbdeshmukh 

This PR refactored `DropTableOperator`, which contains two tasks:
 1. Drop blocks using `StorageManager`, and
 1. Make either change in `CatalogDatabase`:
   * drop the relation, or
   * clear the blocks, but keep the relation.

In the distributed version, `Task 1` might be executed in every `Worker` where its `StorageManager` has loaded some block from the being-dropped relation, while `Task 2` should be executed only once (otherwise, the exception `RelationIdNotFound` will be thrown).

Thus, this PR enables:
 * `Task 1` wraps as `DropTableWorkOrder` to execute in some `Worker`.
 * `Task 2` happens once when `Foreman` fetches `WorkOrder`s for this operator.

Note that there are some extra work to do in future so that `DropTableOperator` could work in the distributed version with multiple workers. For example, `Foreman` should generate multiple `WorkOrder`s, instead of one for now.

### DIFF
--- a/relational_operators/DropTableOperator.cpp
+++ b/relational_operators/DropTableOperator.cpp
@@ -17,10 +17,12 @@
 
 #include "relational_operators/DropTableOperator.hpp"
 
+#include <utility>
 #include <vector>
 
 #include "catalog/CatalogDatabase.hpp"
 #include "catalog/CatalogRelation.hpp"
+#include "catalog/CatalogTypedefs.hpp"
 #include "query_execution/WorkOrdersContainer.hpp"
 #include "storage/StorageBlockInfo.hpp"
 #include "storage/StorageManager.hpp"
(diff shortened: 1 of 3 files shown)


relational_operators/DropTableOperator.cpp
```diff
@@ -32,9 +34,21 @@ namespace quickstep {
 bool DropTableOperator::getAllWorkOrders(WorkOrdersContainer *container) {
   if (blocking_dependencies_met_ && !work_generated_) {
     work_generated_ = true;
+
+    std::vector<block_id> relation_blocks(relation_.getBlocksSnapshot());
+
+    // DropTableWorkOrder only drops blocks, if any.
     container->addNormalWorkOrder(
-        new DropTableWorkOrder(relation_.getID(), only_drop_blocks_),
+        new DropTableWorkOrder(std::move(relation_blocks)),
```
hbdeshmukh: If I remember correctly, this is a pessimizing move. You can pass just a const reference to ``relation_blocks`` and move the contents of the vector in the ``DropTableWorkOrder`` constructor.

zuyu: I don't think so. Passing a const reference would result in a deep copy. In other words, there will be two copies of the contents, instead of one in the destination.

relational_operators/DropTableOperator.cpp
```diff
@@ -32,9 +34,21 @@ namespace quickstep {
...
         op_index_);
+
+    // TODO(zuyu): move the following code to a better place.
```
hbdeshmukh: Do you mean move to a different class? What's the reason?

zuyu: These code changes `Catalog`, and seems not related to `getAllWorkOrders`.

Similar to `CreateTableOperator::getAllWorkOrders`, which executes some tasks, instead of generating work orders.

pateljm:
LTGM. Merging.

relational_operators/DropTableOperator.cpp
hbdeshmukh: No, it won't result in a deep copy. 

``new DropTableWorkOrder(relation_blocks)`` -> Here ``relation_blocks`` is passed as a const reference, so no copy has happened so far. 

In the initialization of blocks_ in the ``DropTableWorkOrder``, you are doing a move. Therefore deep copy is out of the question any way. 
```
explicit DropTableWorkOrder(std::vector<block_id> &&blocks)
      : blocks_(std::move(blocks)) {}
```

hbdeshmukh: Sorry, I take back my argument. I tried a simple program and the const reference case I mentioned doesn't work as I had said.

PR #44, opened 2016-02-24
Refactor passing tmb info to work orders.

zuyu:
This PR uses `RelationalOperator::getAllWorkOrders` to pass TMB-related info to WorkOrders, instead of indirectly passing from the optimizer and then operators..

Note that there will be a similar follow-up PR to refactor passing the Foreman client id for `InsertDestination`s.


pateljm:
LTGM. Merging.

zuyu:
@pateljm Thanks! I've solved the conflicts, and let's wait for the CI results.

pateljm:
Merging.

PR #39, opened 2016-02-18
Table generator functions

jianqiao:
This PR adds the features that support table generator functions. A `GeneratorFunction` is a pre-defined function which populates a table. One example is the `generate_series` function, which in the simplest case takes two integer parameters `(start, end)` and populates a single-column table with values `start, start+1, ..., end`.

Currently only `generate_series` is implemented. But new generator functions can be added conveniently by just subclassing `GeneratorFunction` and `GeneratorFunctionHandle`.

@pateljm Hi Jignesh, would you help suggest a person to review the code? Thanks!

**Overview of main changes:**

- `parser/SqlParser.ypp`: Allow function call in `FROM` clause.
- `parser/ParseGeneratorTableReference.*`: Represent the parsed generator function as a table reference.
- `query_optimizer/resolver/Resolver.*`: Resolve the parsed generator function, link and concretize the function to its implementation. Create a `TableGenerator` logical node to represent the table that is to be populated by the generator function.
- `query_optimizer/logical/*`: Pass through the logical planner. No optimizations.
- `query_optimizer/physical/*`: Pass through the physical planner. No optimizations.
- `query_optimizer/ExecutionGenerator.*`: Generate relational operators for the TableGenerator node.
- `relational_operators/TableGeneratorOperator.*`: Create a WorkOrder to invoke the generator function to populate a relation. 
- `expressions/table_generator/GeneratorFunction*`: Abstractions for the `GeneratorFunction`. To add a new generator function, subclass both `GeneratorFunction` and `GeneratorFunctionHandle` to implement the logics, meanwhile modify `GeneratorFunctionFactory::GetByName()` to make the new function registered.
- `expressions/table_generator/GeneratorSeries*`: Implementation for the `generate_series` function.

PR #45, opened 2016-02-25
Code Style Fix for Table Gen.

zuyu:
This PR fixed code style issues merged in #39. See the wiki for the [code style guide](https://github.com/pivotalsoftware/quickstep/wiki/Quickstep-Style-Guide).

Some highlights:
 * For new files, we need to run `include-what-you-use` script.
 * For return values, the reference symbol (`&`) or the pointer symbol (`*`) should be next to the return values, instead of the function name.
 * For non-static methods, the first letter of the method name should be in the lower case.
 * Always use `std::size_t`, defined in `<cstddef>`.
 * Items in `QueryContext` is sorted by alphabet order.

### DIFF
--- a/expressions/table_generator/GenerateSeries.hpp
+++ b/expressions/table_generator/GenerateSeries.hpp
@@ -1,6 +1,7 @@
 /**
  *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
  *   University of Wisconsin—Madison.
+ *   Copyright 2016 Pivotal Software, Inc.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


expressions/table_generator/GenerateSeries.hpp
```diff
@@ -1,6 +1,7 @@
 /**
  *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
  *   University of Wisconsin—Madison.
+ *   Copyright 2016 Pivotal Software, Inc.
```
jianqiao: Do we need to add this additional notice? @pateljm

jianqiao:
Thanks for the improvement! Other things look good, except that I think we don't need the additional copyright notice, as discussed in a previous email.

expressions/table_generator/GenerateSeries.hpp
pateljm: Don't need it since you wrote the code and are not a Pivotal employee.

pateljm:
LTGM. Merging now.

expressions/table_generator/GenerateSeries.hpp
zuyu: Yes, I'm a part of `Pivotal`.

PR #46, opened 2016-02-27
Refactor passing Foreman client id to InsertDest.

zuyu:
As a follow-up to #44, this PR refactors passing the `Foreman` client id to `InsertDestination` using `QueryContext` instead of the optimizer.


pateljm:
LTGM. Merging. @hbdeshmukh Merging this, but if you see any issues with this please let us know. Would be great for you to take a quick look after your deadline.

PR #47, opened 2016-02-27
Lexer Bug

cramja:
There was a bug in the lexer which caused error conditions in escape strings to put the lexer in an invalid state. 

To recreate the bug do:
```SQL
CREATE TABLE FOO (str VARCHAR(20));
INSERT INTO foo VALUES (E'\555');
SELECT * FROM FOO;
```
The new test case in insert.test covers this case and now everything works correctly.


cramja:
Hey @jianqiao Do you mind looking this over (just a few lines, preprocessed files make it look big) and merging it if it looks good?

jianqiao:
LGTM.

PR #48, opened 2016-02-28
Delay dropping table until all blocks get deleted.

zuyu:
This PR refactors `DropTableOperator` (updated by #43) to delay the table drop until all blocks get deleted. This avoids the dangling reference to the being-dropped `CatalogRelation` in the blocks.

This PR also provides a virtual method `RelationalOperator::updateCatalogOnCompletion` for other operators such as `CreateTableOperator` and `CreateIndexOperator`.

### DIFF
--- a/catalog/CatalogDatabase.cpp
+++ b/catalog/CatalogDatabase.cpp
@@ -53,7 +53,9 @@ bool CatalogDatabase::ProtoIsValid(const serialization::CatalogDatabase &proto) 
   return true;
 }
 
-CatalogDatabase::CatalogDatabase(const serialization::CatalogDatabase &proto) : name_(proto.name()) {
+CatalogDatabase::CatalogDatabase(const serialization::CatalogDatabase &proto)
+    : name_(proto.name()),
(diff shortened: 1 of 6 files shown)


pateljm:
This is neat @zuyu. Can we also mark the catalog as being in a state "pendingBlockDeletion" before the blocks are deleted. Thus, anyone looking at the catalog when the blocks are in this state have an idea that the catalog could be inconsistent (as there is a delay between the blocks deletions being initiated and the `updateCatalogOnCompletion` being called. 

Then, we can file an issue (closed in a different PR) that requires catalog calls to check for a potential inconsistent state before returning from it.

zuyu:
@pateljm Added the suggested `Status` to indicate the inconsistent `CatalogDatabase` due to pending changes.

Please take another look! Thank you!

zuyu:
@pateljm Please see my comment above. The database is `consistent` during the construction. Thanks!

pateljm:
Thanks for the clarification @zuyu. LTGM. Merging.

PR #49, opened 2016-02-28
SQL help messages

cramja:
This change adds a new type of command syntax to the parser as well as the beginnings of a help command.

```SQL
quickstep> .help
Enter '.help {Sql Keyword} for usage information.'

quickstep> .help create
CREATE TABLE
```

In the future, we should be able to easily extend this syntax to add other commands which are non-sql commands and, when executed, perform some action on the quickstep system.

Examples of future dot commands could be:
```SQL
.quit
.version
.read
....
```
Sqlite3 uses a similar grammar as what I have implemented. A valid dot command is:
 *  A single line, starting with a dot, and ending with a newline or EOF
 * The token after the dot is the command name
 * The string following the command name is fed in as argument to the command

The `ParseHelpCommand` class still needs work. We need to find a good way of generating help messages, whether that be with a script on the YPP file or just a hand-crafted message, I don't know the best answer.

PostGres seems to embed a large string in their binary. For example, I extracted their create table help entry string entry.
``` SQL
CREATE [ [ GLOBAL | LOCAL ] { TEMPORARY | TEMP } | UNLOGGED ] TABLE [ IF NOT EXISTS ] %s
    OF %s [ (
  { %s WITH OPTIONS [ %s [ ... ] ]
    | %s }
    [, ... ]
[ WITH ( %s [= %s] [, ... ] ) | WITH OIDS | WITHOUT OIDS ]
[ ON COMMIT { PRESERVE ROWS | DELETE ROWS | DROP } ]
[ TABLESPACE %s ]
[ CONSTRAINT %s ]
{ NOT NULL |
  NULL |
  CHECK ( %s ) [ NO INHERIT ] |
  DEFAULT %s |
  UNIQUE %s |
  PRIMARY KEY %s |
  REFERENCES %s [ ( %s ) ] [ MATCH FULL | MATCH PARTIAL | MATCH SIMPLE ]
    [ ON DELETE %s ] [ ON UPDATE %s ] }
[ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
[ CONSTRAINT %s ]
{ CHECK ( %s ) [ NO INHERIT ] |
  UNIQUE ( %s [, ... ] ) %s |
  PRIMARY KEY ( %s [, ... ] ) %s |
  EXCLUDE [ USING %s ] ( %s WITH %s [, ... ] ) %s [ WHERE ( %s ) ] |
  FOREIGN KEY ( %s [, ... ] ) REFERENCES %s [ ( %s [, ... ] ) ]
    [ MATCH FULL | MATCH PARTIAL | MATCH SIMPLE ] [ ON DELETE %s ] [ ON UPDATE %s ] }
[ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
{ INCLUDING | EXCLUDING } { DEFAULTS | CONSTRAINTS | INDEXES | STORAGE | COMMENTS | ALL }
[ WITH ( %s [= %s] [, ... ] ) ]
[ USING INDEX TABLESPACE %s ]
{ %s | ( %s ) } [ %s ] [ ASC | DESC ] [ NULLS { FIRST | LAST } ]
```


cli/LineReader.cpp
```diff
@@ -77,7 +77,8 @@ std::string LineReader::getNextCommand() {
         special_char_location = multiline_buffer.find_first_of("\"\\", scan_position);
         break;
       case kComment:
-        // A newline which ends the comment and resumes normal SQL parsing.
+      case kCommand:
```
zuyu: Add `// Fall through.` after the symbol `:`.

parser/ParseCommand.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "utility/StringUtil.hpp"
+
+#include "glog/logging.h"
+
+using std::string;
+using std::map;
+
+namespace quickstep {
+
+std::map<const std::string, std::string> ParseHelpCommand::help_messages_;
```
zuyu: The `key` field in the `map` marks `const` by default, so use `std::map<std::string, std::string> ...` instead.

parser/ParseCommand.cpp
```diff
@@ -0,0 +1,87 @@
...
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "parser/ParseCommand.hpp"
+
+#include <cstdio>
+#include <map>
+#include <string>
```
zuyu: Run `iwyu` script on your new files.

For this file, you need to include `<utility>` as you use `second` in `std::pair`.

parser/ParseCommand.cpp
```diff
@@ -0,0 +1,87 @@
...
+    return;
+  }
+
+  // Look up the query in the map. If an exact match is not found, suggest
+  // the closest keyword.
+  const string lower_query = ToLower(query_->value());
+  const map<const string, string> &message_map = getHelpMessageMap();
+
+  // Try to find an exact match.
+  auto exact_itr = message_map.find(lower_query);
```
zuyu: Mark `exact_itr` as `const`.

Also, we typically use `it` to stand for an `iterator` instead of `itr`.

Finally, I'd suggest to use `matched_it` or similar to replace `exact_itr`.

parser/ParseCommand.cpp
```diff
@@ -0,0 +1,87 @@
...
+    return;
+  }
+  // If an exact match cannot be found, try to find a close match and suggest
+  // this keyword.
+  printf("Keyword %s not found\n", query_->value().c_str());
+  for (auto itr = message_map.begin();
+       itr != message_map.end();
+       ++itr) {
+    // Check if the query begins with the map key, or visa-versa.
+    string key_string = itr->first;
```
zuyu: Use `const string &key_string` to avoid a deep copy.

parser/ParseCommand.cpp
```diff
@@ -0,0 +1,87 @@
...
+  auto exact_itr = message_map.find(lower_query);
+
+  if (exact_itr != message_map.end()) {
+    printf("%s\n", exact_itr->second.c_str());
+    return;
+  }
+  // If an exact match cannot be found, try to find a close match and suggest
+  // this keyword.
+  printf("Keyword %s not found\n", query_->value().c_str());
+  for (auto itr = message_map.begin();
```
zuyu: Use `it` instead of `itr`.

parser/ParseCommand.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * @brief A help command.
+ * @details Displays help messages to the user via standard out.
+ * 
+ */
+class ParseHelpCommand : public ParseCommand {
+  // A lazily-constructed map of help messages.
+  // Maps SQL-keyword string (ex. "create"), to a description string.
+  static std::map<const std::string, std::string> help_messages_;
```
zuyu: Remove `const` in the `key` field.

parser/ParseCommand.cpp
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+  }
+}
+
+const map<const string, string>& ParseHelpCommand::getHelpMessageMap() {
+  if (help_messages_.empty()) {
+    // Construct the map.
+    help_messages_["create"] = "CREATE TABLE";
+    help_messages_["type"] = "INT | FLOAT | VARCHAR | CHAR";
```
zuyu: Why missed `LONG`, `DOUBLE`, `Datetime`, and `interval` types?

cramja:
@zuyu Comments noted and corrected. Ran iwyu. Do you generally follow all of its suggestions?

Also, I added a TODO to the method which generates the helpMessageMap. Basically, it's unfinished and I haven't yet thought up a systematic way of coming up with intuitive help messages.

jianqiao:
The overall framework looks good to me. I would suggest two refactorings be done later when we have added a set of commands:
(1) Parse commands as `ParseString`'s instead of tokens, and resolve command semantics after the parsing phase. In this way we don't need to modify lexer and parser each time when adding new commands.
(2) Try to separate the command execution logic from the `ParseCommand` class. Because `/parser/ParseCommand`'s original role is to represent an AST node. Since the execution logic can be arbitrary and will depend on arbitrary modules, there should be a standalone place to put the command execution code.

zuyu:
@cramja No, `IWYU` just provides some suggestions, but also introduces false positive. Use your own judgements to add the headers for new changes. There are many legacy `IWYU` issues to fix in a separate PR.

Adding `TODO` is a good practice.

jianqiao:
@cramja Hi Marc, let's nail it down in #52 whether to use `.command` or `\command` syntax. Then I can merge this PR.

cramja:
Personally, I prefer the '.command'. Do you have a preference?

Also, I will finish this feature by addressing @zuyu s comments and then changing the map creation from being read from a file to being created programatically.

jianqiao:
I see. How about enabling both `.` and `\`? The change won't be much.

pateljm:
Go for it @jianqiao and put in both `.` and `\` !

cramja:
That shouldn't be a problem to implement and I will do it after this, then the PR will be complete.

@jianqiao I have thought about what you have said about abstracting Commands into a separate type of class which is separate of any parser parent class. I polled @hakanmemisoglu and @adalbertgerald  for advice on where these classes should go and we decided that a folder `quickstep/commands` would be the most fitting place for a top level class which is essentially:
```c++
class Command {
  abstract void execute() = 0;
};
```
and then all of the child classes, including `HelpCommand`

What I will do is change the grammar to allow `\cmd` and `.cmd` and then this PR will be finished, however, with the idea that the next PR will be a refactor:
* Change grammar from `\|.TOKEN -> \|.ParseString` as @jianqiao suggested as to not pollute the token namespace with non-sql tokens.
* Make a separate folder for commands.
* Refactor the logic to instantiate commands.

Another quick refactor would be to change the QuitStatement from an SQL statement to a Command.

cramja:
Also a separate issue from the location and design of commands was the issue of how to implement help text. 

We agreed in the meeting that creating a separate file, "Help.txt" which would contain all of the help text and which would be read in at runtime would be the way to go. However, this runs into the problem of if the user moves the binary, the path to that file changes, and you have a non-working help system. 

It might just be best to include all help text in the binary as a c++11 literal string and tell LINT to ignore that file (which can be done with a `CPPLINT.cfg` file):
```
exclude_files=ParseCommand.cpp
```

jianqiao:
Sounds great!

cramja:
@pateljm @jianqiao If there are no other issues, this PR is finished.

pateljm:
Agreed @cramja. LTGM. Merging.

PR #55, opened 2016-03-02
Refactor commands

cramja:
@jianqiao I made the first part of the changes for the ParseCommand refactor.
 * class ParseCommand is now just a simple holder class for the 2 strings which comprises a command: the name and an argument string.

*Note* This also fixes a bug in the parser. Currently, anytime quickstep_yyerror is called in the parser or lexer, a `throw` is executed, leaving the parser in a non-initial state. This didn't cause problems previously because the lexer only had one 'initial' state. 

To reset the state of the parser, in quickstepCLI.cpp I have wrapped the ParserWrapper in a unique_ptr, and I reset it on an error condition.


cramja:
@jianqiao Can we merge this if it looks good?

parser/ParseStatement.hpp
```diff
@@ -885,6 +885,73 @@ class ParseStatementQuit : public ParseStatement {
...
+ * @brief Class to hold the parsed command name and an optional argument string.
+ * @details Commands are non-sql statements which can be issued to quickstep.
+ *          They are entered into the CLI as '.command-name command-string\n'.
+ */
+class ParseCommand : public ParseStatement {
+ public:
+  ParseCommand(const int line_number,
+               const int column_number,
+               ParseString *command,
+               ParseString *arguments)
```
jianqiao: Perhaps we should allow multiple arguments.

jianqiao:
@cramja Hi Marc, do you think we should allow multiple arguments to the command or just one argument?
Other places look good.

parser/ParseStatement.hpp
jianqiao: I note that the rest of the line after the command is passed in as a single string. In this way the splitting of arguments is left to the execution logic. I think it might be better to let lexer do the splitting, the same as command-line arguments which are separated by whitespaces and passed in as a vector of strings.

cramja:
@jianqiao I think that is a good idea, will add.

cramja:
@jianqiao should be good after Travis.

jianqiao:
LGTM. Merging.

PR #56, opened 2016-03-02
Transaction Integration Part 1: DirectedGraph, SCC, Transaction

hakanmemisoglu:
### Classes
This PR introduces 3 classes necessary for locking support:
#### DirectedGraph:
    A class for representing wait for graphs between transactions to 
    detect deadlocks.

#### StronglyConnectedComponents:
    A class for Tarjan's Strongly Connected Coponents algorithm. SCC
    algorithm will give us components in which the cycles are searched.

#### Transaction:
    Included here, because SCC's test includes it but does not use any 
    of its methods. Therefore test for this class is not included.

### Minor changes:
 -  "transaction" folder is added to root folder.
 -  Root CMakeLists.txt is changed, now it contains a command to compile "transaction" folder.
 -  Module support for "transaction" folder: TransactionModule.hpp


transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,113 @@
...
+ **/
+class StronglyConnectedComponents {
+ public:
+  /**
+   * @brief Constructor for StronglyConnectedComponents.
+   *
+   * @param directed_graph Pointer to the directed graph which the 
+   *        algortihm runs.
+   **/
+  explicit StronglyConnectedComponents(DirectedGraph *directed_graph);
```
pateljm: Is there a reason to store the pointer instead of `const DirectedGraph &`

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,113 @@
...
+   * @return Mapping of component ids to list of nodes.
+   **/
+  std::unordered_map<std::uint64_t, std::vector<DirectedGraph::NodeId>> getComponentMapping() const;
+
+ private:
+  // Applies DFS to the node whic has id v.
+  void depthFirstSearch(DirectedGraph::NodeId v);
+
+  // Pointer to the graph.
+  DirectedGraph *directed_graph_;
```
pateljm: Who "owns" the graph that is pointed to. Should make the contract clear, especially if that graph could change between calls to methods here.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: It is not an owner pointer. I used a pointer to emphasize it. Const reference also makes sense since it never acquires a nullptr. I will change it. Thanks Jignesh.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: I used the term "pointer" as non-owner pointer, and "owner-pointer" to denote the pointer that owns the resource. I will change it to a const reference. Thanks for the comments.

transaction/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+add_library(quickstep_transaction_DirectedGraph
```
hbdeshmukh: Need the copyright declaration.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+#ifndef QUICKSTEP_TRANSACTION_DIRECTED_GRAPH_HPP_
+#define QUICKSTEP_TRANSACTION_DIRECTED_GRAPH_HPP_
+
+#include <algorithm>
+#include <memory>
+#include <unordered_map>
+#include <unordered_set>
+#include <stack>
+#include <vector>
```
hbdeshmukh: The includes should be alphabetically sorted. I would highly recommend doing it through an IDE instead of sorting it manually. e.g. in vim, you can select a visual block and issue a sort command. You may want to find something similar for emacs.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+ */
+
+/**
+ * @brief Class for representing a directed graph.
+ *        Vertices are transaction ids, edges are
+ *        wait-for relations.
+ */
+class DirectedGraph {
+ public:
+  using NodeId = std::uint64_t;
```
hbdeshmukh: Our data types are usually written as ``node_id`` instead of ``NodeId``

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * @brief Default constructor
+   */
+  DirectedGraph() {}
+
+  /**
+   * @brief Adds node with given transaction id.
+   * @warning It does not check whether transaction id
```
hbdeshmukh: To get around this warning, you can implement two methods -- ``addNodeCheckExists`` and ``addNodeUnchecked``. In the first method, you can perform a check before addition of the node. If the check is successful, only then the node will be added. The second method is essentially this method.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+   *          is already in the graph.
+   * @warning Pointer ownership will pass to the graph,
```
hbdeshmukh: I don't see where the ownership is getting transferred. Can you explain how it is done?

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+   * @return Id of the newly created node.
+   */
+  inline NodeId addNode(TransactionId *data) {
+    nodes_.emplace_back(data);
+    return nodes_.size() - 1;
+  }
+
+  /**
+   * @brief Adds an edge between nodes.
+   * @warning Does not check arguments are legit.
```
hbdeshmukh: Similar to the comments in ``addNode`` method, you can implement two methods -- an unchecked version and a checked version.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+   *
+   * @return True if there is an edge, false otherwise.
+   */
+  inline bool hasEdge(NodeId from_node, NodeId to_node) const {
+    return nodes_[from_node].hasOutgoingEdge(to_node);
+  }
+
+  /**
+   * @brief Get data (transaction id) contained in the node.
+   * @warning Does not check index validity.
```
hbdeshmukh: You can add a DCHECK to ensure that the node ID belongs to the graph. DCHECK only runs in debug builds and doesn't cause an overhead in the release build. It is immensely helpful in spotting errors.

transaction/DirectedGraph.hpp
hbdeshmukh: Found it, nevermind.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    inline bool hasOutgoingEdge(NodeId to_node) const {
+      return outgoing_edges_.count(to_node) == 1;
+    }
+
+    inline std::vector<NodeId> getOutgoingEdges() const {
+      std::vector<NodeId> result;
+      std::copy(outgoing_edges_.begin(), outgoing_edges_.end(),
+                std::back_inserter(result));
+      return result;
```
hbdeshmukh: You are returning a vector from stack, which can cause unexpected behavior at the caller site. You should change the return type to ``std::vector<NodeId>*``, change the type of result to ``std::unique_ptr<std::vector<NodeId>`` and while returning, do ``result.release()``.

transaction/StronglyConnectedComponents.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include <vector>
+
+namespace quickstep {
+
+namespace transaction {
+
+StronglyConnectedComponents::StronglyConnectedComponents(const DirectedGraph &directed_graph)
+  : directed_graph_(directed_graph)
+  , is_marked_(directed_graph_.count())
```
hbdeshmukh: The commas in the beginning should be moved to the end of each line in constructor initializer list.

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+/**
+ * @brief Class for applying Tarjan's strongly connected components
+ *        algorithm to the 
+ **/
+class StronglyConnectedComponents {
+ public:
+  /**
+   * @brief Constructor for StronglyConnectedComponents.
```
hbdeshmukh: Just saying ``@brief Constructor.`` is enough.

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @param directed_graph Const reference to the directed graph which the
```
hbdeshmukh: The ``const reference`` part can be omitted for simplicity.

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+   **/
+  std::uint64_t getComponentId(DirectedGraph::NodeId node_id) const;
+
+  /**
+   * @brief Getter for total number of strongly connected components.
+   * @warning This method should be called after findStronglyConnectedComponents().
+   *
+   * @return Total number of strongly connected components.
+   **/
+  std::uint64_t getTotalComponents() const;
```
hbdeshmukh: Wouldn't a ``size_t`` be enough to represent the number of components?

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<std::uint64_t> low_ids_;
+
+  // Keeps track of the nodes in the current traversal of the graph.
+  std::stack<DirectedGraph::NodeId> stack_;
+
+  // Keeps track of the current index number of depth first search traversal.
+  std::uint64_t preorder_counter_;
+
+  // Keeps track of the current number of the strongly connected component.
+  std::uint64_t no_of_components_;
```
hbdeshmukh: The comments for each class member are quite helpful!

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+  DISALLOW_COPY_AND_ASSIGN(StronglyConnectedComponents);
+};
+
+/** @} */
+
+}  // namespace transaction
+
+}  // namespace quickstep
+
+#endif
```
hbdeshmukh: Usually the format of the last line is ``#endif  // QUICKSTEP_TRANSACTION_STRONGLY_CONNECTED_COMPONENTS_HPP_``

transaction/Transaction.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+TransactionStatus Transaction::getStatus() const {
+  return status_;
+}
+
+bool Transaction::operator==(const Transaction &other) const {
+  return tid_ == other.tid_;
+}
+
+std::size_t Transaction::TransactionHasher::operator()(const Transaction &transaction) const {
```
hbdeshmukh: All the functions in this cpp file are so small that you can move all of them to the hpp file and mark them inline. No need for a cpp file for short functions.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+namespace transaction {
+
+/** \addtogroup Transaction
+ *  @{
+ */
+
+/**
+ * @brief Class for representing a directed graph.
+ *        Vertices are transaction ids, edges are
```
hbdeshmukh: We allow comments (and code) to be 80 chars wide. So feel free to use that space.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+    inline TransactionId getData() const {
+      return *(data_.get());
+    }
+
+   private:
+    // Owner pointer to transaction id.
+    std::unique_ptr<TransactionId> data_;
+
+    // Endpoint nodes of outgoing edges originated from this node.
+    std::unordered_set<NodeId> outgoing_edges_;
```
hbdeshmukh: Missing ``DISALLOW_COPY_AND_ASSIGN(DirectedGraphsNode)``. Do you intend this class to be copyable (and with an assignment operator)?

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace transaction {
+
+/** \addtogroup Transaction
+ *  @{
+ */
+
+/**
+ * @brief Class for applying Tarjan's strongly connected components
+ *        algorithm to the
```
hbdeshmukh: Incomplete comment. Also, the comment should explain the class as a whole instead of describing one of its function. You can say ``A class that finds and stores strongly connected components of a given directed graph``.

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+   * @brief Constructor for StronglyConnectedComponents.
+   *
+   * @param directed_graph Const reference to the directed graph which the 
+   *        algorithm runs.
+   **/
+  explicit StronglyConnectedComponents(const DirectedGraph &directed_graph);
+
+  /**
+   * @brief Enumerates strongly connected components.
+   * @info This method should be called before calling any other method.
```
hbdeshmukh: Instead of saying any other method you should mention which methods in particular.

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+   **/
+  void findStronglyConnectedComponents();
+
+  /**
+   * @brief Getter for the component id of the node.
```
hbdeshmukh: ``Get the ID of the component to which the given node belongs``

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @param node_id Id of the node in the graph.
+   * @return Component id of the node.
```
hbdeshmukh: ``ID of the component to which the given node belongs.``

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+  // Applies DFS to the node whic has id v.
+  void depthFirstSearch(DirectedGraph::NodeId v);
+
+  // Const reference to the graph.
+  const DirectedGraph &directed_graph_;
+
+  // is_marked[v] == true if depth first search traversed the node with id v.
+  std::vector<bool> is_marked_;
+
+  // component_ids_[v] == i if strongly connected component
```
hbdeshmukh: Do you mean ``component_ids_[id] == i``?

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+  // which the node with id is in, has id i.
```
hbdeshmukh: Or you mean ``which the node with id v``?

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<std::uint64_t> component_ids_;
+
+  // low_ids_[v] == i if lowest traversal id seen by the node v equals i.
+  std::vector<std::uint64_t> low_ids_;
+
+  // Keeps track of the nodes in the current traversal of the graph.
+  std::stack<DirectedGraph::NodeId> stack_;
```
hbdeshmukh: Can we have a more informational name for this variable?

transaction/DirectedGraph.hpp
hakanmemisoglu: Ok, I will change it. Thanks for the warning.

transaction/DirectedGraph.hpp
hakanmemisoglu: It makes sense. I will introduce the  `addNodeCheckExists` method.

transaction/StronglyConnectedComponents.cpp
hakanmemisoglu: Noted. Thanks for the warning.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: That's why programming at night is dangerous. Thanks.

transaction/DirectedGraph.hpp
hakanmemisoglu: Thanks, it is good to know. I will use DCHECK at appropriate locations.

transaction/DirectedGraph.hpp
hakanmemisoglu: I wanted to be copyable. However, this is a private class only exposed to `DirectedGraph` and it does not copy nodes. Therefore disabling copy constructor makes sense. The only problem is that the nodes are stored in the vector and  if I delete the copy constructor, I should introduce move assignment and move constructor, otherwise it will not compile.

transaction/DirectedGraph.hpp
hakanmemisoglu: Thanks for the tip. I will fix it.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,173 @@
...
+  inline TransactionId getDataFromNode(NodeId node) const {
+    return nodes_[node].getData();
+  }
+
+  /**
+   * @brief Calculate how many nodes the graph has.
+   *
+   * @return The number of nodes the graph has.
+   */
+  inline std::size_t count() const {
```
hbdeshmukh: We need a more informative name like - ``getNumNodes()`` or something similar.

transaction/StronglyConnectedComponents.cpp
hbdeshmukh: Even though nothing is incorrect with the initialization of ``vector<bool>`` as you've done, an explicit initialization like ``is_marked_(directed_graph_.count(), false)`` is unambiguous.

transaction/StronglyConnectedComponents.cpp
```diff
@@ -0,0 +1,125 @@
...
+  , component_ids_(directed_graph_.count())
+  , low_ids_(directed_graph_.count())
```
hbdeshmukh: Same comment as for the initialization of ``is_marked_``

transaction/StronglyConnectedComponents.cpp
```diff
@@ -0,0 +1,125 @@
...
+  , preorder_counter_(0)
+  , no_of_components_(0) {
+}
+
+void StronglyConnectedComponents::findStronglyConnectedComponents() {
```
hbdeshmukh: As the sole purpose of this class is to find strongly connected components and store them, why not call this function in the constructor? This will ensure that as soon as we create an instance of this class, we will always have all the strongly connected components present.

transaction/StronglyConnectedComponents.cpp
```diff
@@ -0,0 +1,125 @@
...
+    // which is the (no_nodes + 1).
+    low_ids_[w] = directed_graph_.count();
+    stack_.pop();
+  } while (w != v);
+
+  // Increment current no of component counter.
+  no_of_components_++;
+}
+
+std::uint64_t StronglyConnectedComponents::getComponentId(DirectedGraph::NodeId node_id) const {
```
hbdeshmukh: This function can be marked inline (and will need to be moved to header file).

transaction/StronglyConnectedComponents.cpp
```diff
@@ -0,0 +1,125 @@
...
+  return component_ids_[node_id];
+}
+
+std::uint64_t StronglyConnectedComponents::getTotalComponents() const {
```
hbdeshmukh: This function can be marked inline (and will need to be moved to header file).

transaction/StronglyConnectedComponents.cpp
```diff
@@ -0,0 +1,125 @@
...
+  return no_of_components_;
+}
+
+std::unordered_map<std::uint64_t, std::vector<DirectedGraph::NodeId>>
+    StronglyConnectedComponents::getComponentMapping() const {
+  std::unordered_map<std::uint64_t, std::vector<DirectedGraph::NodeId>> component_mapping;
+  for (std::uint64_t i = 0; i < component_ids_.size(); ++i) {
+    component_mapping[component_ids_[i]].push_back(i);
+  }
+  return component_mapping;
```
hbdeshmukh: Returning a stack variable, I mentioned about it somewhere else. This needs to be fixed.

transaction/StronglyConnectedComponents.hpp
hbdeshmukh: Haha, it sure is!

transaction/Transaction.hpp
```diff
@@ -0,0 +1,104 @@
...
+namespace transaction {
+
+/** \addtogroup Transaction
+ *  @{
+ */
+
+/**
+ * @brief Type that represents id of a transaction.
+ **/
+using TransactionId = std::uint64_t;
```
hbdeshmukh: Let's use typedef instead of ``TransactionId``. Also, the general format of data types is ``transaction_id`` and not ``TransactionId``

transaction/Transaction.hpp
```diff
@@ -0,0 +1,104 @@
...
+  TransactionId getTransactionId() const;
+
+  /**
+   * @brief Setter for transaction status.
+   *
+   * @param status New status of the transaction.
+   **/
+  void setStatus(TransactionStatus status);
+
+  TransactionStatus getStatus() const;
```
hbdeshmukh: Need doxygen. Just a ``@brief`` is enough.

transaction/Transaction.hpp
```diff
@@ -0,0 +1,104 @@
...
+   * @brief Equality operator for Transaction class.
+   *
+   * @param other Other transaction to be compared with.
+   * @return True if other tid and this tid are equal, false otherwise.
+   **/
+  bool operator==(const Transaction &other) const;
+
+  /**
+   * @brief Helper hasher class for transaction.
+   *
```
hbdeshmukh: Extra line.

transaction/Transaction.hpp
```diff
@@ -0,0 +1,104 @@
...
+   **/
+  struct TransactionHasher {
+    std::size_t operator()(const Transaction &transaction) const;
+  };
+
+ private:
+  TransactionId tid_;
+  TransactionStatus status_;
+  // std::unique_ptr<QueryPlan> query_plan_;
+  // std::unique_ptr<TransactionThread> thread_;
```
hbdeshmukh: Are these variables needed? If not, you can get rid of them.

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,114 @@
...
+ **/
+class StronglyConnectedComponents {
+ public:
+  /**
+   * @brief Constructor for StronglyConnectedComponents.
+   *
+   * @param directed_graph Const reference to the directed graph which the 
+   *        algorithm runs.
+   **/
+  explicit StronglyConnectedComponents(const DirectedGraph &directed_graph);
```
hbdeshmukh: Do you assume that there won't be any changes (e.g. addition or removal of a node) in ``directed_graph``? If not, then we need to provide methods which update the strongly connected components based on the changes in the original graph. In that case we will have to change the type of ``directed_graph_`` from const reference to a pointer.

hbdeshmukh:
Hi @hakanmemisoglu 

Just finished a full round on the hpp and cpp files (not the tests). Let me know when you address the comments.

transaction/DirectedGraph.hpp
hbdeshmukh: There's a correction in my comment. Your code will not be incorrect, but it will be inefficient. e.g. when someone calls like this - 
``std::vector<NodeId> outgoing_edges = getOutgoingEdges()``, a copy will be made from ``result`` to ``outgoing_edges``. Depending on the size of the vector, this copy could be quite expensive. My suggestion will help you avoid that copy.

transaction/DirectedGraph.hpp
hakanmemisoglu: Will fix it.

transaction/DirectedGraph.hpp
hakanmemisoglu: Harshad, as you said in your last post, it is correct, caller will not get any stack variable in that sense. For efficiency, it is actually the most efficient method for returning a container with a C++11 compiler. Compiler will apply RVO to eliminate unnecessary copy, if it cannot it will call move assignment for the constructor. If you introduce any indirection to return this container, the compiler cannot introduce any optimization. Unless it is the project convention I want to keep it because it is the 'most' appropriate way to return a container.

transaction/StronglyConnectedComponents.cpp
hakanmemisoglu: Yes, I will fix it. Thanks.

transaction/StronglyConnectedComponents.cpp
hakanmemisoglu: It is true, I will change it so that it will be called in the constructor.

transaction/Transaction.hpp
hakanmemisoglu: I will change all non-template type aliasing to `typedef` and will rename the type names.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: Yes, that is the case. I will fix it.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: I will change its name to the traversal_stack.

transaction/DirectedGraph.hpp
hakanmemisoglu: Yes, `getNumNodes()` is more explanatory.

transaction/StronglyConnectedComponents.cpp
hakanmemisoglu: I will fix it.

transaction/StronglyConnectedComponents.cpp
hakanmemisoglu: Ok, I will move it.

transaction/StronglyConnectedComponents.cpp
hakanmemisoglu: I will make it inline in the header file.

transaction/StronglyConnectedComponents.cpp
hakanmemisoglu: We talked about it, it seems alright at this point.

transaction/Transaction.hpp
hakanmemisoglu: Thanks for the warning. I will add comment.

transaction/Transaction.hpp
hakanmemisoglu: Thanks for the warning.

transaction/Transaction.hpp
hakanmemisoglu: Right now, I don't. I will delete them.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: The directed graph will not change during the exposition to the SCC. Therefore, SCC has const access to the directed graph.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: Noted, thanks.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: Yes, I will change it.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: I responded at below.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: As you said in the other comment, I moved it to the constructor. The user should not bother with it.

transaction/Transaction.cpp
hakanmemisoglu: I will move them.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: I forget to include this last line comment in every header files. I will add it to every header file.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: Yes, `size_t` makes more sense.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: I will remove it.

transaction/StronglyConnectedComponents.hpp
hakanmemisoglu: I will change it.

hakanmemisoglu:
Hi @hbdeshmukh, thanks for all your feedback.
I addressed your comments. Travis also seems happy.

transaction/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+            DirectedGraph.hpp)
+add_library(quickstep_transaction_StronglyConnectedComponents
+            StronglyConnectedComponents.cpp
+            StronglyConnectedComponents.hpp)
+add_library(quickstep_transaction_Transaction
+            ../empty_src.cpp
+            Transaction.hpp)
+
+target_link_libraries(quickstep_transaction_DirectedGraph
+	              glog
```
hbdeshmukh: Need to fix the indentation.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * @warning Pointer ownership will pass to the graph, therefore it
+   *          should not be deleted.
+   *
+   * @param data Pointer to the transaction id that will be contained
+   *        in the node.
+   * @return Id of the newly created node.
+   **/
+  inline node_id addNodeCheckExists(transaction_id *data) {
+    for (std::vector<DirectedGraphNode>::const_iterator
+           it = nodes_.cbegin(), end = nodes_.cend();
```
hbdeshmukh: No need to declare ``end``. The for loop can simply be
```
for (std::vector<DirectedGraphNode>::const_iterator it = node_.cbegin(); it != nodes_.cend(); ++it) {
...
}
```

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  /**
+   * @brief Get data (transaction id) contained in the node.
+   * @warning Does not check index validity.
+   *
+   * @param node Id of the node that the data is got from.
+   * @return Id of the transaction that this node contains.
+   **/
+  inline transaction_id getDataFromNode(node_id node) const {
```
hbdeshmukh: You can use a similar DCHECK here as in  ``hasEdge``.

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,206 @@
...
+    return nodes_.size();
+  }
+
+  /**
+   * @brief Gives the node ids that this node has edges to.
+   *
+   * @param id Id of the corresponding node.
+   * @return Vector of node ids that id has edges to.
+   **/
+  inline std::vector<node_id> getAdjacentNodes(node_id id) const {
```
hbdeshmukh: We had discussed about returning a const reference to the vector. Do you have any concerns with that idea?

transaction/DirectedGraph.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+   private:
+    // Owner pointer to transaction id.
+    std::unique_ptr<transaction_id> data_;
+
+    // Endpoint nodes of outgoing edges originated from this node.
+    std::unordered_set<node_id> outgoing_edges_;
+  };
+
+  // Buffer of nodes that are created. NodeId is the index of that
```
hbdeshmukh: The term buffer may be misleading. You can just say a set or a list.

transaction/StronglyConnectedComponents.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * @brief Getter for the component id of the node.
+   *
+   * @param node_id Id of the node in the graph.
+   * @return Id of the component to which the given node belongs.
+   **/
+  inline std::uint64_t getComponentId(DirectedGraph::node_id node_id) const {
```
hbdeshmukh: You are using uint64_t for component IDs and size_t for total components. You should stick to the same data type.

transaction/Transaction.hpp
```diff
@@ -0,0 +1,120 @@
...
+  explicit Transaction(transaction_id tid)
+    : tid_(tid) {
+  }
+
+  /**
+   * @brief Getter for transaction id.
+   *
+   * @return Transaction id of this.
+   **/
+  inline transaction_id gettransaction_id() const {
```
hbdeshmukh: Rename the function to ``getTransactionId()``

transaction/tests/DirectedGraph_unittest.cpp
```diff
@@ -0,0 +1,110 @@
...
+namespace transaction {
+
+TEST(DirectedGraphTest, AddNode) {
+  DirectedGraph wait_for_graph;
+  transaction_id *tid3 = new transaction_id(3);
+  transaction_id *tid4 = new transaction_id(4);
+  transaction_id *tid5 = new transaction_id(5);
+  transaction_id *tid6 = new transaction_id(6);
+
+  EXPECT_EQ(0u, wait_for_graph.getNumNodes());
```
hbdeshmukh: It's helpful to explain what you are going to test and why it should succeed. e.g. before this line, you could say that - ``As all the transactions are independent of each other, wait for graph is empty``. Trust me, as your tests grow larger such comments will be a life save when you refactor your code.

transaction/tests/DirectedGraph_unittest.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+TEST(DirectedGraphTest, AddEdge) {
+  DirectedGraph wait_for_graph;
+  transaction_id *tid3 = new transaction_id(3);
+  transaction_id *tid4 = new transaction_id(4);
+  transaction_id *tid5 = new transaction_id(5);
+  transaction_id *tid6 = new transaction_id(6);
+
+  using NID = DirectedGraph::node_id;
```
hbdeshmukh: I would prefer using typedef. Also data types are usually lower case.

transaction/tests/DirectedGraph_unittest.cpp
```diff
@@ -0,0 +1,110 @@
...
+  EXPECT_FALSE(wait_for_graph.hasEdge(nid6, nid3));
+  EXPECT_FALSE(wait_for_graph.hasEdge(nid4, nid5));
+}
+
+TEST(DirectedGraphTest, GetAdjacentNodes) {
+  DirectedGraph wait_for_graph;
+  transaction_id *tid3 = new transaction_id(3);
+  transaction_id *tid4 = new transaction_id(4);
+  transaction_id *tid5 = new transaction_id(5);
+  transaction_id *tid6 = new transaction_id(6);
```
hbdeshmukh: The above five lines are getting reused in all the test fixtures. You can create a ``SetUp()`` method and move these lines from all the fixtures to the ``SetUp()`` method and then switch from ``TEST`` to ``TEST_F``

transaction/tests/DirectedGraph_unittest.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  std::vector<NID> result2 = wait_for_graph.getAdjacentNodes(nid3);
+  EXPECT_EQ(2u, result2.size());
+
+  wait_for_graph.addEdgeUnchecked(nid3, nid6);
+  std::vector<NID> result3 = wait_for_graph.getAdjacentNodes(nid3);
+  EXPECT_EQ(3u, result3.size());
+}
+
+}  // namespace transaction
```
hbdeshmukh: This is a nicely done test!

transaction/tests/StronglyConnectedComponents_unittest.cpp
```diff
@@ -0,0 +1,286 @@
...
+  std::uint64_t nid3_component;
+  std::uint64_t nid4_component;
+  std::uint64_t nid5_component;
+  std::uint64_t nid6_component;
+  std::uint64_t nid7_component;
+  std::uint64_t nid8_component;
+  std::uint64_t nid9_component;
+  std::uint64_t nid10_component;
+  std::uint64_t nid11_component;
+  std::uint64_t nid12_component;
```
hbdeshmukh: Please don't use several different variables like above. Switch to a vector.

transaction/tests/StronglyConnectedComponents_unittest.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+  std::uint64_t total_components;
+};
```
hbdeshmukh: I have some ideas about simplifying this test. It will be better if we can talk about it in person sometime today.

hbdeshmukh:
Hey @hakanmemisoglu 

I did another round of review. Things are looking good in the hpp/cpp files, we need to discuss about simplifying the tests. Once that is done, the PR will be good to merge.

transaction/tests/DirectedGraph_unittest.cpp
cramja: Echoing Harshad, I have not seen this style of `using` (inside a block versus directly below the includes) in the code base.

transaction/tests/StronglyConnectedComponents_unittest.cpp
cramja: P.S. Vectors have a nice constructor feature: `vector<int> v({1,2,3,4,5,6})` which will make the code prettier.

hbdeshmukh:
I forgot to mention this before. Have you considered the thread-safety aspect of these methods? Is there an assumption that there will be only one caller thread for these methods?

hakanmemisoglu:
These are the part of the Deadlock Detector. The algorithm will be run by only one thread.

transaction/tests/StronglyConnectedComponents_unittest.cpp
hakanmemisoglu: I switched over the vector. Thanks for the tip.

transaction/tests/StronglyConnectedComponents_unittest.cpp
hakanmemisoglu: I applied the structure we talked.

transaction/tests/DirectedGraph_unittest.cpp
hakanmemisoglu: I changed the structure.

transaction/tests/DirectedGraph_unittest.cpp
hakanmemisoglu: `using` and `typedef` are removed.

transaction/tests/DirectedGraph_unittest.cpp
hakanmemisoglu: Ok I will do that.

transaction/Transaction.hpp
hakanmemisoglu: Noted.

transaction/DirectedGraph.hpp
hakanmemisoglu: Right, I will change it.

transaction/DirectedGraph.hpp
hakanmemisoglu: Harshad, I thought the alternatives. Storing the data in vector, and checking the uniqueness in each `addEdge` operation, is more expensive since the number of `addEdge` calls is more than the number of `getAdjacentNodes` calls. I think, we should keep it this way, for now.

hakanmemisoglu:
Hi @hbdeshmukh, I completed the improvements. The only thing that I did not touch is the converting set to vector. I gave my reasoning in the comments section.

transaction/DirectedGraph.hpp
hbdeshmukh: Sounds good. Can you add a TODO saying something like this -
```
// TODO(hakanmemisoglu) - getOutgoingEdges() performs a copy. In the future, if we detect that there is a performance issue because of this copy, we may need to review the function implementation.
```

transaction/DirectedGraph.hpp
hbdeshmukh: This comments needs to be addressed.

transaction/tests/StronglyConnectedComponents_unittest.cpp
```diff
@@ -0,0 +1,234 @@
...
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "transaction/StronglyConnectedComponents.hpp"
+
+#include <cstdint>
```
hbdeshmukh: Need to include memory header for ``unique_ptr``.

transaction/tests/StronglyConnectedComponents_unittest.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+  std::unique_ptr<GraphConfiguration> graph_configuration_;
+  std::unique_ptr<DirectedGraph> wait_for_graph_;
+  std::unique_ptr<StronglyConnectedComponents> scc_;
+};
+
+class StronglyConnectedComponentsTestWithTwoNodesCycle : public testing::Test {
+ protected:
+  StronglyConnectedComponentsTestWithTwoNodesCycle() {
+    wait_for_graph_ = std::make_unique<DirectedGraph>();
```
hbdeshmukh: Alternatively, you can use ``wait_for_graph_.reset(new DirectedGraph())``

transaction/tests/StronglyConnectedComponents_unittest.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+TEST_F(StronglyConnectedComponentsTest, GetComponentId) {
+  std::vector<DirectedGraph::node_id> nid_list;
+  for (DirectedGraph::node_id nid = 0;
+       nid < wait_for_graph_->getNumNodes();
+       ++nid) {
+    nid_list.push_back(scc_->getComponentId(nid));
+  }
+
+  // Check the result are as expected.
```
hbdeshmukh: Before this check, you can put an assert for the size of ``nid_list``.

transaction/tests/StronglyConnectedComponents_unittest.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+  EXPECT_EQ(1u, nid_list[2]);
+  EXPECT_EQ(1u, nid_list[5]);
+
+  EXPECT_EQ(0u, nid_list[6]);
+  EXPECT_EQ(0u, nid_list[7]);
+  EXPECT_EQ(0u, nid_list[8]);
+  EXPECT_EQ(0u, nid_list[9]);
+  EXPECT_EQ(0u, nid_list[10]);
+  EXPECT_EQ(0u, nid_list[11]);
```
hbdeshmukh: You need not do it, but the above set of ``EXPECT_EQ`` can be made even more compact. You can create a vector called expected_component_count and initialize it with ``{3, 2, 2, ...0}``. Then in a loop, you can check the equality of ``expected_component_count[i]`` and ``nid_list[i]``

hbdeshmukh:
Hey @hakanmemisoglu 

Good job on refactoring the tests! I just have couple more comments, once you address them we can get this PR merged.

hakanmemisoglu:
Hi @hbdeshmukh 
I fixed the parts that you mentioned. I had problems with running iwyu on cs machine, therefore I included and checked the necessary header manually.

hakanmemisoglu:
Hey, @hbdeshmukh 
I fixed the all points that you mentioned and Travis seems to be happy.

transaction/CMakeLists.txt
hbdeshmukh: Not an issue here. Perhaps Google Chrome acting weirdly.

transaction/StronglyConnectedComponents.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <cstdint>
+#include <stack>
+#include <unordered_map>
+#include <vector>
+
+namespace quickstep {
+
+namespace transaction {
+
+StronglyConnectedComponents::StronglyConnectedComponents(
```
hbdeshmukh: This constructor needs to be marked as explicit.

transaction/StronglyConnectedComponents.hpp
hbdeshmukh: Me and @hakanmemisoglu discussed and concluded that size_t is fine for returning a count/size.

hbdeshmukh:
Looks good to me, merging. Great job @hakanmemisoglu !

PR #54, opened 2016-03-02
Fixed the block counting bug.

zuyu:
This PR should fix the block counting bug encountered in `quickstep_queryoptimizer_tests_executiongenerator`. See https://travis-ci.org/pivotalsoftware/quickstep/jobs/112693625 for more info.

```
F0229 21:18:05.125586 17407 StorageBlockBase.hpp:42] Check failed: 0 == ref_count_ (0 vs. 1) Nonzero ref_count_ when deleting block (1, 2)
```

The main issue would be that we access the atomic value in an incorrect way. This should be fixed by accessing the value using the given API `load`, instead of the value name directly.

PR #58, opened 2016-03-03
Revert "Fixed the block counting bug."

zuyu:
Reverts pivotalsoftware/quickstep#54

The bug still exists, as shown in https://travis-ci.org/pivotalsoftware/quickstep/jobs/113138151

### DIFF
--- a/storage/StorageBlockBase.hpp
+++ b/storage/StorageBlockBase.hpp
@@ -1,6 +1,6 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
- *   Copyright 2015-2016 Pivotal Software, Inc.
+ *   Copyright 2015 Pivotal Software, Inc.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pateljm:
@zuyu I think accessing the `ref_count_` via load is the right approach. But we have another bug somewhere. So, I don't think we should revert.

jianqiao:
Since blocks are all owned by `StorageManager`, I suspect the scenario of the bug is that some `CountedReference`'s are not released when `StorageManager` evicts a block.

I suggest one fix, but I'm not sure if it solves the problem because I cannot repeat the bug.

It is in `ExecutionGeneratorTestRunner.cpp`. Add one line of
`query_context_.reset(nullptr);` 
right after
`foreman_->join();`

The reason is that `query_context_` is not a local pointer, its life span covers the execution of multiple queries. It is possible that at the time of block eviction, some data structures (e.g. `JoinHashTable`) in `query_context_` still holds `CountedReference` to the block.

zuyu:
@jianqiao The bug is not related to `QueryContext` at all.

`QueryContext` for now is per query (rather than what you said that across multiple queries), and resets upon a new query has generated the optimal execution plan.

On the other hand, `JoinHashTable` stores in a blob, not a block (which has a reference count bug to fix). Let alone the query execution `DAG` should have `DestroyHashOperator` to drop the hash table immediately after the `HashJoinOperator`, if there is a hash-based join.

PR #63, opened 2016-03-05
Use direct initilization for atomic StorageBlockBase::ref_count_

pateljm:
Use direct initialization for the atomic ref_count_ instead of the previous copy initialization.


zuyu:
I suspect that this fix (assigning the ref count to zero at the beginning) would work, as it seems having issues to call `unref` to reduce the count to zero.

pateljm:
@zuyu Great. Lets then merge this PR. Thanks!

hbdeshmukh:
@pateljm  As discussed, this PR looks good to me, merging.

PR #65, opened 2016-03-05
Add a helper function for instantiating templates that have many bool parameters.

jianqiao:
Implemented the `CreateBoolInstantiatedInstance` function for alleviating the tedious work of instantiating templates that have many bool parameters (say >= 4).

This helper function will be used in the upcoming string pattern matching PR.

The scenario for using this helper function is that, suppose we have a class where all template parameters are bools:
```
template <bool c1, bool c2, bool c3>
class SomeClass : public BaseClass {
  // This simple function will be invoked in computationally-intensive loops.
  inline SomeType someSimpleFunction(...) {
    if (c1) {
      doSomeThing1();
    }
    if (c2) {
      doSomeThing2();
    }
    if (c3) {
      doSomeThing3();
    }
  }
};
```
Typically, this bool-parameterized template is for performance consideration.
That is, we would like to make a copy of code for each configuration of bool values, so that there will be no branchings in someSimpleFunction().

The problem is that, to conditionally instantiate the template, given bool variables c1, c2, c3, we have to do something like this:
```
if (c1) {
  if (c2) {
    if (c3) {
      return new SomeClass<true, true, true>(some_args...);
    } else {
      return new SomeClass<true, true, false>(some_args...);
    }
  } else {
    if (c3) {
      return new SomeClass<true, false, true>(some_args...);
    } else {
      return new SomeClass<true, false, false>(some_args...);
    }
  }
} else {
  ...
}
```
Then there will be power(2,N) branches if the template has N bool parameters, making it tedious to do the instantiation.

Now, this helper function can achieve the branched instantiation in one
statement as:
```
return CreateBoolInstantiatedInstance<SomeClass,BaseClass>(
    std::forward_as_tuple(some_args...), c1, c2, c3);
```

### DIFF
--- a/utility/CMakeLists.txt
+++ b/utility/CMakeLists.txt
@@ -351,6 +351,7 @@ target_link_libraries(TemplateUtil_unittest
                       glog
                       gtest
                       gtest_main
+                      quickstep_utility_Macros
                       quickstep_utility_TemplateUtil)
 add_test(TemplateUtil_unittest TemplateUtil_unittest)
 
--- a/utility/CMakeLists.txt
+++ b/utility/CMakeLists.txt
@@ -165,6 +165,7 @@ add_library(quickstep_utility_StringUtil StringUtil.cpp StringUtil.hpp)
 # Note that TextBasedTest.{hpp, cpp} are not in this static library.
 # Any tests that use them need to include them in the
 # executable.
+add_library(quickstep_utility_TemplateUtil ../empty_src.cpp TemplateUtil.hpp)
 add_library(quickstep_utility_TextBasedTestDriver
             textbased_test/TextBasedTestDriver.cpp
             textbased_test/TextBasedTestDriver.hpp
@@ -225,6 +226,7 @@ target_link_libraries(quickstep_utility_ShardedLockManager
                       quickstep_utility_Macros)
 target_link_libraries(quickstep_utility_StringUtil
                       glog)
+target_link_libraries(quickstep_utility_TemplateUtil)
 target_link_libraries(quickstep_utility_TextBasedTestDriver
                       glog
                       gtest
@@ -263,6 +265,7 @@ target_link_libraries(quickstep_utility
                       quickstep_utility_SortConfiguration_proto
                       quickstep_utility_SqlError
                       quickstep_utility_StringUtil
+                      quickstep_utility_TemplateUtil
                       quickstep_utility_TextBasedTestDriver
                       quickstep_utility_ThreadSafeQueue
                       quickstep_utility_TreeStringSerializable
@@ -342,6 +345,14 @@ target_link_libraries(TreeStringSerializable_unittest
                       gtest_main
                       quickstep_utility_TreeStringSerializable)
 add_test(TreeStringSerializable_unittest TreeStringSerializable_unittest)
+
+add_executable(TemplateUtil_unittest "${CMAKE_CURRENT_SOURCE_DIR}/tests/TemplateUtil_unittest.cpp")
+target_link_libraries(TemplateUtil_unittest
+                      glog
+                      gtest
+                      gtest_main
+                      quickstep_utility_TemplateUtil)
+add_test(TemplateUtil_unittest TemplateUtil_unittest)
 
 # Benchmarks:
 if (UNIX)
--- a/utility/TemplateUtil.hpp
+++ b/utility/TemplateUtil.hpp
@@ -41,7 +41,7 @@ struct Seq {};
  *        Seq<1,2,...,N>.
  *
  * GenSeq is defined here for C++11 compatibility. For C++14 and above,
- * std::make_interger_sequence can be used to achieve the same functionality.
+ * std::make_integer_sequence can be used to achieve the same functionality.
  */
 template<std::size_t N, std::size_t ...S>
 struct GenSeq : GenSeq<N-1, N-1, S...> {};
--- a/utility/TemplateUtil.hpp
+++ b/utility/TemplateUtil.hpp
@@ -140,8 +140,7 @@ inline ReturnT* CreateBoolInstantiatedInstance(Tuple&& args) {
 template <template <bool ...> class T, class ReturnT,
           bool ...bool_values, typename ...Bools,
           typename Tuple>
-inline ReturnT* CreateBoolInstantiatedInstance(
-    Tuple&& args, bool tparam, Bools ...rest_tparams) {
+inline ReturnT* CreateBoolInstantiatedInstance(Tuple&& args, bool tparam, Bools ...rest_tparams) {
   if (tparam) {
     return CreateBoolInstantiatedInstance<T, ReturnT, bool_values..., true>(
         std::forward<Tuple>(args), rest_tparams...);
--- a/utility/TemplateUtil.hpp
+++ b/utility/TemplateUtil.hpp
@@ -1,0 +1,157 @@
+/**
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *   University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#ifndef QUICKSTEP_UTILITY_TEMPLATE_UTIL_HPP_
+#define QUICKSTEP_UTILITY_TEMPLATE_UTIL_HPP_
+
+#include <utility>
+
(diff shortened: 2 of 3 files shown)


pateljm:
Nice @jianqiao! LTGM a soon as Travis comes back green.

pateljm:
Merging.

zuyu:
@jianqiao Sorry for the late comments, but there are style issues to fix.

utility/TemplateUtil.hpp
```diff
@@ -0,0 +1,157 @@
...
+ *  @{
+ */
+
+/**
+ * @brief Represents a compile-time sequence of integers.
+ *
+ * Seq is defined here for C++11 compatibility. For C++14 and above,
+ * std::integer_sequence can be used to achieve the same functionality.
+ */
+template<std::size_t ...>
```
zuyu: Please include `<cstddef>` to use `std::size_t`.

utility/TemplateUtil.hpp
```diff
@@ -0,0 +1,157 @@
...
+struct Seq {};
```
zuyu: Better to use the full word `Sequence`, instead of `Seq`.

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <memory>
+#include <string>
+#include <sstream>
+#include <tuple>
+#include <utility>
+
+#include "utility/Macros.hpp"
+
+#include "glog/logging.h"
```
zuyu: Please remove this unused header.

utility/CMakeLists.txt
```diff
@@ -165,6 +165,7 @@ add_library(quickstep_utility_StringUtil StringUtil.cpp StringUtil.hpp)
 # Note that TextBasedTest.{hpp, cpp} are not in this static library.
 # Any tests that use them need to include them in the
 # executable.
+add_library(quickstep_utility_TemplateUtil ../empty_src.cpp TemplateUtil.hpp)
```
zuyu: This line should be added before the above comment.

utility/TemplateUtil.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+
+/**
+ * @brief Final step of CreateBoolInstantiatedInstance. Now all bool_values are
+ *        ready. Instantiate the template and create (i.e. new) an instance.
+ */
+template <template <bool ...> class T, class ReturnT,
+          bool ...bool_values, std::size_t ...i,
+          typename Tuple>
+inline ReturnT* CreateBoolInstantiatedInstanceInner(Tuple&& args, Seq<i...>&& indices) {
```
zuyu: For the rvalue arguments like `args` and `indices`, `&&` should be next to the argument names, instead of their types.

utility/TemplateUtil.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * @brief Edge case of the recursive CreateBoolInstantiatedInstance function
+ *        when all bool variables have been branched and replaced with compile-time
+ *        bool constants.
+ */
+template <template <bool ...> class T, class ReturnT,
+          bool ...bool_values,
+          typename Tuple>
+inline ReturnT* CreateBoolInstantiatedInstance(Tuple&& args) {
```
zuyu: See above for the `rvalue` argument `args`.

utility/TemplateUtil.hpp
```diff
@@ -0,0 +1,157 @@
...
+ * statement as:
+ * --
+ * return CreateBoolInstantiatedInstance<SomeClass,BaseClass>(
+ *     std::forward_as_tuple(some_args...), c1, c2, c3);
+ * --
+ */
+template <template <bool ...> class T, class ReturnT,
+          bool ...bool_values, typename ...Bools,
+          typename Tuple>
+inline ReturnT* CreateBoolInstantiatedInstance(Tuple&& args, bool tparam, Bools ...rest_tparams) {
```
zuyu: Same for the rvalue argument `args`.

Mark the rest two arguments `const`.

jianqiao:
@zuyu Thanks! Will send a new PR to fix the style issues.

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "gtest/gtest.h"
+
+namespace quickstep {
+
+class SomeArgType {
+ public:
+  explicit SomeArgType(std::string value) : value_(value) {
```
zuyu: The constructor should look like or similar:

```
  explicit SomeArgType(const std::string &value)
      : value_(value) {}
```

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  SomeArgType(SomeArgType &&arg) {
```
zuyu: Mark this constructor `explicit`.

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+    value_ = std::move(arg.value_);
+  }
```
zuyu: This move constructor should look like as following:

```
  SomeArgType(SomeArgType &&orig)
      : value_(std::move(orig.value_)) {}
```

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  std::string toString() const {
+    return value_;
+  }
+
+ private:
+  std::string value_;
```
zuyu: Mark this member `const`.

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+class BaseClass {
+ public:
+  virtual std::string toString() const = 0;
+};
+
+template <bool c1, bool c2, bool c3, bool c4, bool c5, bool c6>
+class SomeClass : public BaseClass {
+ public:
+  SomeClass(int a1, SomeArgType &&a2)
```
zuyu: Mark `a1` `const`.

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (c6) {
+      oss << "c6 ";
+    }
+    oss << "} " << a1_ << " " << a2_.toString();
+    return oss.str();
+  }
+
+ private:
+  int a1_;
+  SomeArgType a2_;
```
zuyu: Mark above two private members `const`.

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+};
+
+void RunTest(bool c1, bool c2, bool c3, bool c4, bool c5, bool c6, std::string expected) {
```
zuyu: We need to mark the arguments as `const` and avoid a deep copy of a string:

```
void RunTest(const bool c1, const bool c2, const bool c3, const bool c4, const bool c5, const bool c6,
             const std::string &expected);
```

utility/tests/TemplateUtil_unittest.cpp
```diff
@@ -0,0 +1,111 @@
...
+  // arg should be perfectly forwarded.
+  SomeArgType arg("xyz");
+
+  std::unique_ptr<BaseClass> base(
+      CreateBoolInstantiatedInstance<SomeClass, BaseClass>(std::forward_as_tuple(10, std::move(arg)),
+                                                           c1, c2, c3, c4, c5, c6));
+  EXPECT_TRUE(base->toString() == expected);
```
zuyu: Replace the expect with the following, where the left part is the expected value, while the right the checked one:

```
EXPECT_STREQ(expected.c_str(), base->toString().c_str());
```

utility/CMakeLists.txt
```diff
@@ -343,6 +346,15 @@ target_link_libraries(TreeStringSerializable_unittest
                       quickstep_utility_TreeStringSerializable)
 add_test(TreeStringSerializable_unittest TreeStringSerializable_unittest)
 
+add_executable(TemplateUtil_unittest "${CMAKE_CURRENT_SOURCE_DIR}/tests/TemplateUtil_unittest.cpp")
+target_link_libraries(TemplateUtil_unittest
+                      glog
```
zuyu: Remove `glog` as you don't use it in the code.

utility/TemplateUtil.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * @brief The helper class for creating Seq. GenSeq<N>::type is equivalent to
+ *        Seq<1,2,...,N>.
+ *
+ * GenSeq is defined here for C++11 compatibility. For C++14 and above,
+ * std::make_integer_sequence can be used to achieve the same functionality.
+ */
+template<std::size_t N, std::size_t ...S>
+struct GenSeq : GenSeq<N-1, N-1, S...> {};
```
zuyu: A better name `GeneratedSequence`?

FYI, `/CMakeLists.txt` detects whether a compiler supports `cxx14`. Therefore, you may want to add a `TODO` or directly use `std::make_integer_sequence` if having `cxx14` support. You could refer to `/threading/ThreadingConfig.h.in` for `QUICKSTEP_HAVE_CPP14_SHARED_TIMED_MUTEX`.

PR #72, opened 2016-03-08
execution generator test format SQL style

jianqiao:
This PR formats the SQL queries in `execution_generator` test so that the queries are structured in an easy-to-read way.


cramja:
LGTM

PR #76, opened 2016-03-09
Introduced ForemanLite.

zuyu:
This PR introduces `ForemanLite`, the base class for various `Foreman` implementations for different cases, including the single node version and the distributed one in the near future.

It also has some clean-ups in `Foreman` unit test.

### DIFF
--- a/query_execution/ForemanLite.hpp
+++ b/query_execution/ForemanLite.hpp
@@ -1,0 +1,85 @@
+/**
+ *   Copyright 2016 Pivotal Software, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pateljm:
Nice @zuyu. LTGM. @hbdeshmukh Looks ok to you?

query_execution/ForemanLite.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include "tmb/message_bus.h"
+
+namespace quickstep {
+
+/** \addtogroup QueryExecution
+ *  @{
+ */
+
+/**
+ * @brief A base class that Foreman implements. This class is used to derive
```
hbdeshmukh: ``A base class that Foreman implements.`` 

This is somewhat odd to read. The base class description should be generic enough, irrespective of the derived class.

hbdeshmukh:
Hi @zuyu @pateljm Looks good to me, except one minor comment on the doxygen of ``ForemanLite``. 
Admittedly, I am bad at naming things, but I think we should have a more descriptive name than ``ForemanLite`` to this class.

query_execution/ForemanLite.hpp
zuyu: This comment is derived from `/storage/InsertDestinationInterface` by Craig.

```
@brief A pure interface class that InsertDestination and its subclasses implement.
```

pateljm:
Merging.

zuyu:
@hbdeshmukh I named it `ForemanLite`, as the `Lite` suffix is commonly used for a base class, for example, in `protobuf`.

PR #81, opened 2016-03-12
Refactored WorkerMessage Factory Methods.

zuyu:
This PR refactored `WorkerMessage` factory methods and thus simplified code in `Foreman`.

### DIFF
--- a/cli/QuickstepCli.cpp
+++ b/cli/QuickstepCli.cpp
@@ -361,12 +361,10 @@ int main(int argc, char* argv[]) {
   style.Broadcast(true);
   Address address;
   address.All(true);
-  WorkerMessage poison_message(WorkerMessage::PoisonMessage());
-  TaggedMessage poison_tagged_message;
-  poison_tagged_message.set_message(
-      &poison_message,
-      sizeof(poison_message),
-      quickstep::kPoisonMessage);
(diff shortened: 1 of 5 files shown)


pateljm:
LTGM. Merging.

hbdeshmukh:
@zuyu The three methods in WorkerMessage whose return type was changed from ``WorkerMessage`` to ``WorkerMessage*`` were actually [named constructors](https://isocpp.org/wiki/faq/ctors#named-ctor-idiom). If their return types are ``WorkerMessage*``, they won't be named constructors anymore. 

Also, I don't really understand why this change is needed. Can you expand on that?

zuyu:
@hbdeshmukh I've changed them to `Factory Method`s, and thus avoid implicit `move` semantics you used and simplifies code.

hbdeshmukh:
Was there any performance issue with the move semantics? Each message object is not heavy, just a couple of integers.

zuyu:
But it is hard to use. That was why you had to add `Foreman::generateWorkerMessage` for the factory methods.

hbdeshmukh:
Okay. The factory methods are not named constructors anymore, because of their return types.

zuyu:
@hbdeshmukh Thanks for pointing it out, and I've fixed the comments in #88.

Issue #70, opened 2016-03-07
Deadlock Caused by makeRoomForBlock


cramja:
I was running SSB queries this past weekend and noticed that some of them would hang indefinitely. 

To reproduce the issue:

1.  Use either a debug or production build of quickstep.

2. Use a large number of worker threads and the default buffer pool size. On the Quickstep box, this looks like: `./quickstep_cli_shell --num_workers=80 --printing_enabled=false`

3. I have a large (6gb) `lineorder` table from SSB already loaded into a shared folder on the Quickstep box if  you would like to use it.

```shell
$QS=your_qs_cli_shell
cd /home/spehlmann/shared
$QS --storage_path=qs_data
```
Now type this a bunch of times. It should hang on the 4-5th repetition. 
```SQL
SELECT COUNT(*) FROM lineorder;
```

**Notes:**
- Running this in script cannot reproduce the issue. As in, giving a bunch of `COUNT` queries to quickstep from a single file like `$QS < count.sql` will not reproduce the issue.
- Using less workers or a larger buffer pool eliminates the issue.

zuyu:
Well, please show the log outputs for the hanging case using the following comment:

```
GLOG_v=5 GLOG_logtostderr=1 /path/to/quickstep_cli_shell
```

cramja:
[I linked it](http://pages.cs.wisc.edu/~spehlmann/assets/tmp/qs-debug-out-lineorder-count-hang.txt) because it's a lot of text.

The file shows the output from a debug version of quickstep after issuing the `SELECT COUNT(*)` that causes the hang up. (Production version doesn't produce log messages).

How I am interpreting this output is that the optimizer successfully passes the query on to the foreman in the case of the 4th COUNT query which caused the hangup as it successfully creates an optimized physical plan: `I0308 14:25:31.845458 12279 PhysicalGenerator.cpp:88] Optimized physical plan:`

cramja:
@pateljm mentioned that he'd like to hold a bug hunting session where @saketj , @jianqiao and whoever else might be interested would go through and try to find the cause of this.

zuyu:
@cramja The logs you provided only shows the outputs from a parser tree to the **optimized physical plan** for the same query four times. (And yes, these outputs created using `DVLOG` won't print out in the release mode.) However, we need more info about this hanging bug after generating the physical plan.

What you need to do is, to add the logging statement in `query_execution`, including `Foreman` and `Worker`.

[...]

PR #85, opened 2016-03-14
Storage manager concurrency bug

cramja:
Fixes the deadlock-causing bug in the storage manager. Concludes #70 .

The essence of the bug was that threads acquired 2 locks during an eviction phase in the method `makeRoomForBlock`. Depending on the interleaving, thread A could request lock 1 while holding lock 2 and thread B could hold lock 2 while requesting 1, classic deadlock.

The change simply moves the call to makeRoomForBlock to a section of calling code where no locks already held, removing the possibility of deadlock. 

This move does not change the correctness of the StorageManager. Though it is possible to get an interleaving which will result in temporarily over-shooting the size limit of the StorageManager, this was possible in the previous code as well. It does not present a serious issue.

**Note:** 
On rare occasions (one out of 50) when I ran the ctest suite on the quickstep box, I got a RefCount bug

```
quickstep_queryoptimizer_tests_ExecutionGeneratorTest: /home/spehlmann/qs/storage/EvictionPolicy.cpp:411:
void quickstep::LRUKEvictionPolicyImpl<K>::blockEvicted(quickstep::block_id) 
[with long unsigned int K = 2ul; quickstep::block_id = long unsigned int]: 
Assertion `blocks_[block].ref_count_ == 0' failed.
```

Which I do not think is related to this patch.


zuyu:
@cramja I was surprised that you observed that the block reference counting bug happens so frequently (one of 50) and exists outside of the CI builds. Could you please create a new issue for this, and describe how to reproduce it? Thank you!

As @craig-chasseur (welcome back) mentioned in the mail, he also has not seen in elsewhere except in `Travis`.

query_optimizer/tests/execution_generator/CMakeLists.txt
```diff
@@ -17,39 +17,39 @@ add_test(quickstep_queryoptimizer_tests_executiongenerator_create
          "../quickstep_queryoptimizer_tests_ExecutionGeneratorTest"
          "${CMAKE_CURRENT_SOURCE_DIR}/Create.test"
          "${CMAKE_CURRENT_BINARY_DIR}/Create.test"
-         "${CMAKE_CURRENT_BINARY_DIR}/")
+         "${CMAKE_CURRENT_BINARY_DIR}/Create/")
```
zuyu: The changes in this file are not related to this bug. Please reset it back.

storage/FileManagerPosix.cpp
```diff
@@ -86,6 +88,7 @@ size_t FileManagerPosix::numSlots(const block_id block) const {
   }
 
   if ((file_stat.st_size % kSlotSizeBytes) != 0) {
+    LOG(ERROR) << "Size of file " << filename.c_str() << " was not integral.";
```
zuyu: This log is not related to this bug. Please remove it.

As an alternative, please remove the below exception, and use `LOG(FATAL)` instead.

storage/tests/StorageManager_unittest.cpp
```diff
@@ -205,46 +205,4 @@ TEST(StorageManagerTest, DifferentNUMANodeBlobTestWithEviction) {
 }
 #endif  // QUICKSTEP_HAVE_LIBNUMA
 
-// Trigger an eviction from the same shard in StorageManager's
```
zuyu: Why remove this test?

storage/StorageManager.hpp
```diff
@@ -208,6 +208,8 @@ class StorageManager {
    **/
   bool blockOrBlobIsLoaded(const block_id block) const;
 
+  bool saveBlockOrBlobLocked(const block_id block, const bool force);
```
zuyu: Add the doxygen comments for this new method.

query_optimizer/tests/execution_generator/CMakeLists.txt
jianqiao: This is a fix to another bug that was discovered during the testing of the concurrency bug.
@cramja Perhaps just reset it here and let me quickly send a separate PR to address this.

zuyu:
@cramja If the block ref counting bug in `EvictionPolicy` happens so frequently after changes in this PR, we should not merge it.

query_optimizer/tests/execution_generator/CMakeLists.txt
zuyu: @jianqiao What's bug then? Each `quickstep_queryoptimizer_tests_ExecutionGeneratorTest` should clean up the environment after running.

cramja: When run concurrently, two tests can manipulate the same `.qsb` file. In this case, we found that it threw an exception in the [FileManager](https://github.com/pivotalsoftware/quickstep/blob/master/storage/FileManagerPosix.cpp#L89). We fixed this problem by having the tests create their `.qsb` files in separate directories.

storage/FileManagerPosix.cpp
cramja: As above, this is related to the secondary bug which is mentioned above. I like the suggestion to change to LOG(FATAL)- addressed.

query_optimizer/tests/execution_generator/CMakeLists.txt
zuyu: @cramja How did you run the tests concurrently?

storage/tests/StorageManager_unittest.cpp
cramja: @zuyu It was because the scenario for which it was designed to test was eliminated by the changes to StorageManager. It was testing a thread interleaving situation where a thread would try to acquire a lock which it already owned. 

When refactoring `makeRoomForBlock` there wouldn't be an attempt to acquire that a second lock. We removed the argument in `makeRoomForBlock(... , const block_id locked_block_id)` because the caller would never have a locked block going in. The unit test was designed to capture that exact scenario.

query_optimizer/tests/execution_generator/CMakeLists.txt
jianqiao: @zuyu `ctest -j<num-of-threads>`, e.g. running with 80 threads: `ctest -j80`.
I just sent a separate PR #86 for this.

storage/StorageManager.hpp
cramja: Addressed.

cramja:
@zuyu The way I tested for the bug was, using a simple python script:

```python
import sys
import os

def main(args):
  r = 0
  trials = 0
  MAX = 100
  out_file = "out.txt"
  while(r == 0 and trials < MAX):
    try:
      os.remove(out_file)
    except OSError:
      pass
    r = os.system("ctest -j80 --output-on-failure 1>{0}".format(out_file))
    trials += 1
  if trials == MAX:
    print "pass"
    return 0
  else:
    print "failed after {0}".format(trials)
    return -1  

if __name__ == '__main__':
  main(sys.argv)

```

It actually failed after 1 trial and I thought that was crazy so I ran it again for 50 times before killing it without seeing the error. So who knows.

query_optimizer/tests/execution_generator/CMakeLists.txt
zuyu: @jianqiao Oh, cool. It makes sense!

zuyu:
@cramja Then run the script more times to figure out why the bug happens. Does it happen in the current `master` branch as well?

storage/StorageManager.hpp
```diff
@@ -209,6 +209,21 @@ class StorageManager {
   bool blockOrBlobIsLoaded(const block_id block) const;
 
   /**
+   * @brief Saves a block or blob in memory to the persistent storage. 
+   * @details Acquires a write lock on the specified block before saving.
```
zuyu: Change to `Save` and `Acquire`, respectively.

storage/tests/StorageManager_unittest.cpp
zuyu: Anyway, keeping this unit test does hurt anything, right?

cramja:
@zuyu I will run this overnight for master on the Quickstep box.

pateljm:
Thanks @jianqiao @cramja @zuyu for all the work and comments here. @cramja Sounds like you are seeing the refcount bug only when you are running your python script - can you confirm?

cramja:
@pateljm I ran the tests on this branch last night and the refcount bug did not make an appearance.

cramja:
@zuyu  As for what to do with this branch, since @jianqiao already merged changes to the `query_optimizer/tests/execution_generator/CMakeLists.txt` file, I can make this PR cleaner by making patches of only the code related to the deadlock bug and make a new PR/close this one.

storage/FileManagerPosix.cpp
```diff
@@ -86,7 +88,7 @@ size_t FileManagerPosix::numSlots(const block_id block) const {
   }
 
   if ((file_stat.st_size % kSlotSizeBytes) != 0) {
-    throw CorruptPersistentStorage();
+    LOG(FATAL) << "The file " << filename << " was corrupted.";
```
zuyu: If you change this file, change other two `FileManager` implementations as well.

zuyu:
@cramja So, what's the conclusion for the ref count bug in `EvictionPolicy` so far? The current master is fine, but it occurs sometimes in this branch?

Note that this is not the same ref count bug (in `StorageBlockBase`) discussed in the e-mail.

cramja:
@zuyu I'm not sure, I cannot reproduce it with the latest changes, after 100 trials of this branch, and 100 trials of master, neither had the problem occur (on the QS box)

relational_operators/SaveBlocksOperator.cpp
```diff
@@ -53,7 +53,7 @@ void SaveBlocksWorkOrder::execute() {
   // It may happen that the block gets saved to disk as a result of an eviction,
   // before this invocation. In either case, we don't care about the return
   // value of saveBlockOrBlob.
-  storage_manager_->saveBlockOrBlob(save_block_id_, force_);
+  storage_manager_->saveBlockOrBlobLocked(save_block_id_, force_);
```
zuyu: How about remaining the API `saveBlockOrBlob`, and then use `saveBlockOrBlobLocked` for the private implementation?

cramja: Okay, so swapping them around so that the Locked one is used internally to StorageManager. Sure, makes sense, will do.

storage/tests/StorageManager_unittest.cpp
cramja: @zuyu I will refactor it to test a similar situation, however the original situation will not happen.

cramja: Also, I noticed while refactoring, that when a new block is created (`StorageManager::createBlock`), that this process does not check if we are exceeding our maximum buffer pool size. (previous to this PR, it did not check).

Do we want to change this?

There's a convenient spot in `StorageManager::createBlock` where we have a write lock on the buffer pool:
```c++
  {
    // Lock 'blocks_shared_mutex_' as briefly as possible to insert an entry
    // for the new block, after the block has already been constructed.
    SpinSharedMutexExclusiveLock<false> write_lock(blocks_shared_mutex_);

    // Because block IDs are generated by atomically incrementing block_index_,
    // there should never be collisions.
    DCHECK(blocks_.find(new_block_id) == blocks_.end());
    blocks_[new_block_id] = new_block_handle;

    // We could add this line:
    makeRoomForBlock(0);
    // Which would evict blocks until we go below the system-defined buffer pool size.
```

storage/FileManagerPosix.cpp
cramja: Gotcha, on it!

pateljm:
I'd suggest that once all the test pass, we close this PR. If the ref count bug shows up again, we can deal with it at that point. The changes here look good to me so OK from my side.

cramja:
@pateljm Sounds good, I will complete this today and comment when finished.

cramja:
If the travis status comes back OK, this is ready to merge.

cramja:
Appears that travis failed internally, making a dummy commit to force Travis to restart.

cramja:
To avoid further merges with master and because we have agreed that it's good to go, I will merge this PR once Travis returns.

zuyu:
@cramja Before you merge the PR, squash all the commits into one, if possible.

storage/FileManagerPosix.cpp
zuyu: Actually, I am thinking the following implementations in all `FileManager` versions:

```
DCHECK_EQ(0, file_stat.st_size % kSlotSizeBytes)
  << "The file " << filename << " was corrupted.";
```

storage/StorageManager.cpp
```diff
@@ -407,7 +412,7 @@ void* StorageManager::allocateSlots(const std::size_t num_slots,
 
 #if defined(QUICKSTEP_HAVE_LIBNUMA)
   if (numa_node != -1) {
-    DEBUG_ASSERT(numa_node < numa_num_configured_nodes());
+    DCHECK(numa_node < numa_num_configured_nodes());
```
zuyu: Change to `DCHECK_LT`.

storage/StorageManager.cpp
```diff
@@ -444,14 +449,21 @@ MutableBlockReference StorageManager::getBlockInternal(
     const block_id block,
     const CatalogRelationSchema &relation,
     const int numa_node) {
+  std::size_t num_slots;
```
zuyu: Initialize `num_slots` with `0u`.

storage/StorageManager.cpp
```diff
@@ -461,13 +473,18 @@ MutableBlockReference StorageManager::getBlockInternal(
   // doesn't know about the block until blockReferenced is called, so
   // chooseBlockToEvict shouldn't return the block.
   if (!ret.valid()) {
+    // Call a best-effort method to evict blocks until the size of our buffer
+    // pool falls below the current buffer pool size plus the size of the
+    // block we are going to retrieve.
+    makeRoomForBlock(num_slots);
```
zuyu: This line should move after `Line 409`, as this is another check for whether the block has been loaded.

storage/tests/StorageManager_unittest.cpp
zuyu: You may want to check the full buffer pool in another PR.

storage/StorageManager.cpp
cramja: Addressed.

zuyu:
Re-fixed by #183.

PR #88, opened 2016-03-15
Fixed the comments in WorkerMessage.

zuyu:
This PR corrected the comments missed in #81.


query_execution/WorkerMessage.hpp
```diff
@@ -37,7 +37,7 @@ class WorkerMessage {
   };
 
   /**
-   * @brief A static named constructor for generating rebuild WorkOrder messages.
+   * @brief A static factory method for generating rebuild WorkOrder messages.
```
hbdeshmukh: Need @return doxygen for all the three methods. Earlier it was a constructor, so we didn't need it.

zuyu: Fixed.

hbdeshmukh:
Looks good to me. Merging.

PR #89, opened 2016-03-16
Added blocks to InsertDestination directly.

zuyu:
As a part of detaching `CatalogRelation` from `InsertDestination`, this PR adds blocks directly when constructing `InsertDestination`s.


pateljm:
LTGM. Merging.

PR #93, opened 2016-03-19
Refactored PartitionScheme

zuyu:
This PR refactors `PartitionScheme`, which thus contains `PartitionSchemeHeader` and partitioned blocks. `PartitionSchemeHeader` is also used in `PartitionAwareInsertDestination` to get partition information without accessing `CatalogRelation` stored in the master `Catalog`.

This PR also improves the algorithm that gets the partition id in `RangePartitionScheme` from `O(n)` using the linear scan to `O(log n)` using the binary search.


zuyu:
Replaced by #95.

PR #95, opened 2016-03-21
Refactored PartitionScheme

zuyu:
This PR refactors `PartitionScheme` into both `PartitionSchemeHeader` and partitioned blocks, and thus `PartitionAwareInsertDestination` could use reconstructed `PartitionSchemeHeader` from `proto` to get the partition information without accessing `CatalogRelation` stored in the master `Catalog`. Note that there will be a follow-up PR to replace `CatalogRelation` in `InsertDestination` with the read-only `CatalogRelationSchema`.

This PR also improves the algorithm that gets the partition id in `RangePartitionScheme` from `O(n)` using the linear scan to `O(log n)` using the binary search.


pateljm:
LTGM. Merging.

PR #98, opened 2016-03-24
Use TMB for Updating Catalog in InsertDestination

zuyu:
This PR refactored `InsertDestination` so that it only accesses the read-only `CatalogRelationSchema` and any `Catalog` updates need to use `TMB` messages. The assumption is that `InsertDestination` is remote to the master `Catalog`, no matter in the single-node or the distributed version.

This PR reorders the arguments in `InsertDestination` constructors based on their properties (e.g., `const`, reference or pointer) and usage (used by setting `TMB` message info).

Note that `Foreman` client id is used on behalf of `InsertDestination` when sending messages to `Foreman`. And in the distributed case, we will use `Shiftboss` one instead.


pateljm:
LTGM. Merging.

PR #99, opened 2016-03-28
Extracted CatalogRelationSchema proto from CatalogRelation proto

zuyu:
This small PR extracted `CatalogRelationSchema` proto from `CatalogRelation` proto so that the latter is derived from the former, as `CatalogRelation` is a subclass of `CatalogRelationSchema`. Note that there will be some follow-up PRs, including refactoring `CatalogDatabase` and adding `CatalogCache`.

It also added the missing serialization for the NUMA placement scheme.


pateljm:
LTGM. Merging.

PR #100, opened 2016-03-28
Transaction Part 2: AccessMode, ResourceId, Lock

hakanmemisoglu:
This PR introduces new classes for transaction support.

- AccessMode: Represents access level of the lock such as S, X, IS, IX. Based on [[Gray et al.]](http://research.microsoft.com/en-us/um/people/gray/papers/Granularity%20of%20Locks%20and%20Degrees%20of%20Consistency%20RJ%201654.pdf)

- ResourceId: Each logical and physical section (database, relation, block) of DBMS has a unique resource id associated with it.

- Lock: Since the concurrency control will be based on Strict 2PL, we will need locks. In a simple manner, it is a tuple of AccessMode and ResourceId.

### DIFF
--- a/catalog/CatalogTypedefs.hpp
+++ b/catalog/CatalogTypedefs.hpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -39,6 +41,10 @@ typedef int numa_node_id;
 // This depends on all the above id types being typedefed to int, except for
 // partition_id.
 const int kCatalogMaxID = INT_MAX;
+
+// Catalog ids use negative values as invalid ids. Mark -1 as constant invalid
+// id for the catalog ids.
+constexpr int kInvalidCatalogId = -1;
 
 /** @} */
 
--- a/transaction/AccessMode.hpp
+++ b/transaction/AccessMode.hpp
@@ -1,0 +1,171 @@
+/**
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#ifndef QUICKSTEP_TRANSACTION_ACCESS_MODE_HPP_
+#define QUICKSTEP_TRANSACTION_ACCESS_MODE_HPP_
+
+#include <cstdint>
+#include <type_traits>
+
+namespace quickstep {
+namespace transaction {
+
+/** \addtogroup Transaction
+ *  @{
+ */
+
+/**
+ * @brief Represents mode type. Possible options are NL, IS, IX, S, SIX, X.
+ **/
+enum class AccessModeType : std::uint8_t {
+  kNoLock = 0,
+  kIsLock,
+  kIxLock,
+  kSLock,
+  kSixLock,
+  kXLock,
+  kNumAccessModeTypes,
+};
+
+/**
+ * @brief Class for representing resource lock's access mode
+ **/
+class AccessMode {
+ public:
+  /**
+   * @brief Only constructor for access mode.
+   *
+   * @param access_mode Mode type of the object.
+   **/
+  explicit AccessMode(const AccessModeType access_mode)
+      : access_mode_(access_mode) {}
+
+  /**
+   * @brief Checks whether this access mode is compatible with the other.
+   *
+   * @param other Other access mode that will be checked against to this one.
+   * @return True if they are compatible, otherwise false.
+   **/
+  inline bool isCompatible(const AccessMode &other) const {
+    const access_mode_underlying_type this_mode =
+        static_cast<access_mode_underlying_type>(access_mode_);
+    const access_mode_underlying_type other_mode =
+        static_cast<access_mode_underlying_type>(other.access_mode_);
+    return AccessMode::kLockCompatibilityMatrix[this_mode][other_mode];
+  }
+
+  /**
+   * @brief Checks whether this access mode is IS mode.
+   *
+   * @return True if it is IS mode, false otherwise.
+   **/
+  inline bool isIntentionShareLock() const {
+    return access_mode_ == AccessModeType::kIsLock;
+  }
+
+  /**
+   * @brief Checks whether this access mode is IX mode.
+   *
+   * @return True if it is IX mode, false otherwise.
+   **/
+  inline bool isIntentionExclusiveLock() const {
+    return access_mode_ == AccessModeType::kIxLock;
+  }
+
+  /**
+   * @brief Checks whether this access mdoe is SIX mode.
+   *
+   * @return True if it is SIX mode, false otherwise.
+   **/
+  inline bool isShareAndIntentionExclusiveLock() const {
+    return access_mode_ == AccessModeType::kSixLock;
+  }
+
+  /**
+   * @brief Checks whether this access mode is S mode.
+   *
+   * @return True if it is S mode, false otherwise.
+   **/
+  inline bool isShareLock() const {
+    return access_mode_ == AccessModeType::kSLock;
+  }
+
+  /**
+   * @brief Checks whether this access mode is X mode.
+   *
+   * @return True if it is X mode, false otherwise.
+   **/
+  inline bool isExclusiveLock() const {
+    return access_mode_ == AccessModeType::kXLock;
+  }
+
+  /**
+   * @brief Checks whether this access mode is in
+   *        the same level with other mode.
+   *
+   * @return True if both modes have the same level.
+   **/
+  inline bool operator==(const AccessMode &other) const {
+    return access_mode_ == other.access_mode_;
+  }
+
+  /**
+   * @brief Checks whether this access mode is in
+   *        the different level with other mode.
+   *
+   * @return True if the modes have different levels.
+   **/
+  inline bool operator!=(const AccessMode &other) const {
+    return access_mode_ != other.access_mode_;
+  }
+
+ private:
+  typedef std::underlying_type<AccessModeType>::type
+      access_mode_underlying_type;
+
+  // The compatibility matrix should be N by N. kNumberLocks == N.
+  static constexpr std::uint64_t kNumberLocks =
+      static_cast<access_mode_underlying_type>(
+          AccessModeType::kNumAccessModeTypes);
+
+  // Compatibility matrix for checking access modes.
+  // True means they are compatible.
+  static constexpr bool kLockCompatibilityMatrix[kNumberLocks][kNumberLocks] = {
+/*           NL     IS     IX      S     SIX     X    */
+/*  NL  */ {true , true , true , true , true , true },
+/*  IS  */ {true , true , true , true , true , false},
+/*  IX  */ {true , true , true , false, false, false},
+/*  S   */ {true , true , false, true , false, false},
+/*  SIX */ {true , true , false, false, false, false},
+/*  X   */ {true , false, false, false, false, false}
+  };
+
+  // Type of access, the possible values are
+  // NoLock, IsLock, IxLock, SLock, SixLock, XLock
+  AccessModeType access_mode_;
+};
+
+/** @} */
+
+// static constexpr fields must be declared here, otherwise gcc
+// gives linkage errors.
+constexpr bool AccessMode::kLockCompatibilityMatrix[][kNumberLocks];
+
+}  // namespace transaction
+}  // namespace quickstep
+
+#endif  // QUICKSTEP_TRANSACTION_ACCESS_MODE_HPP_
--- a/transaction/CMakeLists.txt
+++ b/transaction/CMakeLists.txt
@@ -13,9 +13,18 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+add_library(quickstep_transaction_AccessMode
+            ../empty_src.cpp
+            AccessMode.hpp)
 add_library(quickstep_transaction_DirectedGraph
             ../empty_src.cpp
             DirectedGraph.hpp)
+add_library(quickstep_transaction_Lock
+            ../empty_src.cpp
+            Lock.hpp)
+add_library(quickstep_transaction_ResourceId
+            ResourceId.cpp
+            ResourceId.hpp)
 add_library(quickstep_transaction_StronglyConnectedComponents
             StronglyConnectedComponents.cpp
             StronglyConnectedComponents.hpp)
@@ -24,9 +33,18 @@ add_library(quickstep_transaction_Transaction
             Transaction.hpp)
 
 target_link_libraries(quickstep_transaction_DirectedGraph
-	              glog
+                      glog
                       quickstep_transaction_Transaction
                       quickstep_utility_Macros)
+target_link_libraries(quickstep_transaction_Lock
+                      quickstep_transaction_AccessMode
+                      quickstep_transaction_ResourceId)
+target_link_libraries(quickstep_transaction_ResourceId
+                      quickstep_catalog_CatalogTypedefs
+                      quickstep_storage_StorageBlockInfo
+                      quickstep_utility_HashPair)
+target_link_libraries(quickstep_transaction_ResourceId
+                      glog)
 target_link_libraries(quickstep_transaction_StronglyConnectedComponents
                       quickstep_transaction_DirectedGraph
                       quickstep_utility_Macros)
@@ -35,9 +53,20 @@ add_library(quickstep_transaction
             ../empty_src.cpp
             TransactionModule.hpp)
 target_link_libraries(quickstep_transaction
+                      quickstep_transaction_AccessMode
                       quickstep_transaction_DirectedGraph
+                      quickstep_transaction_Lock
+                      quickstep_transaction_ResourceId
                       quickstep_transaction_StronglyConnectedComponents
                       quickstep_transaction_Transaction)
+
+add_executable(AccessMode_unittest
+               "${CMAKE_CURRENT_SOURCE_DIR}/tests/AccessMode_unittest.cpp")
+target_link_libraries(AccessMode_unittest
+                      gtest
+                      gtest_main
+                      quickstep_transaction_AccessMode)
+add_test(AccessMode_unittest AccessMode_unittest)
 
 add_executable(DirectedGraph_unittest
                "${CMAKE_CURRENT_SOURCE_DIR}/tests/DirectedGraph_unittest.cpp")
@@ -47,6 +76,24 @@ target_link_libraries(DirectedGraph_unittest
                       quickstep_transaction_DirectedGraph
                       quickstep_transaction_Transaction)
 add_test(DirectedGraph_unittest DirectedGraph_unittest)
+
+add_executable(Lock_unittest
+               "${CMAKE_CURRENT_SOURCE_DIR}/tests/Lock_unittest.cpp")
+target_link_libraries(Lock_unittest
+               gtest
+               gtest_main
+               quickstep_transaction_AccessMode
+               quickstep_transaction_Lock
+               quickstep_transaction_ResourceId)
+add_test(Lock_unittest Lock_unittest)
+
+add_executable(ResourceId_unittest
+               "${CMAKE_CURRENT_SOURCE_DIR}/tests/ResourceId_unittest.cpp")
+target_link_libraries(ResourceId_unittest
+                      gtest
+                      gtest_main
+                      quickstep_transaction_ResourceId)
+add_test(ResourceId_unittest ResourceId_unittest)
 
 add_executable(StronglyConnectedComponents_unittest
                "${CMAKE_CURRENT_SOURCE_DIR}/tests/StronglyConnectedComponents_unittest.cpp")
--- a/transaction/Lock.hpp
+++ b/transaction/Lock.hpp
@@ -1,0 +1,103 @@
+/**
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#ifndef QUICKSTEP_TRANSACTION_LOCK_HPP_
+#define QUICKSTEP_TRANSACTION_LOCK_HPP_
+
+#include "transaction/AccessMode.hpp"
+#include "transaction/ResourceId.hpp"
+
+namespace quickstep {
+namespace transaction {
+
+/** \addtogroup Transaction
+ *  @{
+ */
+
+/**
+ * @brief Class for representing resource locks.
+ **/
+class Lock {
+ public:
+  /**
+   * @brief Constructor.
+   *
+   * @param rid Resource id of the resource on which the lock is taken.
+   * @param access_mode Mode of the lock access.
+   **/
+  Lock(const ResourceId &rid, const AccessMode &access_mode)
+      : rid_(rid),
+        access_mode_(access_mode) {
+  }
+
+  /**
+   * @brief Equality operator for lock.
+   *
+   * @param other Reference to the other Lock object that will be
+   *        compared against to this.
+   * @return True if both locks have equal resource ids
+   *         and access modes, false otherwise.
+   **/
+  inline bool operator==(const Lock &other) const {
+    return rid_ == other.rid_ && access_mode_ == other.access_mode_;
+  }
+
+  /**
+   * @brief Inequality operator for lock.
+   *
+   * @param other Reference to the other Lock object that will be
+   *        compared against to this.
+   * @return False if one of them has different resource id
+   *         or access mode, true otherwise.
+   **/
+  inline bool operator!=(const Lock &other) const {
+    return !(*this == other);
+  }
+
+  /**
+   * @brief Getter for resource id.
+   *
+   * @return Resource id of the lock.
+   **/
+  inline const ResourceId& getResourceId() const {
+    return rid_;
+  }
+
+  /**
+   * @brief Getter for access mode.
+   *
+   * @return Access mode of the lock.
+   **/
+  inline const AccessMode& getAccessMode() const {
+    return access_mode_;
+  }
+
+ private:
+  // Id of the resource that is locked.
+  const ResourceId rid_;
+
+  // Lock's access type.
+  AccessMode access_mode_;
+};
+
+/** @} */
+
+}  // namespace transaction
+
+}  // namespace quickstep
+
+#endif  // QUICKSTEP_TRANSACTION_LOCK_HPP_
--- a/transaction/ResourceId.cpp
+++ b/transaction/ResourceId.cpp
@@ -1,0 +1,74 @@
+/**
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "transaction/ResourceId.hpp"
+
+#include <cstddef>
+#include <functional>
+#include <string>
+
+#include "catalog/CatalogTypedefs.hpp"
+#include "storage/StorageBlockInfo.hpp"
+#include "utility/HashPair.hpp"
+
+#include "glog/logging.h"
+
+namespace quickstep {
+namespace transaction {
+
+bool ResourceId::operator==(const ResourceId &other) const {
+  return db_id_ == other.db_id_
+      && rel_id_ == other.rel_id_
+      && block_id_ == other.block_id_
+      && tuple_id_ == other.tuple_id_;
+}
+
+ResourceId ResourceId::getParentResourceId() const {
+  if (isRelationAccess()) {
+    return ResourceId(db_id_);
+  } else if (isBlockAccess()) {
+    return ResourceId(db_id_, rel_id_);
+  } else if (isTupleAccess()) {
+    return ResourceId(db_id_, rel_id_, block_id_);
+  } else {
+    LOG(FATAL) << "Database level does not have any parent level.";
+  }
+}
+
+std::string ResourceId::toString() const {
+  return "ResourceId(" +
+    std::to_string(db_id_) + ", " +
+    std::to_string(rel_id_) + ", " +
+    std::to_string(block_id_) + ", " +
+    std::to_string(tuple_id_) + ")";
+}
+
+std::size_t
+ResourceId::ResourceIdHasher::operator()(const ResourceId &rid) const {
+  const std::size_t hash1 = std::hash<database_id>()(rid.db_id_);
+  const std::size_t hash2 = std::hash<relation_id>()(rid.rel_id_);
+  const std::size_t hash3 = std::hash<block_id>()(rid.block_id_);
+  const std::size_t hash4 = std::hash<tuple_id>()(rid.tuple_id_);
+
+  const std::size_t comb1 = CombineHashes(hash1, hash2);
+  const std::size_t comb2 = CombineHashes(hash3, hash4);
+
+  return CombineHashes(comb1, comb2);
+}
+
+}  // namespace transaction
+}  // namespace quickstep
--- a/transaction/ResourceId.hpp
+++ b/transaction/ResourceId.hpp
@@ -1,0 +1,216 @@
+/**
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#ifndef QUICKSTEP_TRANSACTION_RESOURCE_ID_HPP_
+#define QUICKSTEP_TRANSACTION_RESOURCE_ID_HPP_
+
+#include <cstddef>
+#include <limits>
+#include <string>
+
+#include "catalog/CatalogTypedefs.hpp"
+#include "storage/StorageBlockInfo.hpp"
+
+namespace quickstep {
+namespace transaction {
+
+/** \addtogroup Transaction
+ *  @{
+ */
+
+/**
+ * @brief Class for representing unique id for each database resource
+ *        e.g Tuple, block, relation, database etc...
+ **/
+class ResourceId {
+ public:
+  /**
+   * @brief Constructor
+   *
+   * @param db_id Unique id of the database resource.
+   * @param rel_id Unique id of the relation resource in the database.
+   * @param block_id Unique id of the block resource in the relation.
+   * @param tuple_id Unique id of the tuple resource in the block.
+   **/
+  explicit ResourceId(const database_id db_id = kDatabaseIdPlaceholder,
+                      const relation_id rel_id = kRelationIdPlaceholder,
+                      const block_id block_id = kBlockIdPlaceholder,
+                      const tuple_id tuple_id = kTupleIdPlaceholder)
+      : db_id_(db_id),
+        rel_id_(rel_id),
+        block_id_(block_id),
+        tuple_id_(tuple_id) {
+  }
+
+  /**
+   * @brief Copy constructor.
+   *
+   * @param other The ResourceId that will be copied.
+   **/
+  ResourceId(const ResourceId &other)
+      : db_id_(other.db_id_),
+        rel_id_(other.rel_id_),
+        block_id_(other.block_id_),
+        tuple_id_(other.tuple_id_) {
+  }
+
+  /**
+   * @brief Hasher class for ResourceId class to use it in the hash map.
+   **/
+  struct ResourceIdHasher {
+    /**
+     * @brief Functor of the class.
+     *
+     * @param rid Resource id to be hashed.
+     * @return Hash of the resource id.
+     **/
+    std::size_t operator()(const ResourceId &rid) const;
+  };
+
+  /**
+   * @brief Equality operator for ResourceId class.
+   *
+   * @param other Resource to be compared.
+   * @return True if this and other are the id of the same
+   *         resource, false otherwise.
+   **/
+  bool operator==(const ResourceId &other) const;
+
+   /**
+   * @brief Inequality operator for ResourceId class.
+   *
+   * @param other Resource to be compared.
+   * @return True if this and other are the ids of the different
+   *         resources, false otherwise.
+   **/
+  inline bool operator!=(const ResourceId &other) const {
+    return !(*this == other);
+  }
+
+  /**
+   * @brief Checks whether this resource id has a parent in resource hierarchy.
+   *
+   * @return False if this resource id is database level, true otherwise.
+   **/
+  inline bool hasParent() const {
+    return !isDatabaseAccess();
+  }
+
+  /**
+   * @brief Getter for this resource id's parent in the hierarchy.
+   *
+   * @return The resource id of this resource id's parent.
+   **/
+  ResourceId getParentResourceId() const;
+
+  /**
+   * @brief Checks whether this is a database level resource id.
+   *
+   * @return True if this is a database level access, false otherwise.
+   **/
+  inline bool isDatabaseAccess() const {
+    return !isDatabaseIdPlaceholder()
+        && isRelationIdPlaceholder()
+        && isBlockIdPlaceholder()
+        && isTupleIdPlaceholder();
+  }
+
+  /**
+   * @brief Checks whether this is a relation level resource id.
+   *
+   * @return True if this is a relation level access, false otherwise.
+   **/
+  bool isRelationAccess() const {
+    return !isDatabaseIdPlaceholder()
+        && !isRelationIdPlaceholder()
+        && isBlockIdPlaceholder()
+        && isTupleIdPlaceholder();
+  }
+
+  /**
+   * @brief Checks whether this is a block level resource id.
+   *
+   * @return True if this is a block level access, false otherwise.
+   **/
+  bool isBlockAccess() const {
+    return !isDatabaseIdPlaceholder()
+        && !isRelationIdPlaceholder()
+        && !isBlockIdPlaceholder()
+        && isTupleIdPlaceholder();
+  }
+
+  /**
+   * @brief Checks whether this is a tuple level resource id.
+   *
+   * @return True if this is a tuple level access, false otherwise.
+   **/
+  bool isTupleAccess() const {
+    return !isDatabaseIdPlaceholder()
+        && !isRelationIdPlaceholder()
+        && !isBlockIdPlaceholder()
+        && !isTupleIdPlaceholder();
+  }
+
+  /**
+   * @brief This is a helper method for string representation
+   *        of the resource id.
+   *
+   * @return String representation of the reosurce id.
+   **/
+  std::string toString() const;
+
+ private:
+  // Negative value is invalid id for database id. Use -1 as placeholder.
+  static constexpr database_id kDatabaseIdPlaceholder = kInvalidCatalogId;
+
+  // Negative value is invalid id for relation id. Use -1 as placeholder.
+  static constexpr relation_id kRelationIdPlaceholder = kInvalidCatalogId;
+
+  // Zero is invalid id for block id. Use zero as the placeholder.
+  static constexpr block_id kBlockIdPlaceholder = kInvalidBlockId;
+
+  // Negative tuple id os invalid, therefore use a negative value.
+  static constexpr tuple_id kTupleIdPlaceholder = kInvalidCatalogId;
+
+  inline bool isDatabaseIdPlaceholder() const {
+    return db_id_ == kDatabaseIdPlaceholder;
+  }
+
+  inline bool isRelationIdPlaceholder() const {
+    return rel_id_ == kRelationIdPlaceholder;
+  }
+
+  inline bool isBlockIdPlaceholder() const {
+    return block_id_ == kBlockIdPlaceholder;
+  }
+
+  inline bool isTupleIdPlaceholder() const {
+    return tuple_id_ == kTupleIdPlaceholder;
+  }
+
+  const database_id db_id_;
+  const relation_id rel_id_;
+  const block_id block_id_;
+  const tuple_id tuple_id_;
+};
+
+/** @} */
+
+}  // namespace transaction
+}  // namespace quickstep
+
+#endif  // QUICKSTEP_RESOURCE_ID_HPP_
--- a/transaction/tests/AccessMode_unittest.cpp
+++ b/transaction/tests/AccessMode_unittest.cpp
@@ -1,0 +1,171 @@
+/**
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "transaction/AccessMode.hpp"
+
+#include "gtest/gtest.h"
+
+namespace quickstep {
+namespace transaction {
+
+class AccessModeTest : public ::testing::Test {
+ protected:
+  AccessModeTest()
+      : nl_mode_(AccessModeType::kNoLock),
+        is_mode_(AccessModeType::kIsLock),
+        ix_mode_(AccessModeType::kIxLock),
+        s_mode_(AccessModeType::kSLock),
+        six_mode_(AccessModeType::kSixLock),
+        x_mode_(AccessModeType::kXLock) {
+  }
+
+  const AccessMode nl_mode_;
+  const AccessMode is_mode_;
+  const AccessMode ix_mode_;
+  const AccessMode s_mode_;
+  const AccessMode six_mode_;
+  const AccessMode x_mode_;
+};
+
+TEST_F(AccessModeTest, ModeCompatibilty) {
+  EXPECT_TRUE(nl_mode_.isCompatible(nl_mode_));
+  EXPECT_TRUE(nl_mode_.isCompatible(is_mode_));
+  EXPECT_TRUE(nl_mode_.isCompatible(ix_mode_));
+  EXPECT_TRUE(nl_mode_.isCompatible(s_mode_));
+  EXPECT_TRUE(nl_mode_.isCompatible(six_mode_));
+  EXPECT_TRUE(nl_mode_.isCompatible(x_mode_));
+
+  EXPECT_TRUE(is_mode_.isCompatible(nl_mode_));
+  EXPECT_TRUE(is_mode_.isCompatible(is_mode_));
+  EXPECT_TRUE(is_mode_.isCompatible(ix_mode_));
+  EXPECT_TRUE(is_mode_.isCompatible(s_mode_));
+  EXPECT_TRUE(is_mode_.isCompatible(six_mode_));
+  EXPECT_FALSE(is_mode_.isCompatible(x_mode_));
+
+  EXPECT_TRUE(ix_mode_.isCompatible(nl_mode_));
+  EXPECT_TRUE(ix_mode_.isCompatible(is_mode_));
+  EXPECT_TRUE(ix_mode_.isCompatible(ix_mode_));
+  EXPECT_FALSE(ix_mode_.isCompatible(s_mode_));
+  EXPECT_FALSE(ix_mode_.isCompatible(six_mode_));
+  EXPECT_FALSE(ix_mode_.isCompatible(x_mode_));
+
+  EXPECT_TRUE(s_mode_.isCompatible(nl_mode_));
+  EXPECT_TRUE(s_mode_.isCompatible(is_mode_));
+  EXPECT_FALSE(s_mode_.isCompatible(ix_mode_));
+  EXPECT_TRUE(s_mode_.isCompatible(s_mode_));
+  EXPECT_FALSE(s_mode_.isCompatible(six_mode_));
+  EXPECT_FALSE(s_mode_.isCompatible(x_mode_));
+
+  EXPECT_TRUE(six_mode_.isCompatible(nl_mode_));
+  EXPECT_TRUE(six_mode_.isCompatible(is_mode_));
+  EXPECT_FALSE(six_mode_.isCompatible(ix_mode_));
+  EXPECT_FALSE(six_mode_.isCompatible(s_mode_));
+  EXPECT_FALSE(six_mode_.isCompatible(six_mode_));
+  EXPECT_FALSE(six_mode_.isCompatible(x_mode_));
+
+  EXPECT_TRUE(x_mode_.isCompatible(nl_mode_));
+  EXPECT_FALSE(x_mode_.isCompatible(is_mode_));
+  EXPECT_FALSE(x_mode_.isCompatible(ix_mode_));
+  EXPECT_FALSE(x_mode_.isCompatible(s_mode_));
+  EXPECT_FALSE(x_mode_.isCompatible(six_mode_));
+  EXPECT_FALSE(x_mode_.isCompatible(x_mode_));
+}
+
+TEST_F(AccessModeTest, ModeQueryChecks) {
+  EXPECT_FALSE(nl_mode_.isIntentionShareLock());
+  EXPECT_FALSE(nl_mode_.isIntentionExclusiveLock());
+  EXPECT_FALSE(nl_mode_.isShareLock());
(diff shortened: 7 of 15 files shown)


transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+ **/
+
+#ifndef QUICKSTEP_TRANSACTION_ACCESS_MODE_HPP_
+#define QUICKSTEP_TRANSACTION_ACCESS_MODE_HPP_
+
+#include <cstdint>
+#include <type_traits>
+
+namespace quickstep {
+
```
zuyu: We can remove this empty line.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * @brief Represents mode type. Possible options are NL, IS, IX, S, SIX, X.
+ **/
+enum class AccessModeType : std::uint8_t {
+  kNoLock = 0,
+  kIsLock = 1,
+  kIxLock = 2,
+  kSLock = 3,
+  kSixLock = 4,
+  kXLock = 5,
```
zuyu: Remove the value assignments from `1` to `5`. It is automatically set.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+ * @brief Class for representing resource lock's access mode
+ **/
+class AccessMode {
+ public:
+  /**
+   * @brief Only constructor for access mode.
+   *
+   * @param access_mode Mode type of the object.
+   **/
+  explicit AccessMode(AccessModeType access_mode)
```
zuyu: Mark `access_mode` `const`.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+    : access_mode_(access_mode) {}
```
zuyu: Add two more whitespace indents.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * @brief Checks whether this access mode is compatible with the other.
+   *
+   * @param other Other access mode that will be checked against to this one.
+   * @return True if they are compatible, otherwise false.
+   **/
+  inline bool isCompatible(const AccessMode &other) const {
+    typedef std::underlying_type<AccessModeType>::type type_to_cast;
+    type_to_cast this_mode = static_cast<type_to_cast>(this->access_mode_);
```
zuyu: Remove `this->`.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+  // True means they are compatible.
+  static constexpr bool kCompMatrix[kNumberLocks][kNumberLocks] = {
+/*           NL     IS     IX      S     SIX     X    */
+/*  NL  */ {true , true , true , true , true , true },
+/*  IS  */ {true , true , true , true , true , false},
+/*  IX  */ {true , true , true , false, false, false},
+/*  S   */ {true , true , false, true , false, false},
+/*  SIX */ {true , true , false, false, false, false},
+/*  X   */ {true , false, false, false, false, false}
+  };
```
zuyu: Move the definition to the source file.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * @brief Checks whether this access mode is compatible with the other.
+   *
+   * @param other Other access mode that will be checked against to this one.
+   * @return True if they are compatible, otherwise false.
+   **/
+  inline bool isCompatible(const AccessMode &other) const {
+    typedef std::underlying_type<AccessModeType>::type type_to_cast;
+    type_to_cast this_mode = static_cast<type_to_cast>(this->access_mode_);
+    type_to_cast other_mode = static_cast<type_to_cast>(other.access_mode_);
```
zuyu: I was wondering why you could directly access `other`'s private member `access_mode_`.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * @return True if both modes have the same level.
+   **/
+  bool operator==(const AccessMode &other) const {
+    return access_mode_ == other.access_mode_;
+  }
+
+ private:
+  // The compatibility matrix should be N by N. kNumberLocks == N.
+  static constexpr std::uint64_t kNumberLocks = 6;
```
zuyu: Remove this line (which includes a magic number and introduces potential inconsistency if someone adds a new lock but forgets to increase this value), and use the enum class to determine the number of locks.

transaction/CMakeLists.txt
```diff
@@ -13,9 +13,18 @@
...
 
+add_library(quickstep_transaction_AccessMode
+            ../empty_src.cpp
+            AccessMode.hpp)
 add_library(quickstep_transaction_DirectedGraph
             ../empty_src.cpp
             DirectedGraph.hpp)
+add_library(quickstep_transaction_Lock
+	    ../empty_src.cpp
+	    Lock.hpp)
```
zuyu: Add one more whitespace indent for above two lines.

transaction/CMakeLists.txt
```diff
@@ -13,9 +13,18 @@
...
+add_library(quickstep_transaction_ResourceId
+	    ResourceId.cpp
+	    ResourceId.hpp)
```
zuyu: Add one more whitespace indent for above two lines.

transaction/CMakeLists.txt
```diff
@@ -27,6 +36,14 @@ target_link_libraries(quickstep_transaction_DirectedGraph
 	              glog
                       quickstep_transaction_Transaction
                       quickstep_utility_Macros)
+target_link_libraries(quickstep_transaction_Lock
+		      quickstep_transaction_AccessMode
+		      quickstep_transaction_ResourceId)
```
zuyu: Add one more whitespace indent for above two lines.

transaction/CMakeLists.txt
```diff
@@ -27,6 +36,14 @@ target_link_libraries(quickstep_transaction_DirectedGraph
...
+target_link_libraries(quickstep_transaction_ResourceId
+		      quickstep_catalog_CatalogTypedefs
+		      quickstep_storage_StorageBlockInfo
+		      quickstep_utility_HashPair
+		      quickstep_utility_Macros)
```
zuyu: Add one more whitespace indent for above four lines.

transaction/CMakeLists.txt
```diff
@@ -35,10 +52,21 @@ add_library(quickstep_transaction
             ../empty_src.cpp
             TransactionModule.hpp)
 target_link_libraries(quickstep_transaction
+	              quickstep_transaction_AccessMode
                       quickstep_transaction_DirectedGraph
+	              quickstep_transaction_Lock
+	              quickstep_transaction_ResourceId
```
zuyu: Add one more whitespace indent for above three lines.

transaction/Lock.hpp
```diff
@@ -0,0 +1,92 @@
...
+ **/
+
+#ifndef QUICKSTEP_TRANSACTION_LOCK_HPP_
+#define QUICKSTEP_TRANSACTION_LOCK_HPP_
+
+#include "transaction/AccessMode.hpp"
+#include "transaction/ResourceId.hpp"
+
+namespace quickstep {
+
```
zuyu: Remove this empty line.

transaction/Lock.hpp
```diff
@@ -0,0 +1,92 @@
...
+ **/
+class Lock {
+ public:
+  /**
+   * @brief Constructor.
+   *
+   * @param rid Resource id of the resource on which the lock is taken.
+   * @param access_mode Mode of the lock access.
+   **/
+  Lock(ResourceId rid, AccessMode access_mode)
```
zuyu: Mark two arguments `const`.

transaction/Lock.hpp
```diff
@@ -0,0 +1,92 @@
...
+  inline bool operator==(const Lock &other) const {
+    return rid_ == other.rid_ && access_mode_ = other.access_mode_;
+  }
+
+  /**
+   * @brief Getter for resource id.
+   *
+   * @return Resource id of the lock.
+   **/
+  inline ResourceId getResourceId() const {
```
zuyu: Return the const reference instead of a copy?

transaction/Lock.hpp
```diff
@@ -0,0 +1,92 @@
...
+    return rid_;
+  }
+
+  /**
+   * @brief Getter for access mode.
+   *
+   * @return Access mode of the lock.
+   **/
+  inline AccessMode getAccessMode() const {
```
zuyu: Return the const reference instead of a copy?

transaction/Lock.hpp
```diff
@@ -0,0 +1,92 @@
...
+    return access_mode_;
+  }
+
+ private:
+  // Id of the resource that is locked.
+  ResourceId rid_;
```
zuyu: Mark it `const`?

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+ **/
+
+#include "transaction/ResourceId.hpp"
+
+#include <cstddef>
+#include <string>
+#include <utility>
+
+namespace quickstep {
+
```
zuyu: Remove this empty line.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+namespace transaction {
+
+bool ResourceId::operator==(const ResourceId &other) const {
+  return (db_id_ == other.db_id_) &&
+    (rel_id_ == other.rel_id_) &&
+    (block_id_ == other.block_id_) &&
+    (tuple_id_ == other.tuple_id_);
```
zuyu: I think it looks better:

```
  return db_id_ == other.db_id_
      && rel_id_ == other.rel_id_
      && block_id_ == other.block_id_
      && tuple_id_ == other.tuple_id_;
```

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+ResourceId ResourceId::getParentResourceId() const {
+  if (isRelationAccess()) {
+    return ResourceId(db_id_);
+  } else if (isBlockAccess()) {
+    return ResourceId(db_id_, rel_id_);
+  } else if (isTupleAccess()) {
+    return ResourceId(db_id_, rel_id_, block_id_);
+  } else {
+    FATAL_ERROR("Trying to call getParentResourceId on database");
```
zuyu: Use `LOG(FATAL)` in `glog`.

zuyu:
@hakanmemisoglu You need to run `iwyu` for all files you touched.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+bool ResourceId::isDatabaseAccess() const {
+  return (db_id_ != kDbIdPlaceholder) &&
+    (rel_id_ == kRelIdPlaceholder) &&
+    (block_id_ == kBlockIdPlaceholder) &&
+    (tuple_id_ == kTupleIdPlaceholder);
```
zuyu: Ditto for refactoring `ResourceId::operator==`.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+bool ResourceId::isRelationAccess() const {
+  return (db_id_ != kDbIdPlaceholder) &&
+    (rel_id_ != kRelIdPlaceholder) &&
+    (block_id_ == kBlockIdPlaceholder) &&
+    (tuple_id_ == kTupleIdPlaceholder);
```
zuyu: Ditto for refactoring `ResourceId::operator==`.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+bool ResourceId::isBlockAccess() const {
+  return (db_id_ != kDbIdPlaceholder) &&
+    (rel_id_ != kRelIdPlaceholder) &&
+    (block_id_ != kBlockIdPlaceholder) &&
+    (tuple_id_ == kTupleIdPlaceholder);
```
zuyu: Ditto for refactoring `ResourceId::operator==`.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+bool ResourceId::isTupleAccess() const {
+  return (db_id_ != kDbIdPlaceholder) &&
+    (rel_id_ != kRelIdPlaceholder) &&
+    (block_id_ != kBlockIdPlaceholder) &&
+    (tuple_id_ != kTupleIdPlaceholder);
```
zuyu: Ditto for refactoring `ResourceId::operator==`.

transaction/ResourceId.cpp
zuyu: Actually, I suggest to provide four `private` methods to check whether the given four resource id is the placeholder, and then use the combinations of the four methods for this method and three more below.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+std::string ResourceId::toString() const {
+  return "ResourceId(" + std::to_string(db_id_) + ", " +
+    std::to_string(rel_id_) + ", " + std::to_string(block_id_) +
+    ", " + std::to_string(tuple_id_) + ")";
```
zuyu: Suggest to

```
  return "ResourceId(" +
      std::to_string(db_id_) + ", " +
      std::to_string(rel_id_) + ", " +
      std::to_string(block_id_) + ", " +
      std::to_string(tuple_id_) + ")";
```

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+std::size_t
+ResourceId::ResourceIdHasher::operator()(const ResourceId &rid) const {
+  std::size_t hash1 = std::hash<database_id>()(rid.db_id_);
+  std::size_t hash2 = std::hash<relation_id>()(rid.rel_id_);
+  std::size_t hash3 = std::hash<block_id>()(rid.block_id_);
+  std::size_t hash4 = std::hash<tuple_id>()(rid.tuple_id_);
+
+  std::size_t comb1 = CombineHashes(hash1, hash2);
+  std::size_t comb2 = CombineHashes(hash3, hash4);
```
zuyu: Mark all six variables `const`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+#include <string>
+#include <type_traits>
+
+#include "catalog/CatalogTypedefs.hpp"
+#include "storage/StorageBlockInfo.hpp"
+#include "utility/HashPair.hpp"
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+
```
zuyu: Remove this empty line.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+class ResourceId {
+ public:
+  static constexpr database_id kDbIdPlaceholder =
+    std::numeric_limits<database_id>::max();
+  static constexpr relation_id kRelIdPlaceholder =
+    std::numeric_limits<relation_id>::max();
+  static constexpr block_id kBlockIdPlaceholder =
+    std::numeric_limits<block_id>::max();
+  static constexpr tuple_id kTupleIdPlaceholder =
+    std::numeric_limits<tuple_id>::max();
```
zuyu: Define all four placeholders in the source file.

Question: what's the meaning of these placeholders (i.e., invalid ids?), and why they are all the max value?

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+                 kBlockIdPlaceholder,
+                 kTupleIdPlaceholder) {
+  }
+
+  /**
+   * @brief Constructor for database level resource id.
+   *
+   * @param db_id unique id of database resource.
+   **/
+  explicit ResourceId(database_id db_id)
```
zuyu: Mark `db_id` `const`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * @param db_id Unique id of the database resource.
+   * @param rel_id Unique id of the relation resource in the database.
+   * @param block_id Unique id of the block resource in the relation.
+   * @param tuple_id Unique id of the tuple resource in the block.
+   **/
+  ResourceId(database_id db_id,
+             relation_id rel_id,
+             block_id block_id,
+             tuple_id tuple_id)
```
zuyu: Mark all four arguments `const`.

Why not give these arguments the default placeholder values?

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /**
+   * @brief Constructor for block level resource id.
+   *
+   * @param db_id unique id of the databse resource.
+   * @param rel_id unique id of the relation resource in the database.
+   * @param block_id unique id of the block resource in the relation.
+   **/
+  ResourceId(database_id db_id, relation_id rel_id, block_id block_id)
```
zuyu: Mark all three arguments `const`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+                 kTupleIdPlaceholder) {
+  }
+
+  /**
+   * @brief Constructor for relation level resource id.
+   *
+   * @param db_id unique id of database resource.
+   * @param rel_id unique id of relation resource in database.
+   **/
+  ResourceId(database_id db_id, relation_id rel_id)
```
zuyu: Mark two arguments `const`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+  /**
+   * @brief Copy constructor.
+   *
+   * @param other The ResourceId that will be copied.
+   **/
+  ResourceId(const ResourceId &other) {
+    this->db_id_ = other.db_id_;
+    this->rel_id_ = other.rel_id_;
+    this->block_id_ = other.block_id_;
+    this->tuple_id_ = other.tuple_id_;
```
zuyu: Remove all `this->`, and suggest to use the init list:

```
  ResourceId(const ResourceId &other)
      : db_id_(other.db_id_),
        ...,
        tuple_id_(other.tuple_id_) {}
```

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * @return String representation of the reosurce id.
+   **/
+  std::string toString() const;
+
+ private:
+  database_id db_id_;
+  relation_id rel_id_;
+  block_id block_id_;
+  tuple_id tuple_id_;
```
zuyu: Mark all four members `const`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,217 @@
...
+};
+
+/** @} */
+
+}  // namespace transaction
+
```
zuyu: Remove this empty line.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+std::size_t
+ResourceId::ResourceIdHasher::operator()(const ResourceId &rid) const {
+  std::size_t hash1 = std::hash<database_id>()(rid.db_id_);
+  std::size_t hash2 = std::hash<relation_id>()(rid.rel_id_);
+  std::size_t hash3 = std::hash<block_id>()(rid.block_id_);
+  std::size_t hash4 = std::hash<tuple_id>()(rid.tuple_id_);
+
+  std::size_t comb1 = CombineHashes(hash1, hash2);
```
zuyu: Include `utility/HashPair.hpp`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,114 @@
...
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "transaction/AccessMode.hpp"
+
+#include "gtest/gtest.h"
+
+namespace quickstep {
+
```
zuyu: Remove this empty line.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace transaction {
+
+TEST(AccessModeTest, ModeCompatibilty) {
+  AccessMode is_mode = AccessMode(AccessModeType::kIsLock);
+  AccessMode ix_mode = AccessMode(AccessModeType::kIxLock);
+  AccessMode s_mode = AccessMode(AccessModeType::kSLock);
+  AccessMode x_mode = AccessMode(AccessModeType::kXLock);
```
zuyu: Mark all `const`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,114 @@
...
+  EXPECT_FALSE(x_mode.isCompatible(ix_mode));
+  EXPECT_FALSE(x_mode.isCompatible(s_mode));
+  EXPECT_FALSE(x_mode.isCompatible(x_mode));
+}
+
+TEST(AccessModeTest, ModeQueryChecks) {
+  AccessMode is_mode = AccessMode(AccessModeType::kIsLock);
+  AccessMode ix_mode = AccessMode(AccessModeType::kIxLock);
+  AccessMode s_mode = AccessMode(AccessModeType::kSLock);
+  AccessMode x_mode = AccessMode(AccessModeType::kXLock);
```
zuyu: Mark all `const`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,114 @@
...
+TEST(AccessModeTest, Equality) {
+  AccessMode is_mode_1 = AccessMode(AccessModeType::kIsLock);
+  AccessMode ix_mode_1 = AccessMode(AccessModeType::kIxLock);
+  AccessMode s_mode_1 = AccessMode(AccessModeType::kSLock);
+  AccessMode x_mode_1 = AccessMode(AccessModeType::kXLock);
+
+  AccessMode is_mode_2 = AccessMode(AccessModeType::kIsLock);
+  AccessMode ix_mode_2 = AccessMode(AccessModeType::kIxLock);
+  AccessMode s_mode_2 = AccessMode(AccessModeType::kSLock);
+  AccessMode x_mode_2 = AccessMode(AccessModeType::kXLock);
```
zuyu: Mark all `const`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  EXPECT_FALSE(s_mode_1 == is_mode_1);
+  EXPECT_FALSE(s_mode_1 == ix_mode_1);
+  EXPECT_TRUE(s_mode_1 == s_mode_2);
+  EXPECT_FALSE(s_mode_1 == x_mode_1);
+
+  EXPECT_FALSE(x_mode_1 == is_mode_1);
+  EXPECT_FALSE(x_mode_1 == ix_mode_1);
+  EXPECT_FALSE(x_mode_1 == s_mode_1);
+  EXPECT_TRUE(x_mode_1 == x_mode_2);
```
zuyu: Try to replace `EXPECT_TRUE` with `EXPECT_EQ` (the left-hand side is the expected value, and the right-hand is the checked), `EXPECT_FALSE` with `EXPECT_NE`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+}  // namespace transaction
+
```
zuyu: Remove this empty line.

transaction/tests/ResourceId_unittest.cpp
```diff
@@ -0,0 +1,73 @@
...
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "transaction/ResourceId.hpp"
+
+#include "gtest/gtest.h"
+
+namespace quickstep {
+
```
zuyu: Remove this empty line.

transaction/tests/ResourceId_unittest.cpp
```diff
@@ -0,0 +1,73 @@
...
+namespace transaction {
+
+TEST(ResourceIdTest, CheckAccessLevels) {
+  ResourceId db2(2);
+  ResourceId rel4_db2(2, 4);
+  ResourceId block5_rel4_db2(2, 4, 5);
+  ResourceId tpl8_block5_rel4_db2(2, 4, 5, 8);
```
zuyu: Mark all `const`.

transaction/tests/ResourceId_unittest.cpp
```diff
@@ -0,0 +1,73 @@
...
+  EXPECT_FALSE(tpl8_block5_rel4_db2.isBlockAccess());
+
+  EXPECT_FALSE(db2.isTupleAccess());
+  EXPECT_FALSE(rel4_db2.isTupleAccess());
+  EXPECT_FALSE(block5_rel4_db2.isTupleAccess());
+  EXPECT_TRUE(tpl8_block5_rel4_db2.isTupleAccess());
+}
+
+TEST(ResourceIdTest, Equality) {
+}
```
zuyu: ?

transaction/tests/ResourceId_unittest.cpp
```diff
@@ -0,0 +1,73 @@
...
+  ResourceId rel2_db2(2, 2);
+  ResourceId rel4_db2(2, 4);
+  ResourceId rel5_db3(3, 5);
+  ResourceId rel9_db3(3, 9);
+  ResourceId block2_rel2_db2(2, 2, 2);
+  ResourceId block3_rel2_db2(2, 2, 3);
+  ResourceId block7_rel5_db3(3, 5, 7);
+  ResourceId block9_rel5_db3(3, 5, 9);
+  ResourceId tuple4_block2_rel2_db2(2, 2, 2, 4);
+  ResourceId tuple7_block2_rel2_db2(2, 2, 2, 7);
```
zuyu: Mark all `const`, and finish the unit tests.

transaction/AccessMode.hpp
hakanmemisoglu: Ok, I will change it, thanks.

transaction/AccessMode.hpp
hakanmemisoglu: Noted.

transaction/AccessMode.hpp
hakanmemisoglu: I will add const qualifier.

transaction/AccessMode.hpp
hakanmemisoglu: Thanks, I will change it.

transaction/AccessMode.hpp
hakanmemisoglu: I will remove it.

transaction/AccessMode.hpp
hakanmemisoglu: There is no getter method. If it is a problem, I can introduce it.

transaction/AccessMode.hpp
hakanmemisoglu: Do you know any proper method for that. What do you mean by "use the enum class to determine the number of locks." Is it something like:

`enum class SomeEnum {
    variant1 = 0,
    variant2,
    variant3,
    SIZE,
};`

transaction/CMakeLists.txt
hakanmemisoglu: Hi Zuyu,
Can you check them again? It seems they are properly indented.

transaction/CMakeLists.txt
hakanmemisoglu: Same comment as above.

transaction/CMakeLists.txt
hakanmemisoglu: Same comment above.

transaction/CMakeLists.txt
hakanmemisoglu: Same comment above.

transaction/CMakeLists.txt
hakanmemisoglu: Same comment above.

transaction/Lock.hpp
hakanmemisoglu: I will remove it.

transaction/Lock.hpp
hakanmemisoglu: I will change it.

transaction/Lock.hpp
hakanmemisoglu: I will change it.

transaction/Lock.hpp
hakanmemisoglu: I will change it.

transaction/Lock.hpp
hakanmemisoglu: I will add const qualifier.

transaction/ResourceId.cpp
hakanmemisoglu: I will remove it.

transaction/ResourceId.cpp
hakanmemisoglu: I agree that it is visually better. I will change it, thanks for the suggestion.

transaction/ResourceId.cpp
hakanmemisoglu: I will change it, thanks.

transaction/ResourceId.cpp
hakanmemisoglu: It makes sense, thank I will change it.

transaction/ResourceId.cpp
hakanmemisoglu: I will change them.

transaction/ResourceId.hpp
hakanmemisoglu: Noted.

transaction/ResourceId.hpp
hakanmemisoglu: I will do it.

transaction/ResourceId.hpp
hakanmemisoglu: I will change it.

transaction/ResourceId.hpp
hakanmemisoglu: Noted.

transaction/ResourceId.hpp
hakanmemisoglu: Noted.

transaction/ResourceId.cpp
hakanmemisoglu: I will add it.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Thanks for the suggestion.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/ResourceId_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/ResourceId_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/ResourceId_unittest.cpp
hakanmemisoglu: I will fix this.

transaction/tests/ResourceId_unittest.cpp
hakanmemisoglu: I will fix it, thanks.

transaction/ResourceId.hpp
hakanmemisoglu: Noted.

transaction/ResourceId.hpp
hakanmemisoglu: Noted.

transaction/ResourceId.hpp
hakanmemisoglu: RIDs are represented as tuple of 4 ids. Placeholders values mean they do not represent anything. For example when a RID has the value of (4, 5, PH, PH), it means database 4's block 5.

transaction/ResourceId.hpp
hakanmemisoglu: It makes sense. I will switch to your suggestion.

transaction/AccessMode.hpp
zuyu: Yes.

`kNum AccessModeTypes  // Not a real AccessModeType. Used for counting.`

transaction/CMakeLists.txt
zuyu: Ah, you're ok. I think it's just the display issue under `Chrome`.

transaction/ResourceId.hpp
zuyu: Then I suggest to use the invalid values for each category instead, because some placeholder values here are valid for some cases, including `database_id`, `relation_id`, and `block_id`.

hakanmemisoglu: I know these are valid values, but the max value is what we decided at the meeting with you and @pateljm  as a placeholder.

zuyu: No, at that time I only suggested to use invalid values.

transaction/AccessMode.hpp
hakanmemisoglu: AccessMode is header only. Should I create a cpp file for just one static field? Moreover, kCompMatrix has a role of documentation.

zuyu: Yes, you should create the cpp file.

Btw, rename `kCompMatrix` to `kLockCompatibilityMatrix`.

transaction/ResourceId.hpp
hakanmemisoglu: You suggested and later we decided to use max values. The reason was introducing invalid ids as negative values removes the half of the domain. I can change it, no problem with that if you want us to keep using negative domain as invalid ids and avoid changing original code.

zuyu: The main thing here is about consistency so that the values we choose here should be invalid. If we choose to use the max values, we need to change the typedefs for `database_id` and `relation_id` etc, and mark the max values as invalid.

transaction/AccessMode.hpp
hakanmemisoglu: I will change the name. By the way, C++ standard says, a constexpr initialized immediately. I cannot move definition to a cpp file.

zuyu: Change `constexpr` to `const` then.

zuyu: Alright, move it before the constructor, according to [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Declaration_Order).

transaction/ResourceId.cpp
hakanmemisoglu: I will implement the private methods.

transaction/ResourceId.cpp
hakanmemisoglu: Noted.

transaction/ResourceId.cpp
hakanmemisoglu: Noted.

transaction/ResourceId.cpp
hakanmemisoglu: Noted.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * @brief Checks whether this access mode is compatible with the other.
+   *
+   * @param other Other access mode that will be checked against to this one.
+   * @return True if they are compatible, otherwise false.
+   **/
+  inline bool isCompatible(const AccessMode &other) const {
+    typedef std::underlying_type<AccessModeType>::type type_to_cast;
+    type_to_cast this_mode = static_cast<type_to_cast>(access_mode_);
+    type_to_cast other_mode = static_cast<type_to_cast>(other.access_mode_);
```
zuyu: Mark the above two `const`; or, use them directly as the index to the lock compatibility matrix.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,167 @@
...
+      : access_mode_(access_mode) {}
+
+  /**
+   * @brief Checks whether this access mode is compatible with the other.
+   *
+   * @param other Other access mode that will be checked against to this one.
+   * @return True if they are compatible, otherwise false.
+   **/
+  inline bool isCompatible(const AccessMode &other) const {
+    typedef std::underlying_type<AccessModeType>::type type_to_cast;
```
zuyu: You could move this `typedef` after `public` before the constructor.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,167 @@
...
+   **/
+  inline bool operator!=(const AccessMode &other) const {
+    return access_mode_ != other.access_mode_;
+  }
+
+ private:
+  // The compatibility matrix should be N by N. kNumberLocks == N.
+  static constexpr std::uint64_t kNumberLocks =
+    static_cast<std::underlying_type<AccessModeType>::type>(
+      AccessModeType::kNumAccessModeTypes);
```
zuyu: Add two more whitespace indents.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,167 @@
...
+   * @return True if the modes have different levels.
+   **/
+  inline bool operator!=(const AccessMode &other) const {
+    return access_mode_ != other.access_mode_;
+  }
+
+ private:
+  // The compatibility matrix should be N by N. kNumberLocks == N.
+  static constexpr std::uint64_t kNumberLocks =
+    static_cast<std::underlying_type<AccessModeType>::type>(
```
zuyu: Add two more whitespace indents, and use the above mentioned `typedef` to simplify this expression.

FYI, move all static members after `public` and before the constructor.

transaction/Lock.hpp
```diff
@@ -0,0 +1,91 @@
...
+class Lock {
+ public:
+  /**
+   * @brief Constructor.
+   *
+   * @param rid Resource id of the resource on which the lock is taken.
+   * @param access_mode Mode of the lock access.
+   **/
+  Lock(const ResourceId &rid, const AccessMode &access_mode)
+    : rid_(rid),
```
zuyu: Add two more whitespace indents.

transaction/Lock.hpp
```diff
@@ -0,0 +1,91 @@
...
+      access_mode_(access_mode) {
```
zuyu: Add two more whitespace indents for above two lines.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,111 @@
...
+ResourceId ResourceId::getParentResourceId() const {
+  if (isRelationAccess()) {
+    return ResourceId(db_id_);
+  } else if (isBlockAccess()) {
+    return ResourceId(db_id_, rel_id_);
+  } else if (isTupleAccess()) {
+    return ResourceId(db_id_, rel_id_, block_id_);
+  } else {
+    LOG(FATAL) << "Database level does not have any parent level.";
+    return ResourceId();
```
zuyu: Remove this line. No need for this, as `glog FATAL` terminates the program.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,195 @@
...
+      : db_id_(other.db_id_),
+        rel_id_(other.rel_id_),
+        block_id_(other.block_id_),
+        tuple_id_(other.tuple_id_) {
+  }
+
+  /**
+   * @brief Hasher class for ResourceId class to use it in the hash map.
+   **/
+  struct ResourceIdHasher {
```
zuyu: We may want to move it under `private:`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,195 @@
...
+   * @brief This is a helper method for string representation
+   *        of the resource id.
+   *
+   * @return String representation of the reosurce id.
+   **/
+  std::string toString() const;
+
+ private:
+  // Placeholder values of for non-existent ids.
+  static constexpr database_id kDbIdPlaceholder =
```
zuyu: I prefer `kDatabaseIdPlaceholder`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,195 @@
...
+    std::numeric_limits<database_id>::max();
+
+  static constexpr relation_id kRelIdPlaceholder =
```
zuyu: I prefer `kRelationIdPlaceholder`.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,195 @@
...
+    std::numeric_limits<relation_id>::max();
+
+  static constexpr block_id kBlockIdPlaceholder =
+    std::numeric_limits<block_id>::max();
+
+  static constexpr tuple_id kTupleIdPlaceholder =
+    std::numeric_limits<tuple_id>::max();
```
zuyu: Per the discussion, the values for placeholders should be invalid ones. Also, move all static members between `public:` and the constructor.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "gtest/gtest.h"
+
+namespace quickstep {
+namespace transaction {
+
+TEST(AccessModeTest, ModeCompatibilty) {
+  const AccessMode is_mode = AccessMode(AccessModeType::kIsLock);
+  const AccessMode ix_mode = AccessMode(AccessModeType::kIxLock);
+  const AccessMode s_mode = AccessMode(AccessModeType::kSLock);
+  const AccessMode x_mode = AccessMode(AccessModeType::kXLock);
```
zuyu: Why missed two more lock types?

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,112 @@
...
+  EXPECT_FALSE(x_mode.isCompatible(ix_mode));
+  EXPECT_FALSE(x_mode.isCompatible(s_mode));
+  EXPECT_FALSE(x_mode.isCompatible(x_mode));
+}
+
+TEST(AccessModeTest, ModeQueryChecks) {
+  const AccessMode is_mode = AccessMode(AccessModeType::kIsLock);
+  const AccessMode ix_mode = AccessMode(AccessModeType::kIxLock);
+  const AccessMode s_mode = AccessMode(AccessModeType::kSLock);
+  const AccessMode x_mode = AccessMode(AccessModeType::kXLock);
```
zuyu: Add two more lock modes.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,112 @@
...
+  EXPECT_FALSE(x_mode.isIntentionExclusiveLock());
+  EXPECT_FALSE(x_mode.isShareLock());
+  EXPECT_TRUE(x_mode.isExclusiveLock());
+}
+
+TEST(AccessModeTest, Equality) {
+  const AccessMode is_mode_1 = AccessMode(AccessModeType::kIsLock);
+  const AccessMode ix_mode_1 = AccessMode(AccessModeType::kIxLock);
+  const AccessMode s_mode_1 = AccessMode(AccessModeType::kSLock);
+  const AccessMode x_mode_1 = AccessMode(AccessModeType::kXLock);
```
zuyu: Rename `1` to `expected`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  const AccessMode is_mode_2 = AccessMode(AccessModeType::kIsLock);
+  const AccessMode ix_mode_2 = AccessMode(AccessModeType::kIxLock);
+  const AccessMode s_mode_2 = AccessMode(AccessModeType::kSLock);
+  const AccessMode x_mode_2 = AccessMode(AccessModeType::kXLock);
```
zuyu: Rename `2` to `checked`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  EXPECT_NE(s_mode_1, is_mode_1);
+  EXPECT_NE(s_mode_1, ix_mode_1);
+  EXPECT_EQ(s_mode_1, s_mode_2);
+  EXPECT_NE(s_mode_1, x_mode_1);
+
+  EXPECT_NE(x_mode_1, is_mode_1);
+  EXPECT_NE(x_mode_1, ix_mode_1);
+  EXPECT_NE(x_mode_1, s_mode_1);
+  EXPECT_EQ(x_mode_1, x_mode_2);
```
zuyu: Make sure that the left side is the expected, and the right one is the checked.

transaction/ResourceId.hpp
hakanmemisoglu: These constants won't be used anywhere else. Why should we do them public?

transaction/ResourceId.hpp
hakanmemisoglu: I will change it.

transaction/tests/ResourceId_unittest.cpp
```diff
@@ -0,0 +1,104 @@
...
+  EXPECT_EQ(db2_rel4_block5_tpl8_copy, db2_rel4_block5_tpl8);
+}
+
+TEST(ResourceIdTest, ParentResourceIds) {
+  const ResourceId db2(2);
+  const ResourceId db2_rel2(2, 2);
+  const ResourceId db2_rel2_block2(2, 2, 2);
+  const ResourceId db2_rel2_block3(2, 2, 3);
+  const ResourceId db2_rel2_block2_tuple4(2, 2, 2, 4);
+  const ResourceId db2_rel2_block2_tuple7(2, 2, 2, 7);
```
zuyu: I suggest to use a test class derived by `::testing::Test` to hold all the `ResourceId`s, and then test each functionalities using one fixture.

In addition, you need to test every case for the `ResourceId`s, including valid values, invalid, and partially valid ones.

hakanmemisoglu:
@zuyu Thanks for the feedback. I have installation problems with iwyu. Therefore I could not run it. Let's wait until I can install it.

transaction/ResourceId.cpp
hakanmemisoglu: I will remove it.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,111 @@
...
+      && !isBlockIdPlaceholder()
+      && isTupleIdPlaceholder();
+}
+
+bool ResourceId::isTupleAccess() const {
+  return !isDatabaseIdPlaceholder()
+      && !isRelationIdPlaceholder()
+      && !isBlockIdPlaceholder()
+      && !isTupleIdPlaceholder();
+}
```
hakanmemisoglu: Noted.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,111 @@
...
+      && tuple_id_ == other.tuple_id_;
+}
+
+bool ResourceId::operator!=(const ResourceId &other) const {
+  return !(*this == other);
+}
+
+bool ResourceId::hasParent() const {
+  return !isDatabaseAccess();
+}
```
hakanmemisoglu: Noted.

transaction/Lock.hpp
hakanmemisoglu: Noted. Thanks for the warning.

transaction/AccessMode.hpp
hakanmemisoglu: Noted.

transaction/AccessMode.hpp
hakanmemisoglu: It is not a public typedef. This function is the only user of this type alias.

transaction/AccessMode.hpp
hakanmemisoglu: Noted.

transaction/AccessMode.hpp
hakanmemisoglu: Actually let's make this typedef private since we introduce another usage as you said in the kNumberLocks

transaction/AccessMode.hpp
hakanmemisoglu: Noted. However, I still emphasize the typedefs should be private.

transaction/ResourceId.cpp
```diff
@@ -0,0 +1,111 @@
...
+bool ResourceId::operator==(const ResourceId &other) const {
+  return db_id_ == other.db_id_
+      && rel_id_ == other.rel_id_
+      && block_id_ == other.block_id_
+      && tuple_id_ == other.tuple_id_;
+}
+
+bool ResourceId::operator!=(const ResourceId &other) const {
+  return !(*this == other);
+}
```
hakanmemisoglu: I will move the inequality operator.

transaction/ResourceId.hpp
hakanmemisoglu: I intended to make it public. Do you see any possible problems?

transaction/ResourceId.hpp
hakanmemisoglu: I will change it.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: I removed them, because I did not use them in the lock manager. However your are right, since we added them in public. I will add them, thanks for the warning.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Ok.

transaction/tests/ResourceId_unittest.cpp
hakanmemisoglu: Thanks for the suggestion, I will change it.

zuyu:
@hakanmemisoglu What issues did you encounter during installing `iwyu`? On `Mac`, I use `brew`.

transaction/AccessMode.hpp
zuyu: Make sense.

transaction/AccessMode.hpp
zuyu: No problem.

transaction/tests/AccessMode_unittest.cpp
zuyu: No problem. The meaning of this test is to ensure no incorrect value assignments in the compatibility matrix.

transaction/ResourceId.hpp
zuyu: OK, let's keep them `private`.

But the placeholders' values are the design issues.

hakanmemisoglu:
I am using the computer at the office.

transaction/ResourceId.hpp
hakanmemisoglu: Ok I will change them. Let me look at the source code to find the invalid values.

hakanmemisoglu:
I pushed the code, but I have not changed the placeholders, and applied iwyu. Let's review it after fixing them.

transaction/ResourceId.hpp
```diff
@@ -0,0 +1,216 @@
...
+  static constexpr database_id kDatabaseIdPlaceholder = -1;
+
+  // Negative value is invalid id for relation id. Use -1 as placeholder.
+  static constexpr relation_id kRelationIdPlaceholder = -1;
+
+  // Zero is invalid id for block id. Use zero as the placeholder.
+  static constexpr block_id kBlockIdPlaceholder = 0;
+
+  // Negative tuple id os invalid, therefore use a negative value.
+  static constexpr tuple_id kTupleIdPlaceholder = -1;
```
zuyu: Instead of declaring these placeholder values and using magic numbers, just use the invalid values directly, for example, `kInvalidBlockId`.

Also, just add `static constexpr int kInvalidCatalogId = -1;` in `/catalog/CatalogTypedefs.hpp`, and then use it for all catalog ids.

Finally, you could rename the following methods to sth like `isInvalidDatabaseId`.

transaction/ResourceId.hpp
hakanmemisoglu: Container `unordered_map` should see this one, that's why I did it. I can try `friend`.

transaction/ResourceId.hpp
hakanmemisoglu: Thanks for the suggestion, I will change them.

hakanmemisoglu: I changed the usage of magic numbers. However, I guess we should not change the name of the functions since the logic between a placeholder and an invalid is is different despite they are same values.

transaction/ResourceId.hpp
hakanmemisoglu: Hi @zuyu,
Actually, friending is not a good solution for this one. If that would be case, I should add each class which uses an unordered STL container with ResourceId, as a friend. 
A possible alternatives: 
1- C++ documentation recommends to specialize std::hash in the user code. However, Google C++ Guide forbids it. 
Do you have any other idea?

transaction/AccessMode.hpp
zuyu: @hakanmemisoglu Are you going to create the `typedef`?

hakanmemisoglu: I will, sorry for the delay.

transaction/AccessMode.hpp
hakanmemisoglu: GC++SG says it should be moved to before constructor within **section**. In this case, it is at the beginning of the private section, where it is the place the style guide dictates.

zuyu: Never mind for this one.

catalog/CatalogTypedefs.hpp
```diff
@@ -36,6 +36,10 @@ typedef int attribute_id;
 typedef std::size_t partition_id;
 typedef int numa_node_id;
 
+// Catalog ids use negative values as invalid ids. Mark -1 as constant invalid
+// id for the catalog ids.
+constexpr int kInvalidCatalogId = -1;
```
zuyu: Move this below the comment `This depends on all the above id types being typedefed to int...`, as the comment applies to it as well.

Also, we need a separate PR to replacing all `-1` that represents invalid ids with this new value in `/catalog`.

transaction/AccessMode.hpp
```diff
@@ -0,0 +1,171 @@
...
+      access_mode_underlying_type;
+
+  // The compatibility matrix should be N by N. kNumberLocks == N.
+  static constexpr std::uint64_t kNumberLocks =
+      static_cast<access_mode_underlying_type>(
+          AccessModeType::kNumAccessModeTypes);
+
+  // Compatibility matrix for checking access modes.
+  // True means they are compatible.
+  static constexpr bool kLockCompatibilityMatrix[kNumberLocks][kNumberLocks] = {
```
zuyu: I prefer to use `AccessModeType::kNumAccessModeTypes`, which is `uint8_t`, instead of another static member `kNumberLocks`.

hakanmemisoglu: It is not actually uint8_t (but the representation is). It is not implicitly convertible. So you have to write:
`static constexpr bool kLockCompatibilityMatrix[static_cast<access_mode_underlying_type>(AccessModeType::kNumAccessModeTypes][static_cast<access_mode_underlying_type>(AccessModeType::kNumAccessModeTypes)]`

catalog/CatalogTypedefs.hpp
hakanmemisoglu: I have 2 questions:
1-) Should I also delete my comment?
2-) Should I change the copyright?

transaction/ResourceId.hpp
zuyu: Fine, we can leave it here for now.

catalog/CatalogTypedefs.hpp
zuyu: 1. No, but you need to move your comment along with `kInvalidCatalogId`, below the comment I mentioned above.
2. Yes, do update the copyright for every file you touch.

transaction/AccessMode.hpp
zuyu: OK, ignore this.

hakanmemisoglu:
Hi @zuyu, I pushed the recent changes, and Travis seems happy.

transaction/tests/Lock_unittest.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "transaction/ResourceId.hpp"
+
+#include "gtest/gtest.h"
+
+namespace quickstep {
+namespace transaction {
+
+class LockTest : public ::testing::Test {
+ protected:
+  LockTest()
```
zuyu: Please typically use `SetUp` method to initialize all data members.

If you need an example, look at `Catalog_unittest.cpp`.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include "transaction/AccessMode.hpp"
+
+#include "gtest/gtest.h"
+
+namespace quickstep {
+namespace transaction {
+
+class AccessModeTest : public ::testing::Test {
+ protected:
+  AccessModeTest()
```
zuyu: Use `SetUp` to do initialization.

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,185 @@
...
+  const AccessMode s_mode_expected = AccessMode(AccessModeType::kSLock);
+  const AccessMode six_mode_expected = AccessMode(AccessModeType::kSixLock);
+  const AccessMode x_mode_expected = AccessMode(AccessModeType::kXLock);
+
+  const AccessMode nl_mode_checked = AccessMode(AccessModeType::kNoLock);
+  const AccessMode is_mode_checked = AccessMode(AccessModeType::kIsLock);
+  const AccessMode ix_mode_checked = AccessMode(AccessModeType::kIxLock);
+  const AccessMode s_mode_checked = AccessMode(AccessModeType::kSLock);
+  const AccessMode six_mode_checked = AccessMode(AccessModeType::kSixLock);
+  const AccessMode x_mode_checked = AccessMode(AccessModeType::kXLock);
```
zuyu: Remove all these `*_mode_expected` and `*_mode_checked`, and use data members in `AccessModeTest` directly, e.g., `nl_mode_`.

Thus it looks like:

```
  EXPECT_EQ(nl_mode_, nl_mode_);
  EXPECT_NE(nl_mode_, is_mode_);
```

transaction/tests/AccessMode_unittest.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include "gtest/gtest.h"
+
+namespace quickstep {
+namespace transaction {
+
+class AccessModeTest : public ::testing::Test {
+ protected:
+  AccessModeTest()
+    : nl_mode_(AccessModeType::kNoLock),
```
zuyu: Add two more whitespace indents.

transaction/tests/Lock_unittest.cpp
```diff
@@ -0,0 +1,102 @@
...
+  const std::vector<Lock> locks_on_resource_a_;
+  const std::vector<Lock> locks_on_resource_b_;
+  const std::vector<Lock> locks_on_resource_a_copy_;
+};
+
+TEST_F(LockTest, LockEquality) {
+  // Locks are equal if they are on the same resource with same
+  // access mode.
+  for (std::size_t i = 0; i < locks_on_resource_a_.size(); ++i) {
+    for (std::size_t j = 0; j < locks_on_resource_a_copy_.size(); ++j) {
```
zuyu: You don't actually a copy, use `locks_on_resource_a_` to replace `locks_on_resource_a_copy_`, and remove `locks_on_resource_a_copy_` from `LockTest`.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/AccessMode_unittest.cpp
hakanmemisoglu: Noted.

transaction/tests/Lock_unittest.cpp
hakanmemisoglu: Noted.

zuyu: Remove `resource_a_copy_` as well.

zuyu:
@hakanmemisoglu After fixing a minor issue in `Lock_unittest` and squashing the commits, we could merge.

transaction/tests/Lock_unittest.cpp
hakanmemisoglu: Noted.

hakanmemisoglu:
I fixed the issue. I will squash the commits.

zuyu:
@hakanmemisoglu Please remove the following words in the commit message:

```
Fixed typos.

Applied the review suggestions.

Applied the fixes.

Applied iwyu fixes for includes.

Unit test for Lock class is added.

Change the placeholder values as invalid ids that are previously used.

Fixed CMakeLists.txt for ResourceId.

Use of invalid id variable instead of magic numbers.

AccessMode underlying type refactored.

Fixed test bloat.

Updated copyright, moved the declaration of kInvalidId below.

Fixed Lock_unittest.
```

hakanmemisoglu:
Done.

zuyu:
LGTM!

PR #97, opened 2016-03-22
Refactored StorageManager's createBlock API.

zuyu:
This small PR refactors `createBlock` in `StorageManager`, as a part of replacing `CatalogRelation` with the read-only `CatalogRelationSchema` in `InsertDestination`.

PR #101, opened 2016-03-29
Fixes InsertDestination layout creation

cramja:
I noticed that BlockProperties were not being correctly instantiated because the associated StorageBlockLayoutDescription proto was not being selected by the InsertDestination. 

This was due to a change in how the InsertDestination was used in #97

### DIFF
--- a/storage/StorageBlockLayout.cpp
+++ b/storage/StorageBlockLayout.cpp
@@ -23,6 +23,7 @@
 
 #include "glog/logging.h"
 
+#include "catalog/CatalogDatabase.hpp"
 #include "catalog/CatalogRelationSchema.hpp"
 #include "storage/StorageBlockLayout.pb.h"
 #include "storage/StorageConstants.hpp"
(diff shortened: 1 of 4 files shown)


storage/StorageBlockLayout.cpp
```diff
@@ -136,6 +137,20 @@ void StorageBlockLayout::copyHeaderTo(void *dest) const {
   }
 }
 
+StorageBlockLayout* GenerateLayout(const CatalogRelationSchema &relationSchema) {
```
zuyu: Rename `relationSchema` to `relation_schema`.

storage/StorageBlockLayout.cpp
```diff
@@ -136,6 +137,20 @@ void StorageBlockLayout::copyHeaderTo(void *dest) const {
...
+  if (relationSchema.isTemporary()) {
+    // Temporary blocks use the default insert destination type.
+    return StorageBlockLayout::GenerateDefaultLayout(
+        relationSchema, relationSchema.isVariableLength());
+  } else {
+    // A non-temporary relation may have user-assigned block properties.
+    const CatalogRelation *relation = relationSchema.getParent().getRelationById(relationSchema.getID());
```
zuyu: `getParent`, which returns `nullptr` in the distributed case, won't work.

zuyu:
@cramja Your change doesn't make sense.

`InsertDestination::ReconstructFromProto` should reconstruct `InsertDestination` with the given storage block layout, set by the optimizer in the `InsertDestination` proto, if any.

cramja:
Hmm, okay I see this line in ReconstructFromProto
```c++
   StorageBlockLayout *layout = nullptr;
   if (proto.has_layout()) {
     // InsertDestination ctor will own this layout.
     layout = new StorageBlockLayout(relation, proto.layout());
   }
```

So there probably needs to be a change to `InsertOperator`

zuyu:
@cramja I'm not sure, but you might want to check `/query_optimizer/ExecutionGenerator.cpp`.

Btw, where is the definitions of `BlockProperties` proto?

cramja:
@zuyu I will check there. 

The `BlockProperties` works by translating the parsed BlockProperties clause into `StorageBlockLayoutDescription`. So no extra proto had been created.

cramja:
@zuyu , Closing this PR because I found the proper way to make this correction in the execution generator as you suggested.

PR #102, opened 2016-03-30
Added CatalogDatabaseLite.

zuyu:
This PR added the abstraction class called `CatalogDatabaseLite` for the catalog database, so that both the master `CatalogDatabase` and `CatalogDatabaseCache`, reconstructed by `Shiftboss` in the distributed version could be used to reconstruct `QueryContext` and its elements.

It also has minor fixes for `CatalogDatabase`, including variable alignments, and a bug fix when getting mutable relations by name.

Note that refactoring how `CatalogDatabase` stores `CatalogRelation`s (to avoid holes in the vector after relation deletions, and simplify deserialization) is not included by this PR.


pateljm:
LTGM. Merging.

PR #104, opened 2016-03-30
Inform InsertDestination of LayoutDescription

cramja:
From #101 (closed)
>I noticed that BlockProperties were not being correctly instantiated because the associated StorageBlockLayoutDescription proto was not being selected by the InsertDestination.

>This was due to a change in how the InsertDestination was used in #97

**This PR properly hands-off the layout description to the InsertDestination** It changes the ExecutionGeneration to include some extra info in the `InsertDestination` proto.

### DIFF
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -778,6 +778,8 @@ void ExecutionGenerator::convertCopyFrom(
 
   insert_destination_proto->set_insert_destination_type(S::InsertDestinationType::BLOCK_POOL);
   insert_destination_proto->set_relation_id(output_relation->getID());
+  insert_destination_proto->mutable_layout()->MergeFrom(
+      output_relation->getDefaultStorageBlockLayout().getDescription());
 
   const vector<block_id> blocks(physical_plan->catalog_relation()->getBlocksSnapshot());
   for (const block_id block : blocks) {
@@ -993,12 +995,13 @@ void ExecutionGenerator::convertInsertTuple(
 
   insert_destination_proto->set_insert_destination_type(S::InsertDestinationType::BLOCK_POOL);
   insert_destination_proto->set_relation_id(input_relation.getID());
+  insert_destination_proto->mutable_layout()->MergeFrom(
+      input_relation.getDefaultStorageBlockLayout().getDescription());
 
   const vector<block_id> blocks(input_relation.getBlocksSnapshot());
   for (const block_id block : blocks) {
     insert_destination_proto->AddExtension(S::BlockPoolInsertDestination::blocks, block);
   }
-
 
   const QueryPlan::DAGNodeIndex insert_operator_index =
       execution_plan_->addRelationalOperator(


cramja:
It seems like my local machine created a different name of method for the InsertDestinationProto, my machine created `set_allocated_layout` and Travis created `set_layout` (which is correct according to Protobuf's docs)

cramja:
Now I'm really confused, Travis apparently doesn't work with `set_layout` either.

`InsertDestination.proto` has the line:
```
optional StorageBlockLayoutDescription layout = 3;
```
in the InsertDestination message. According to [Google](https://developers.google.com/protocol-buffers/docs/cpptutorial#the-protocol-buffer-api), the method `set_layout` should have been generated. I am not sure why on my local machine it created `set_allocated_layout` instead of the expected `set_layout` and on Travis, neither worked.

cramja:
@zuyu you're good with proto messages - do you have any hints at what is going on here? 

```c++
insert_destination_proto->set_allocated_layout(
      new StorageBlockLayoutDescription(
          input_relation.getDefaultStorageBlockLayout().getDescription()));
```

compiles fine on my machine and the quickstep box but not travis. Travis gives the error that `set_allocated_layout` is a non-existant method.

zuyu:
@cramja You should search for `mutable_layout` in the codebase for more info, as `StorageBlockLayout` is a proto message, not a basic type.

cramja:
Yep, the code that my `protoc` generates:

```c++
class InsertDestination : public ::google::protobuf::Message {
 public:
  InsertDestination();
  virtual ~InsertDestination();

  ...

  // optional .quickstep.StorageBlockLayoutDescription layout = 3;
  inline bool has_layout() const;
  inline void clear_layout();
  static const int kLayoutFieldNumber = 3;
  inline const ::quickstep::StorageBlockLayoutDescription& layout() const;
  inline ::quickstep::StorageBlockLayoutDescription* mutable_layout();
  inline ::quickstep::StorageBlockLayoutDescription* release_layout();
  inline void set_allocated_layout(::quickstep::StorageBlockLayoutDescription* layout);

 ...
};
```

Noticing the `set_allocated_layout`, I'm not sure why this call didn't work on Travis.

I then poked around with your suggestion: `mutable_layout()` and discovered that you can use it:
```bash
(gdb) p insert_destination_proto->mutable_layout()
$2 = (quickstep::StorageBlockLayoutDescription *) 0x2ec5f20
(gdb) p insert_destination_proto->mutable_layout()->DebugString()
$3 = ""
(gdb) p insert_destination_proto->mutable_layout()->CopyFrom(input_relation.getDefaultStorage
BlockLayout().getDescription())
$4 = void
(gdb) p insert_destination_proto->mutable_layout()->DebugString()
$5 = "num_slots: 1\ntuple_store_description {\n  sub_block_type: BASIC_COLUMN_STORE\n  [quick
step.BasicColumnStoreTupleStorageSubBlockDescription.sort_attribute_id]: 0\n}\n"

```

cramja:
Squashed

zuyu:
@cramja You should read carefully about generated APIs in the protobuf document. `set_allocated_*` seems not use publicly.

Btw, what you have changed so far has the same effect as what `InsertDestination` constructor does if `layout` is `NULL`.

cramja:
@zuyu There is a minor difference as for what the `InsertDestination`, does. `InsertDestination` calls a static method of `StorageBlockLayout` (`GenerateDefaultLayout`). The change in this PR calls `CatalogRelation`'s method `getDefaultStorageBlockLayout` which, if set by BlockProperties will be the user's layout, or if not set, will be the same as `static GenerateDefaultLayout`.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -993,13 +995,15 @@ void ExecutionGenerator::convertInsertTuple(
...
   insert_destination_proto->set_insert_destination_type(S::InsertDestinationType::BLOCK_POOL);
   insert_destination_proto->set_relation_id(input_relation.getID());
+  insert_destination_proto->mutable_layout()->CopyFrom(
+      input_relation.getDefaultStorageBlockLayout().getDescription());
 
   const vector<block_id> blocks(input_relation.getBlocksSnapshot());
   for (const block_id block : blocks) {
-    insert_destination_proto->AddExtension(S::BlockPoolInsertDestination::blocks, block);
+    insert_destination_proto->AddExtension(
+        S::BlockPoolInsertDestination::blocks, block);
```
zuyu: This change is note related to this PR, and for the code, we have `120` char limits.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -778,6 +778,8 @@ void ExecutionGenerator::convertCopyFrom(
 
   insert_destination_proto->set_insert_destination_type(S::InsertDestinationType::BLOCK_POOL);
   insert_destination_proto->set_relation_id(output_relation->getID());
+  insert_destination_proto->mutable_layout()->CopyFrom(
+      output_relation->getDefaultStorageBlockLayout().getDescription());
```
zuyu: Use `MergeFrom` instead.

zuyu:
@cramja You need to update everywhere `insert_destination_proto` creates, in particular in `createTemporaryCatalogRelation`.

query_optimizer/ExecutionGenerator.cpp
cramja: Why is this a better choice than CopyFrom?

cramja:
@zuyu Assuming you mean `convertSample` in `ExecutionGenerator` what needs to change here? The idea behind the InsertTuple and CopyFrom converts was that a non-temporary block might be created. In sample, the created block is temporary as far as I can tell.

Also, I can file an issue for the gflag... perhaps it can be implemented as a `\command` for use at runtime.

query_optimizer/ExecutionGenerator.cpp
cramja: Talked with @saketj about this- he said it was more efficient because there'd be one less copy operation. IMO seems like a trivial difference unless the code were to be in a tight loop.

zuyu: Just use the minimum APIs when possible.

zuyu:
@cramja I assume `SampleOperator` should generate the same block layout for the result table, isn't it?

I prefer to use `gflag` for the default block layout in temp relations without setting `attribute_id`s. And if some complex block layout is needed, we could do either your `\command` way or derive from the original table.

cramja:
@zuyu Opened issue #108 in relation to the `gflag` recommendation.

>  I assume SampleOperator should generate the same block layout for the result table, isn't it?

The current policy with temporary tables is to use the DefaultLayout, a row or split rowstore.

cramja:
@zuyu  BTW, squashing doesn't seem to work too well when the commits are already on github. I bring this up because this PR is only 2 statements but 5 commits if I make the changes you pointed out.

I tried squashing yesterday with the instructions from [this guide](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit). However, github is keeping a record of the old commits.

If I open a new PR with these exact changes, it's for this reason.

zuyu:
@cramja No, you don't need a new PR for just squashing commits. The same PR works, and the guide shows the correct comments.

What does you mean for 
> 2 statements but 5 commits

This PR has only one commit.

cramja:
Oh, how strange. When I pushed the Squashed code yesterday and then looked at github, it showed as having the 4 original commits and then the final squash of them all. It probably just needed me to hit refresh in the browser.

zuyu:
Yes, you do need to refresh the page :-)

zuyu:
LGTM.

zuyu:
Please re-edit your commit message to remove these words:

```
Ditto for CopyFrom

My local machine created diiferent proto

Trying with the mutable method.
```

cramja:
Addressed

zuyu:
LGTM.

PR #106, opened 2016-03-31
Added CatalogDatabaseCache.

zuyu:
This PR added `CatalogDatabaseCache`, a subclass of `CatalogDatabaseLite` and a sibling class of `CatalogDatabase`, used in the distributed version to access the read-only `CatalogRelationSchema`s by `StorageManager` and `WorkOrder`s.

This PR also fixes the `iwyu` style issues for the `Catalog` unit test.


catalog/CatalogDatabaseCache.hpp
```diff
@@ -0,0 +1,124 @@
...
+   **/
+  explicit CatalogDatabaseCache(const serialization::CatalogDatabase &proto);
+
+  /**
+   * @brief Destructor which recursively destroys children.
+   **/
+  ~CatalogDatabaseCache() override {
+  }
+
+  bool hasRelationWithId(const relation_id id) const override {
```
pateljm: Add doxygen style comments to all the functions here.

catalog/CatalogDatabaseCache.hpp
```diff
@@ -0,0 +1,124 @@
...
+class CatalogRelationSchema;
+
+namespace serialization { class CatalogDatabase; }
+
+/** \addtogroup Catalog
+ *  @{
+ */
+
+/**
+ * @brief A database cache used in the distributed version.
```
pateljm: Quick description of the role of the cache and expected usage here would be great (can be refined over time) so that it is easier for a reader to understand how it fits in the overall architecture. Who uses it? When does a cache entry get invalidate? ...

catalog/CatalogDatabaseCache.hpp
zuyu: No need for overridden methods, if they have no additional info comparing to the virtual method.

catalog/CatalogDatabaseCache.hpp
zuyu: We could add more detailed comments once having `Shiftboss`, which would drop the query result table in the cache once the cli prints the result, if any.

For cache invalidations, we will drop a relation in the cache only as a part of executing a drop query for the relation. In fact, currently `Quickstep` does not support schema changes (a.k.a, ALTER statement).

pateljm: Ok -- can you make a note of it so we don't forget.

pateljm:
LTGM. Merging.

PR #109, opened 2016-03-31
Index creation for Compound Keys

saketj:
This is a small PR that changes the previous behaviour of CREATE INDEX from creating separate indices on each attribute to now creating an index with compound keys, when multiple attributes are specified. Effectively all the following SQL statements will succeed without error now:

```
CREATE TABLE foo (col1 INT, col2 INT);
CREATE INDEX csb_1 ON foo(col1) USING CSBTREE;
CREATE INDEX csb_2 ON foo(col2) USING CSBTREE;
CREATE INDEX csb_compound ON foo(col1, col2) USING CSBTREE;

```
The last SQL statement creates a CSB index with compound key made up of two attributes. 

Please also note that the following statement will now have the effect of creating a single index with a compound key made up of all attributes in the table:

`CREATE INDEX csb_compound ON foo USING CSBTREE;`

### DIFF
--- a/catalog/IndexScheme.cpp
+++ b/catalog/IndexScheme.cpp
@@ -17,10 +17,8 @@
 
 #include "catalog/IndexScheme.hpp"
 
-#include <cstddef>
 #include <memory>
 #include <string>
-#include <vector>
 
 #include "catalog/Catalog.pb.h"
 #include "storage/IndexSubBlockDescriptionFactory.hpp"
@@ -36,14 +34,11 @@ bool IndexScheme::ProtoIsValid(const serialization::IndexScheme &proto) {
     return false;
   }
 
-  // Check that each index entry contains initialized valid index descriptions.
+  // Check that each index entry contains initialized valid index description.
   for (int i = 0; i < proto.index_entries_size(); ++i) {
     const serialization::IndexScheme_IndexEntry &index_entry = proto.index_entries(i);
-    for (int j = 0; j < index_entry.index_description_size(); ++j) {
-      const IndexSubBlockDescription &index_description = index_entry.index_description(j);
-      if (!IndexSubBlockDescriptionFactory::ProtoIsValid(index_description)) {
-        return false;
-      }
+    if (!IndexSubBlockDescriptionFactory::ProtoIsValid(index_entry.index_description())) {
+      return false;
     }
   }
   return true;
@@ -58,16 +53,12 @@ IndexScheme* IndexScheme::ReconstructFromProto(const serialization::IndexScheme 
 
   for (int index_num = 0; index_num < proto.index_entries_size(); ++index_num) {
     const serialization::IndexScheme_IndexEntry &index_entry = proto.index_entries(index_num);
-    std::vector<IndexSubBlockDescription> index_descriptions;
-    for (int i = 0; i < index_entry.index_description_size(); ++i) {
-      index_descriptions.emplace_back(index_entry.index_description(i));
-    }
     // Make sure that index with same name does not already exist.
     DCHECK(index_scheme->index_map_.find(index_entry.index_name())
            == index_scheme->index_map_.end())
         << "Attempted to create IndexScheme from proto with duplicate index names.";
-    // Store the index_name and corresponding list of index descriptions in map.
-    index_scheme->index_map_.emplace(index_entry.index_name(), std::move(index_descriptions));
+    // Store the index_name and corresponding index description in map.
+    index_scheme->index_map_.emplace(index_entry.index_name(), index_entry.index_description());
   }
   return index_scheme.release();
 }
@@ -81,13 +72,7 @@ serialization::IndexScheme IndexScheme::getProto() const {
 
     // Populate the details of the index entry.
     index_entry->set_index_name(cit->first);
-    const std::vector<IndexSubBlockDescription> &index_descriptions = cit->second;
-    for (auto index_description_it = index_descriptions.begin();
-         index_description_it != index_descriptions.end();
-         ++index_description_it) {
-      // Add each index description to the index entry.
-      index_entry->add_index_description()->MergeFrom(*index_description_it);
-    }
+    index_entry->mutable_index_description()->MergeFrom(cit->second);
   }
   return proto;
 }
--- a/catalog/IndexScheme.hpp
+++ b/catalog/IndexScheme.hpp
@@ -20,10 +20,8 @@
 
 #include <algorithm>
 #include <cstddef>
-#include <memory>
 #include <string>
 #include <unordered_map>
-#include <vector>
 
 #include "catalog/Catalog.pb.h"
 #include "storage/StorageBlockLayout.pb.h"
@@ -103,17 +101,13 @@ class IndexScheme {
    * @param index_descripton Index Description to check against.
    * @return Whether a similar index description was already defined or not.
    **/
-  bool hasIndexWithDescription(const IndexSubBlockDescription &index_description) const {
-    // Iterate through every vector of descriptions corresponding to all keys in the index map.
+  bool hasIndexWithDescription(const IndexSubBlockDescription &index_description_checked) const {
+    // Iterate through every index description corresponding to each key in the index map.
     for (auto cit = index_map_.cbegin(); cit != index_map_.cend(); ++cit) {
-      const std::vector<IndexSubBlockDescription> &index_descriptions = cit->second;
-      for (auto index_description_it = index_descriptions.cbegin();
-           index_description_it != index_descriptions.cend();
-           ++index_description_it) {
-        // Check if the stored description matches the given description.
-        if (areIndexDescriptionsSame(*index_description_it, index_description)) {
-          return true;
-        }
+      const IndexSubBlockDescription &index_description_expected = cit->second;
+      // Check if the stored description matches the given description.
+      if (areIndexDescriptionsSame(index_description_expected, index_description_checked)) {
+        return true;
       }
     }
     return false;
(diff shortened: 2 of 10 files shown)


relational_operators/CreateIndexOperator.hpp
```diff
@@ -67,7 +67,7 @@ class CreateIndexOperator : public RelationalOperator {
                       const std::vector<IndexSubBlockDescription> &index_descriptions)
       : relation_(DCHECK_NOTNULL(relation)),
         index_name_(index_name),
-        index_descriptions_(index_descriptions) {}
+        index_descriptions_(std::move(index_descriptions)) {}
```
zuyu: You should change Line 67 to `rvalue`, as well as use `move` in `ExecutionGenerator::convertCreateIndex`.

cramja:
@saketj and I spoke about this change in relation to added SMA to the Create Index SQL command. Apart from the `rval` flaw, the change related to indexed attributes LGTM.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -830,21 +830,21 @@ void ExecutionGenerator::convertCreateIndex(
...
+  }
+  if (input_relation->hasIndexWithDescription(index_description)) {
+    // Check if the given index description already exists in the relation.
+    THROW_SQL_ERROR() << "The relation " << input_relation->getName()
+        << " already defines this index on the given attribute(s).";
+  } else if (!index_description.IsInitialized()) {
+    // Check if the given index description is valid.
+    THROW_SQL_ERROR() << "The index with given properties cannot be created.";
+  } else {
+    index_descriptions.push_back(std::move(index_description));
```
zuyu: @saketj If I understand correctly, `index_descriptions` has one and only one element, right? If so, why we need a `vector`?

relational_operators/CreateIndexOperator.hpp
zuyu: For this file, run `iwyu`. At lease, you need `<utility>`.

query_optimizer/ExecutionGenerator.cpp
saketj: @zuyu Yeah, you are right. I was thinking about it too earlier and was wondering whether I should refactor it. It will make this PR a little longer, if I do that. It might also evolve in the future to have multiple index descriptions for some indices, so we may end up using a vector later. Should I refactor it?

zuyu: Yes, I think we should remove the vector for now, as the physical plan only produces only one `IndexSubBlockDescription`.

```
IndexSubBlockDescription index_description(*physical_plan->index_description());
```

FYI, the `move` usage is still applied for the single object.

saketj: @zuyu I have removed the vector of index descriptions and refactored the rest of the code around it. I will squash all the commits, once it passes the review.

catalog/IndexScheme.hpp
```diff
@@ -150,27 +146,22 @@ class IndexScheme {
...
    *       necessary locks before invoking it.
    *
-   * @param index_name The name of index to add (key)
-   * @param index_description The index description for this index (value)
+   * @param index_name The name of index to add (key).
+   * @param index_description The index description for this index (value).
    **/
   bool addIndexMapEntry(const std::string &index_name,
-                        const std::vector<IndexSubBlockDescription> &index_descriptions) {
+                        const IndexSubBlockDescription &index_description) {
```
zuyu: Since you `move index_description` later, it should not be marked as `const`. Instead, change to `IndexSubBlockDescription &&index_description` and the caller to use `move`.

catalog/IndexScheme.cpp
```diff
@@ -58,16 +55,12 @@ IndexScheme* IndexScheme::ReconstructFromProto(const serialization::IndexScheme
...
-      index_descriptions.emplace_back(index_entry.index_description(i));
-    }
     // Make sure that index with same name does not already exist.
     DCHECK(index_scheme->index_map_.find(index_entry.index_name())
            == index_scheme->index_map_.end())
         << "Attempted to create IndexScheme from proto with duplicate index names.";
-    // Store the index_name and corresponding list of index descriptions in map.
-    index_scheme->index_map_.emplace(index_entry.index_name(), std::move(index_descriptions));
+    // Store the index_name and corresponding index description in map.
+    index_scheme->index_map_.emplace(index_entry.index_name(), std::move(index_entry.index_description()));
```
zuyu: I don't know why the compiler allows `move` on `index_entry`, a `const` referenced object, but I would remove `move`. It is ok to call copy constructor here.

relational_operators/CreateIndexOperator.hpp
```diff
@@ -19,25 +19,20 @@
...
-#include <vector>
 
-#include "catalog/CatalogAttribute.hpp"
-#include "catalog/CatalogDatabase.hpp"
 #include "catalog/CatalogRelation.hpp"
 #include "relational_operators/RelationalOperator.hpp"
 #include "storage/StorageBlockLayout.hpp"
 #include "storage/StorageBlockLayout.pb.h"
 #include "utility/Macros.hpp"
+#include "tmb/id_typedefs.h"
```
zuyu: Reset the change for this. We treat `tmb` as a third party.

relational_operators/CreateIndexOperator.hpp
```diff
@@ -55,19 +50,16 @@ class CreateIndexOperator : public RelationalOperator {
...
-   *
    * @param relation The relation to create index upon.
    * @param index_name The index to create.
-   * @param index_descriptions Set of index_descriptions associated with this index.
+   * @param index_description The index_description associated with this index.
    **/
   CreateIndexOperator(CatalogRelation *relation,
                       const std::string &index_name,
-                      const std::vector<IndexSubBlockDescription> &index_descriptions)
+                      IndexSubBlockDescription&& index_description)  // NOLINT
```
zuyu: Change to `IndexSubBlockDescription &&index_description)`, and remove the comment for `NOLINT` (we don't need that hint for the style checker anymore).

zuyu:
LGTM, except for some minor issues.

relational_operators/CreateIndexOperator.hpp
saketj: @zuyu I tried using this syntax too, but somehow I am still getting the following lint error from style checker. 
`/catalog/CatalogRelation.hpp:270:  Missing spaces around &&  [whitespace/operators] [3]`
Is it something that I have change in the lint code itself to make this pass?

catalog/IndexScheme.cpp
```diff
@@ -81,13 +74,8 @@ serialization::IndexScheme IndexScheme::getProto() const {
...
     index_entry->set_index_name(cit->first);
-    const std::vector<IndexSubBlockDescription> &index_descriptions = cit->second;
-    for (auto index_description_it = index_descriptions.begin();
-         index_description_it != index_descriptions.end();
-         ++index_description_it) {
-      // Add each index description to the index entry.
-      index_entry->add_index_description()->MergeFrom(*index_description_it);
-    }
+    const IndexSubBlockDescription &index_description = cit->second;
+    index_entry->mutable_index_description()->MergeFrom(index_description);
```
saketj: Fixed.

catalog/IndexScheme.hpp
saketj: I have removed the const from here and applied move semantics at caller.

catalog/IndexScheme.cpp
saketj: Updated.

relational_operators/CreateIndexOperator.hpp
saketj: Done.

saketj:
@zuyu  Thanks for the comments. I have squashed the commits and applied all the suggested edits except the `//NOLINT` issue for rvalue references. No matter what syntax I use, I am still getting the whitespace error around `&&` from the style checker. I guess, the style checker fails to identify rvalue references from normal operator usage of `&&. I have left it, with `//NOLINT` hints currently.

zuyu:
Leaving `NOLINT` is ok, and just for your info `rvalue` should work:

https://github.com/pivotalsoftware/quickstep/blob/master/relational_operators/SortMergeRunOperator.hpp#L234

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -830,26 +830,23 @@ void ExecutionGenerator::convertCreateIndex(
     index_attributes.emplace_back(catalog_attribute);
   }
 
-  // Corresponding to each attribute, create a copy of index description.
-  std::vector<IndexSubBlockDescription> index_descriptions;
+  // Create a copy of index description and add all the specified attributes to it.
```
zuyu: Since you're here, refactor `input_relation` in Line 812 to

```
  const CatalogRelation &input_relation = *input_relation_info->relation;
```

and then replace `input_relation->` with `input_relation.`.

relational_operators/CreateIndexOperator.hpp
```diff
@@ -85,7 +78,7 @@ class CreateIndexOperator : public RelationalOperator {
  private:
   CatalogRelation *relation_;
   const std::string &index_name_;
-  const std::vector<IndexSubBlockDescription> index_descriptions_;
+  IndexSubBlockDescription index_description_;
```
zuyu: Mark `const`.

zuyu:
@saketj Sorry for two late comments. After addressing them, it should be ok to merge then. Thanks!

relational_operators/CreateIndexOperator.hpp
saketj: @zuyu I removed const from this member instance because I am moving this in updateCatalogOnCompletion() method. Should I then favour not moving this and pass it by reference?

saketj:
I see. So, it should have worked without `NOLINT` too. Strangely, I also found another instance of this `NOLINT` usage at https://github.com/pivotalsoftware/quickstep/blob/master/utility/textbased_test/TextBasedTest.hpp
This reminds me to also add the specific lint to be escaped `NOLINT(whitespace/operators)`. Thanks!

relational_operators/CreateIndexOperator.hpp
zuyu: No, ignore this. Thank you for this hint!

zuyu:
@saketj The file you provided is too old before I enabled the style checker to allow `c++14`.

zuyu:
@saketj Btw, have you run `iwyu` again, it seems that at lease `relational_operators/CreateIndexOperator.[ch]pp` have issues.

```
relational_operators/CreateIndexOperator.hpp should remove these lines:
- #include "catalog/CatalogRelation.hpp"  // lines 23-23
- #include "storage/StorageBlockLayout.hpp"  // lines 25-25

relational_operators/CreateIndexOperator.cpp should add these lines:
#include <utility>                      // for move
#include "catalog/CatalogRelation.hpp"  // for CatalogRelation
```

So, please run `iwyu` for all files touched by this PR.

saketj:
@zuyu Oh, I missed the IWYU for this file, after I added back the move function to updateCatalogOnCompletion() in last commit. I will fix this. Thanks!

zuyu:
Finally, the copyright of `CreateIndexOperator.[ch]pp` should have

```
 *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
 *     University of Wisconsin—Madison.
```

and remove the lines including `Quickstep Technologies LLC` and `Pivotal Software, Inc`.

Do so for all the files you've created.

zuyu:
Finally, use `// NOLINT(whitespace/operators)` wherever you need the `NOLINT`.

query_optimizer/ExecutionGenerator.cpp
saketj: @zuyu The `input_relation_info->relation` expression returns a const reference to the relation, whereas the ExecutionGenerator needs to pass a mutable reference of CatalogRelation to CreateIndexOperator here. So, I guess this change would then not apply.

zuyu: Oh, you're right. I thought we just need a `const` version.

saketj:
Added //NOLINT(whitespace/operators), updated license and IWYU. I have also squashed commits now. Thanks!

zuyu:
LGTM!

zuyu:
@saketj It turns out that it should be a bug in the style checker (aka `cpplint`) that does not recognize your statement that contains `&&` as `rvalue`, instead of  an `and` operator.

PR #111, opened 2016-04-01
Allows GROUP BY without aggregate expressions

jianqiao:
This PR allows `GROUP BY` without aggregate expressions in a query, e.g.
```
SELECT x, y + z
FROM some_relation
GROUP BY x, y, z;
```

It is a pre-PR for the distinct aggregation feature.
Tests will be added shortly after.

### DIFF
--- a/storage/AggregationOperationState.cpp
+++ b/storage/AggregationOperationState.cpp
@@ -32,6 +32,7 @@
 #include "expressions/aggregation/AggregateFunction.hpp"
 #include "expressions/aggregation/AggregateFunctionFactory.hpp"
 #include "expressions/aggregation/AggregationHandle.hpp"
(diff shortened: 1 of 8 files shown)


storage/AggregationOperationState.cpp
```diff
@@ -79,58 +78,73 @@ AggregationOperationState::AggregationOperationState(
     group_by_types.emplace_back(&group_by_element->getType());
   }
 
-  // Set up each individual aggregate in this operation.
```
jianqiao: Note: there is no big change starting from here though the file diff shows large chunks of modification. The modifications are just
```
if (aggregate_functions.size() == 0) {
  // New code here
} else {
  // The original big chunk of code indented two more spaces.
}

pateljm:
LTGM. Merging.

PR #112, opened 2016-04-01
Added the query id in QueryHandle.

zuyu:
This small PR adds the query id in `QueryHandle`, which will be used to detect different queries once supporting concurrent queries.


pateljm:
LTGM. Merging.

PR #113, opened 2016-04-02
Use const CatalogDatabase reference in QueryContext.

zuyu:
This small PR refactored to use the const reference of `CatalogDatabaseLite` in `QueryContext`, instead of a mutable pointer, as a reconstructed `InsertDestination` in `QueryContext` does not directly update the catalog any more (#98).


query_execution/Foreman.hpp
```diff
@@ -122,7 +122,7 @@ class Foreman final : public ForemanLite {
    **/
   inline void reconstructQueryContextFromProto(const serialization::QueryContext &proto) {
     query_context_.reset(
-        new QueryContext(proto, catalog_database_, storage_manager_, foreman_client_id_, bus_));
+        new QueryContext(proto, *catalog_database_, storage_manager_, foreman_client_id_, bus_));
```
pateljm: Why not `&catalog_database_`?

relational_operators/tests/AggregationOperator_unittest.cpp
```diff
@@ -260,7 +260,7 @@ class AggregationOperatorTest : public ::testing::Test {
 
     // Set up the QueryContext.
     query_context_.reset(new QueryContext(query_context_proto,
-                                          db_.get(),
+                                          *db_,
```
pateljm: Why not `&db_`? Same comment applies below.

zuyu: `db_` is a `unique_ptr`. So `*db_` returns the reference.

query_execution/Foreman.hpp
zuyu: `catalog_database_` is a pointer, so `* catalog_database_` is a reference.

pateljm:
LTGM. Merging.

PR #114, opened 2016-04-03
Use an explicit Agent (TMB Client) to send messages to Foreman.

zuyu:
This PR refactored to use an explicit TMB client (called `agent`) to send messages on behalf of `InsertDestination`s and some `WorkOrder`s to Foreman.

The previous implementation uses a `globally accessible map` to implicitly use `Worker`'s client id as sender, but it is hard to understand / debug.

In the distributed version, `Shiftboss` would act as the `agent` so that Foreman could have data locality information based on the message sender's client id for better scheduling on later `WorkOrder`s.


pateljm:
LTGM. @hbdeshmukh if you have questions, perhaps we can address in a separate PR. Merging to let @zuyu move on to the next PR.

PR #115, opened 2016-04-03
Optimized BulkInsert code.

cramja:
This minor change to the `bulkInsert` methods of PackedRowStore improves performance by as much as 70%.

Results can be viewed on [this spreadsheet](https://docs.google.com/spreadsheets/d/1FGJr6HqUVkuVAqJwo9EI0CYKw2-cezt9x03aZlXy_NY/edit?usp=sharing). Click to the tab called **Optimize RowstoreInsert** and the changes here are refered to a 'Opt # 2'.

*Note:* I tried the same technique on SplitRowStore, and the result was a slight slow-down. Therefore, I have not included them in this PR but the changes for SplitRow exist in the branch `Optimize-PackedRowstore`. Results are on the [same spreadsheet](https://docs.google.com/spreadsheets/d/1FGJr6HqUVkuVAqJwo9EI0CYKw2-cezt9x03aZlXy_NY/edit?usp=sharing), under the tab "Insert SplitRow".

### DIFF
--- a/storage/PackedRowStoreTupleStorageSubBlock.cpp
+++ b/storage/PackedRowStoreTupleStorageSubBlock.cpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -23,6 +25,7 @@
 
 #include "catalog/CatalogAttribute.hpp"
 #include "catalog/CatalogRelationSchema.hpp"
+#include "catalog/CatalogTypedefs.hpp"
 #include "storage/PackedRowStoreValueAccessor.hpp"
 #include "storage/StorageBlockInfo.hpp"
 #include "storage/StorageBlockLayout.pb.h"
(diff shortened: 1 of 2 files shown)


storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -143,46 +143,72 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
...
+         attr_it != relation_.end();
+         ++attr_it) {
+      attrs_max_size.push_back(attr_it->getType().maximumByteLength());
+      attrs_total_size += attrs_max_size.back();
+    }
+
     if (num_nullable_attrs != 0) {
+      // Index i of null_attr_idxs is an index greater than 0 if the i-th
+      // attribute of this relation is nullable, -1 otherwise.
+      std::vector<int> null_attr_idxs;
```
zuyu: Use `attribute_id`, instead of `int`.

Rename to `nullable_attr_indices`, as you seem referring `nullable` attributes, not `NULL` attributes (aka, deleted attributes during schema changes, although we are not supporting yet:-)).

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -143,46 +143,72 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
...
+      for (CatalogRelationSchema::const_iterator attr_it = relation_.begin();
+           attr_it != relation_.end(); ++attr_it) {
```
zuyu: Prefer to have a new line for `++attr_it) {`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -143,46 +143,72 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
...
+        null_attr_idxs.push_back(
+            relation_.getNullableAttributeIndex(attr_it->getID()));
+      }
```
zuyu: Better if you could add a new method called `getNullableAttributeIndices()` in `CatalogRelationSchema` to return the vector as a whole.

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -143,46 +143,72 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
...
-        for (CatalogRelationSchema::const_iterator attr_it = relation_.begin();
-             attr_it != relation_.end();
-             ++attr_it) {
-          const std::size_t attr_size = attr_it->getType().maximumByteLength();
-          const int nullable_idx = relation_.getNullableAttributeIndex(
-              attr_it->getID());
-          if (nullable_idx != -1) {
+        for (std::size_t curr_attr = 0; curr_attr < num_attrs; ++curr_attr) {
+          // If this attribute is nullable, check for a returned null value.
+          if (null_attr_idxs[curr_attr] != -1) {
```
zuyu: Use `kInvalidCatalogId` from `/catalog/CatalogTypedefs.hpp`, instead of the magic number `-1`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -143,46 +143,72 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
...
+            relation_.getNullableAttributeIndex(attr_it->getID()));
+      }
+
       while (this->hasSpaceToInsert<true>(1) && accessor->next()) {
-        attribute_id accessor_attr_id = 0;
-        for (CatalogRelationSchema::const_iterator attr_it = relation_.begin();
-             attr_it != relation_.end();
-             ++attr_it) {
-          const std::size_t attr_size = attr_it->getType().maximumByteLength();
-          const int nullable_idx = relation_.getNullableAttributeIndex(
```
zuyu: You don't have to remove this one, but just change the assigned value to `null_attr_idxs[curr_attr]`.

Btw, replace `int` with `attribute_id`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
zuyu: Actually, I did not get that why you need to add `null_attr_idxs`, instead of using the existing method.

zuyu:
@cramja There are cpplint errors to fix, so I cancelled the build.

pateljm:
Thanks @zuyu. Nice comments. Neat work @cramja -- this is an awesome improvement - thanks! Would love to find out more about the slowdown with the SplitRowStore - we can discuss tomorrow in the group meeting and add the notes to the weekly meeting log.

storage/PackedRowStoreTupleStorageSubBlock.cpp
cramja: Is this:
```c++
std::vector<attribute_id> getNullableAttributeIndices() const {
    std::vector<attribute_id> copy = nullable_attribute_indices_;
    return copy;
  }
```

The correct way to implement this method in CatalogRelationSchema? (I'm worried about double copy-ing.)

storage/PackedRowStoreTupleStorageSubBlock.cpp
cramja: @zuyu As in, that because the getter is `inline` that it will be just as fast?

zuyu: I guess they have the similar performance, but your new way needs an extra stack space for `null_attr_idxs`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
zuyu: Actually, as I pointed out before, we don't need `null_attr_idxs` at all.

Use the previous implementation including `attr_size` and `nullable_idx` is good enough, which should have the similar performance as yours, but without the stack space for `null_attr_idxs`.

zuyu:
@cramja Could you also please describe the benchmark results and the command to reproduce them? A great example is #31 by @pateljm.

cramja:
@zuyu TY for comments, issues are addressed. Also I linked the benchmark results in the description. It's on google sheets and is globally viewable.

zuyu:
@cramja You may want to only contain the changes related to `PackedRowStore`, and save those for `SplitRowStore` in another PR.

zuyu:
@cramja Sorry, never mind for the previous comment. Two branch names are too similar.

cramja:
@zuyu They're similar because of the way I made the changes during development. This PR only focuses on PackedRow while the other branch has changes for SplitRow (but that are undesirable).

cramja:
If Travis passes, is this Good To Go?

pateljm:
LTGM. @zuyu?

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -143,46 +144,66 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
...
-              attr_it->getID());
-          if (nullable_idx != -1) {
+        for (std::size_t curr_attr = 0; curr_attr < num_attrs; ++curr_attr) {
+          const std::size_t attr_size = attrs_max_size[curr_attr];
+          const int nullable_idx = relation_.getNullableAttributeIndex(curr_attr);
```
zuyu: Use `attribute_id` to replace `int`.

Also, rename `nullable_idx` to `nullable_attr_index`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -205,45 +226,50 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuplesWithRemappedAttribu
...
```
zuyu: Ditto for `int` and `nullable_idx`.

zuyu:
LGTM, except for two comments above.

zuyu:
@cramja So the key idea in this PR is to eliminate virtual function calls to `Type:: maximumByteLength`, right?

Btw, you need to update the copyright for all the file you touched.

pateljm:
@zuyu: Basically moved the loop iterating over each attribute schema outside the tight loop. @cramja Once the copyright is fixed, I can merge.

cramja:
Updated Copyrights to both files. Changed int -> attribute_id.

@zuyu Yes, therefore eliminating the need to create an iterator object.

Also, @pateljm and I talked about refactoring vector creations like:
```c++
    // Create a vector containing the maximum sizes of the to-be extracted
    // attributes.
    // NOTE(Marc): This is an optimization so that we need not use an iterator
    // object in the following inner loops.
    std::vector<std::size_t> attrs_max_size;
    for (CatalogRelationSchema::const_iterator attr_it = relation_.begin();
         attr_it != relation_.end();
         ++attr_it) {
      attrs_max_size.push_back(attr_it->getType().maximumByteLength());
      attrs_total_size += attrs_max_size.back();
    }
```

To be calls to the `CatalogRelationSchema` in order to avoid repetitions.

zuyu:
LGTM!

zuyu:
@pateljm When you merge the PR, please copy the PR link (https://github.com/pivotalsoftware/quickstep/pull/115) in the commit message.

pateljm:
PR #115 LTGM. Merging.

PR #116, opened 2016-04-04
Initialize a database without worrying about the catalog.pb.bin file

pateljm:
Has this every happened to you, where you type in `./quickstep_cli_shell` and get and error message `Unable to read catalog from file qsstor/catalog.pb.bin`? 

If all you were trying to do is to start Quickstep and initialize an empty database, now you can do the following: `./quickstep_cli_shell -initialize_db=true` or ``./quickstep_cli_shell  -storage_path=./dbtest -initialize_db=true`. These commands will create the directories `./qsstor` or `./dbtest` respectively, **and** automatically create the required `catalog.pb.bin` file in that directory. If any of these directories already exist, then a fatal error is thrown, so that the user does not accidentally overwrite an existing database.

Notes:
- In a later PR, we should remove the catalog.pb.bin.orig file and the code that is used to generate it.
- The code in QuickstepCli.cpp is also ripe for refactoring. This should be another PR.

### DIFF
--- a/cli/QuickstepCli.cpp
+++ b/cli/QuickstepCli.cpp
@@ -142,7 +142,7 @@ DEFINE_string(worker_affinities, "",
               "the affinity mask of the Quickstep process, which typically "
               "means that they will all be runable on any CPU according to "
               "the kernel's own scheduling policy).");
-DEFINE_bool(initialize_db, true, "If true, initialize a database.");
+DEFINE_bool(initialize_db, false, "If true, initialize a database.");
 }  // namespace quickstep
 
 int main(int argc, char* argv[]) {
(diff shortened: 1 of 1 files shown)


cli/QuickstepCli.cpp
```diff
@@ -189,6 +198,45 @@ int main(int argc, char* argv[]) {
...
+    {
+      string path_name = "mkdir " + fixed_storage_path;
+      if (std::system(path_name.c_str())) {
+        LOG(FATAL) << "Failed when attempting to create the directory: " << fixed_storage_path << "\n";
+      }
+    }
+#endif
+
+    // Create the defaulty catalog file.
+    std::ofstream catalog_file(fixed_storage_path + "catalog.pb.bin");
```
zuyu: Use the already computed `catalog_path` instead.

zuyu:
@pateljm The idea behind this PR is great, and I suggest to create a new file called `GenerateInitialDefaultCatalog.hpp` to wrap the functionality, thus both the cli and `/catalog/GenerateInitialDefaultCatalog.cpp` could just call the static method in the new file.

pateljm:
@zuyu -- As noted above there is much in this file that needs  to be refactored. Will be a separate PR.

zuyu:
LGTM, and there will be a few follow-up PRs.

PR #118, opened 2016-04-04
Quickstep catalog print

rogersjeffreyl:
This PR adds the execution support for the following commands to the CLI

\d
This command lists all the tables in the database
```
quickstep> \d
       List of relations
 Name | Type |
+------+------+
 test    | table 
 test1  | table 
```

\d <table_name>
This command is the equivalent of describe table.  The columns and their types are  shown.
Additionally index information is also shown

```
quickstep> \d test
 Table "test"
Column     |Type       
+-----------+-----------+
a          |Int        
Indexes:
"test_index" CSB_TREE (a)
```

\dt 
Lists all the tables in the database

```
quickstep> \dt
       List of relations
 Name | Type |
+------+------+
 test    | table 
 test1  | table 
```

\dt< table_name>
Lists the table name from the list of tables in the database

```
quickstep> \dt test
       List of relations

 Name | Type |
+------+------+
 test    | table 

```


cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    column_width = attr_it->getType().isNullable() && (column_width < 4)
+                       ? 4
+                       : column_width;
+    max_column_width =
+        column_width > max_column_width ? column_width : max_column_width;
+  }
+  column_widths.push_back(max_column_width);
+  column_widths.push_back(max_column_width);
+
+  fprintf(out, "%*s \"%s\"", 6, "Table", table_name_val.c_str());
```
pateljm: Why use fprintf instead of iostreams?

cli/CommandExecutor.hpp
```diff
@@ -0,0 +1,96 @@
...
+class CommandExecutor {
+ public:
+  /**
+   * @brief Executes the command by calling the command handler.
+   *        
+   * @param statement The parsed statement from the cli.
+   * @param catalog_database The catalog information about the current database.
+   * @param out The stream where the output of the command has to be redirected to.      
+   */
+  static void executeCommand(const ParseStatement &statement,
```
cramja: ParseStatement -> ParseCommandStatement, no?

cli/QuickstepCli.cpp
```diff
@@ -289,8 +291,14 @@ int main(int argc, char* argv[]) {
         }
 
         if (result.parsed_statement->getStatementType() == ParseStatement::kCommand) {
-          // TODO(marc): Add executer to this parsed command statement.
```
cramja: Good to see this being taken care of!

cli/QuickstepCli.cpp
```diff
@@ -289,8 +291,14 @@ int main(int argc, char* argv[]) {
...
-          continue;
+          try {
```
cramja: Indent `try`

cli/CommandExecutor.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+
+/** \addtogroup CLI
+ *  @{
+ */
+
+/**
+ * @brief Execution handlers for various quickstep cli commands
+ */
+class CommandExecutor {
```
cramja: Basic structure question: Did you consider making the Executor class abstract, and then having subclasses which implement an `execute` method, similar to the `ParseCommandStatement` class' relation to the `ParseStatement` class?

cramja:
@rogersjeffreyl I added some initial comments but did not look at the workhorse code in the Command.cpp file. I'd say looks like a good start, especially since the line printing methods can be tricky to write correctly. I will be around this afternoon from 2-5 if you want to talk in person then.

cli/CommandExecutor.cpp
zuyu: We should avoid `c++ streams` when possible, and use `c stdio` instead for performance issues (no additional buffer sync overheads).

zuyu:
@rogersjeffreyl Before diving into the code, I have the following questions.

Do we have `\help`, `\h`, or similar to show the help messages for all the commands you added?

Are `\d` and `\dt` two alias to the same commends?

Also, is it possible to print the relations better? For example in the demonstrated result below, the print length of all the relation names should be the same, so is the type. And the bar should align with `+`.

```
quickstep> \d
       List of relations
  Name   | Type   |
+--------+--------+
 test    | table 
 test1   | table 
```

Thanks!

zuyu:
@rogersjeffreyl Please update the copyright information for all the files you touched.

rogersjeffreyl:
@zuyu 

\d and \dt are not exactly the same commands. They behave the same way when  they take no arguments.  Following is the output from postgres:
postgres=# \d
       List of relations
 Schema | Name  | Type  | Owner 
--------+-------+-------+-------
 public | test  | table | rl
 public | test1 | table | rl

postgres=# \dt
       List of relations
 Schema | Name  | Type  | Owner 
--------+-------+-------+-------
 public | test  | table | rl
 public | test1 | table | rl


However , when the table name is applied as argument the behavior changes:\

postgres=# \dt test
       List of relations
 Schema | Name | Type  | Owner 
--------+------+-------+-------
 public | test | table | rl


postgres=# \d test
     Table "public.test"
 Column |  Type   | Modifiers 
--------+---------+-----------
 a      | integer | 
Indexes:
    "test_index" btree (a)

rogersjeffreyl:
@zuyu  We  donot have  support for \h or \help yet.   Regarding the output, when i paste the output into github I think it gets formatted.  When I ran them on the cli the output is cleaner

cli/CommandExecutor.hpp
rogersjeffreyl: @cramja the definition of the  executeCommand does the casting before calling on the actual command executors. (This is similar to the design of  Resolver.cpp. )

cli/CMakeLists.txt
```diff
@@ -31,6 +31,18 @@ configure_file (
   "${CMAKE_CURRENT_SOURCE_DIR}/CliConfig.h.in"
   "${CMAKE_CURRENT_BINARY_DIR}/CliConfig.h"
 )
+add_library(quickstep_cli_CommandExecutor CommandExecutor.cpp CommandExecutor.hpp)
+target_link_libraries(quickstep_cli_CommandExecutor
```
zuyu: We typically separate `add_library` and `target_link_libraries`.

So, move `target_link_libraries` after `# Link dependencies:`, and sort in alphabet order.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+using std::string;
+
+namespace quickstep {
+
+void CommandExecutor::executeCommand(const ParseStatement &statement,
+                                     const CatalogDatabase *catalog_database,
+                                     FILE *out) {
+  const ParseCommand &command = static_cast<const ParseCommand &>(statement);
+  const PtrVector<ParseString> *arguments = command.arguments();
+  std::string command_str = command.command()->value();
```
zuyu: Use the reference to avoid a deep copy.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+  if (command_str.compare("\\dt") == 0) {
+    executeDescribeDatabase(arguments, catalog_database, out);
+  } else if (command_str.compare("\\d") == 0) {
```
zuyu: We may want to define all the command as static const c-strings (never `c++ string`) in a new file under `/cli`, including `\\dt`, `\\d`, and future `\\h`.

Also give each command a name, for example, `kCommandHelp` for `\\h`. So, perhaps `kCommandDisplay` and `kCommandDisplayTable`?

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+      executeDescribeTable(arguments, catalog_database, out);
+    }
+  } else {
+    THROW_SQL_ERROR_AT(command.command()) << "Invalid Command";
+  }
+}
+
+void CommandExecutor::executeDescribeDatabase(
+    const PtrVector<ParseString> *arguments,
+    const CatalogDatabase *catalog_database, FILE *out) {
```
zuyu: Why not use the reference instead?

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+  std::size_t relation_count = 0;
+  // Column width initialized to 6 to take into account the header name
+  // and the column value table
+  int max_column_width = 6;
```
zuyu: `6` is a magic number, so we should define it as `kMaxColumnWidth` in some header.

Btw, why we not compute the max width based on the length of the table name or something related?

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+  const CatalogRelation *relation;
```
zuyu: Remove the declaration here, and use the definition instead when you use it.

cli/CommandExecutor.cpp
rogersjeffreyl: the max width is computed in the later code based on the column width.  The header table is 5 characters. If the table names are lesser than 5 characters we  still need a column width of atleast 5 to display the column header. hence the initialization to 6.

cramja:
@rogersjeffreyl and I talked about whether to make separate command classes versus static methods in a single command class. We came to the conclusion that at this point, using static methods is fine because there is no persistent state.

Also, @zuyu , `\help` hasn't been created yet. The original idea was to have `\help` describe a SQL command usage, (`\help select` would show the user the possible syntax). However, this not implemented because of the large number of grammars that need to be described.

A simple `\help` could display a list of available `\` commands similar to sqlite's interface. A seperate command could be used for SQL interpretation, something like `\usage SQL`.

--my 2p

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+  if (arguments->size() == 0) {
+    for (auto rel_it = catalog_database->begin();
+         rel_it != catalog_database->end(); ++rel_it) {
```
zuyu: How about this:

```
for (const CatalogRelation &relation : *catalog_database) {
  ...
}
```

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+      int column_width = static_cast<int>(rel_it->getName().length());
+      max_column_width =
+          max_column_width < column_width ? column_width : max_column_width;
+      relation_count++;
```
zuyu: Move this statement out of the relation, and use `catalog_database->size()` to get the number of relations.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+  } else {
+    const ParseString &table_name = arguments->front();
+    const std::string &table_name_val = table_name.value();
+    relation = catalog_database->getRelationByName(table_name_val);
+    int column_width = static_cast<int>(relation->getName().length());
+    max_column_width =
+        max_column_width < column_width ? column_width : max_column_width;
+    relation_count++;
```
zuyu: It should always be `1`, right?

Also, never use `i++`, unless you need the value of `i` before `++`.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+  } else {
+    THROW_SQL_ERROR_AT(command.command()) << "Invalid Command";
+  }
+}
+
+void CommandExecutor::executeDescribeDatabase(
+    const PtrVector<ParseString> *arguments,
+    const CatalogDatabase *catalog_database, FILE *out) {
+  std::size_t relation_count = 0;
```
zuyu: Rename to `num_relations`.

cli/CommandExecutor.cpp
rogersjeffreyl: @zuyu 

Sure. I did think about giving names to the command. The ParseCommand class  returns  the command string. The only way to know what the command is would be  from the command string.  
Do you have any suggestions on what would be a better design of mapping the command strings to the names and the resolving the names from the command string that comes from the parser?

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    max_column_width =
+        max_column_width < column_width ? column_width : max_column_width;
+    relation_count++;
+  }
+  // only if we have relations
+  if (relation_count > 0) {
+    vector<int> column_widths;
+    column_widths.reserve(relation_count);
+    column_widths.push_back(max_column_width);
+    column_widths.push_back(max_column_width);
```
zuyu: Why you need to reserve the capacity, while only appending two elements?

I would refactor to `const vector<std::size_t> column_widths(2, max_column_width);`, assuming you will change the type of `max_column_width` to `std::size_t`.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    fputs("       List of relations\n\n", out);
+    fprintf(out, "%-*s|", max_column_width, " Name");
+    fprintf(out, "%-*s|", 6, " Type");
```
zuyu: Don't use the magic number `6`. Create a static const variable for that.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    fputc('\n', out);
```
zuyu: You could merge `\n` in the previous line as `%-*s|\n`, right?

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    PrintToScreen::printHBar(column_widths, out);
+    //  if there are no argument print the entire list of tables
+    //  else print the particular table only
+    if (arguments->size() == 0) {
+      for (auto rel_it = catalog_database->begin();
+           rel_it != catalog_database->end(); ++rel_it) {
```
zuyu: Suggest to use the `for-loop` mentioned above.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+        fprintf(out, " %-*s|", max_column_width, rel_it->getName().c_str());
+        fprintf(out, " %-*s", max_column_width, "table");
+        fputc('\n', out);
```
zuyu: Merge `\n` to the previous line.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+      }
+    } else {
+      fprintf(out, " %-*s|", max_column_width, relation->getName().c_str());
+      fprintf(out, " %-*s", max_column_width, "table");
+      fputc('\n', out);
```
zuyu: Merge `\n` to the previous line.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+}
+
+void CommandExecutor::executeDescribeTable(
+    const PtrVector<ParseString> *arguments,
+    const CatalogDatabase *catalog_database, FILE *out) {
+  const ParseString &table_name = arguments->front();
+  const std::string &table_name_val = table_name.value();
+  const CatalogRelation *relation =
+      catalog_database->getRelationByName(table_name_val);
+  vector<int> column_widths;
```
zuyu: Use `std::size_t` instead of `int`.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+  // initialize it to the number of columns that need to be printed
+  column_widths.reserve(2);
+
+  int max_column_width = 0;
+  for (CatalogRelation::const_iterator attr_it = relation->begin();
+       attr_it != relation->end(); ++attr_it) {
```
zuyu: Suggest to use `for (const CatalogAttribute &attr : *relation) { ... }`.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+      fprintf(out, " %-*s", max_column_width, "table");
+      fputc('\n', out);
+    }
+    fputc('\n', out);
+  }
+}
+
+void CommandExecutor::executeDescribeTable(
+    const PtrVector<ParseString> *arguments,
+    const CatalogDatabase *catalog_database, FILE *out) {
```
zuyu: Why not use reference for `catalog_database`?

cli/CommandExecutor.cpp
rogersjeffreyl: @zuyu  which statement?

cli/CommandExecutor.cpp
rogersjeffreyl: @zuyu you mean the argument size. 
No  there can be a \dt or \d without any argument

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+       attr_it != relation->end(); ++attr_it) {
+    // Printed column needs to be wide enough to print:
+    //   1. The attribute name (in the printed "header").
+    //   2. Any value of the attribute's Type.
+    //   3. If the attribute's Type is nullable, the 4-character string "NULL".
+    // We pick the largest of these 3 widths as the column width.
+    int column_width = static_cast<int>(attr_it->getDisplayName().length());
+    column_width = column_width < attr_it->getType().getPrintWidth()
+                       ? attr_it->getType().getPrintWidth()
+                       : column_width;
```
zuyu: Include `<algorithm>`, and refactor to `column_width = std::max(column_width, attr_it->getType().getPrintWidth())`, so that we don't call the virtual function twice.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+                       ? 4
+                       : column_width;
+    max_column_width =
+        column_width > max_column_width ? column_width : max_column_width;
+  }
+  column_widths.push_back(max_column_width);
+  column_widths.push_back(max_column_width);
+
+  fprintf(out, "%*s \"%s\"", 6, "Table", table_name_val.c_str());
+  fputc('\n', out);
```
zuyu: Merge `\n` into the previous line.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,200 @@
...
+  fprintf(out, "%-*s|", max_column_width, "Column");
+  fprintf(out, "%-*s", max_column_width, "Type");
+  fputc('\n', out);
```
zuyu: Merge `\n` into the previous line.

cli/CommandExecutor.cpp
pateljm: The alignment of the output is off, and we should make sure the output is nicely aligned along the column. 

e.g. Currently when we list the database table for SSB, it shows (notice the header row is mis-aligned): 

```quickstep> \d
       List of relations

 Name    | Type |
+---------+---------+
 lineorder| table    
 part     | table    
 supplier | table    
 customer | table    
 ddate    | table 
``` 
PostgreSQL is a good model to follow. Here are some output from it.

```
jignesh=# \d
          List of relations
 Schema |   Name    | Type  |  Owner  
--------+-----------+-------+---------
 public | customer  | table | jignesh
 public | ddate     | table | jignesh
 public | lineorder | table | jignesh
 public | part      | table | jignesh
 public | supplier  | table | jignesh
(5 rows)
```
and
```
jignesh=# \d
                        List of relations
 Schema |                 Name                  | Type  |  Owner  
--------+---------------------------------------+-------+---------
 public | averylongtablenamethatseemstoneverend | table | jignesh
 public | customer                              | table | jignesh
 public | ddate                                 | table | jignesh
 public | lineorder                             | table | jignesh
 public | part                                  | table | jignesh
 public | supplier                              | table | jignesh
(6 rows)
```
and 
```
jignesh=# \dt averylongtablenamethatseemstoneverend
                        List of relations
 Schema |                 Name                  | Type  |  Owner  
--------+---------------------------------------+-------+---------
 public | averylongtablenamethatseemstoneverend | table | jignesh
(1 row)
```

rogersjeffreyl: @pateljm Will fix the header alignment issue in the describe database command.

cli/CommandExecutor.cpp
rogersjeffreyl: @zuyu static void printHBar(const std::vector<int> &column_widths,
                        FILE *out); 

takes a vector of integers.  This is an existing method in PrintToScreen. Should we change this method to take a vector of std::size_t instead of integers?

cli/CommandExecutor.cpp
rogersjeffreyl: @zuyu relation  is  defined inside an else block and also used in a different place after the else block

cli/CommandExecutor.cpp
zuyu: We just store the commands as static const c-strings in a new file (called `CommandTypedef` or sth else), which does not do anything else, including name resolving.

This `executeCommand` method is used to resolve names.

cli/CommandExecutor.cpp
zuyu: `relation_count++`

cli/CommandExecutor.cpp
zuyu: I mean `relation_count`.

cli/CommandExecutor.cpp
zuyu: Yes, we should also change that.

cli/CommandExecutor.cpp
zuyu: OK. Ignore this. I missed that part.

rogersjeffreyl:
@pateljm 
have fixed the alignment issue in the output

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,189 @@
...
+    fputc('\n', out);
+  }
+}
+
+void CommandExecutor::executeDescribeTable(
+    const PtrVector<ParseString> *arguments,
+    const CatalogDatabase *catalog_database, FILE *out) {
+  const ParseString &table_name = arguments->front();
+  const std::string &table_name_val = table_name.value();
+  const CatalogRelation *relation =
```
cramja: Careful here. Unless I missed a check elsewhere in the code, `getRelationByName` will return NULL if a relation does not exist by that name.

cli/CommandExecutor.hpp
cramja: Gotcha!

cli/PrintToScreen.hpp
```diff
@@ -1,6 +1,6 @@
 /**
- *   Copyright 2011-2015 Quickstep Technologies LLC.
- *   Copyright 2015 Pivotal Software, Inc.
```
cramja: I'm pretty sure we are supposed to leave the old copyright.

cramja:
@rogersjeffreyl Are we going to add unit tests in this PR?

cli/PrintToScreen.hpp
rogersjeffreyl: @cramja @zuyu suggested to change the copyright.

rogersjeffreyl:
@cramja unit tests will be a separate pr

cli/PrintToScreen.hpp
zuyu: @cramja is right. No deletions for the copyright as a rule.

zuyu:
@rogersjeffreyl We typically do not merge new code without any unit test coverages. Please add the unit test in the same PR.

cli/CommandExecutor.cpp
zuyu: @rogersjeffreyl I took it back. I just found that we have to use `int` here, which will be used to specify `printf`'s field width in another file. Therefore, please revert any changes that uses `std::size_t`. Sorry for that!

cli/tests/command_executor/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+         "../quickstep_cli_tests_CommandExecutorTest"
+         "${CMAKE_CURRENT_SOURCE_DIR}/Dt.test"
+         "${CMAKE_CURRENT_BINARY_DIR}/Dt.test"
+         "${CMAKE_CURRENT_BINARY_DIR}/Dt/")
+
+
+# Create the folders where the unit tests will store their data blocks for the
+# duration of their test.
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/D)
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Dt)
```
cramja: Good idea to keep them separate. We ran into problems in the past with this.

cli/tests/command_executor/Dt.test
```diff
@@ -0,0 +1,42 @@
+CREATE TABLE FOO (col1 INT,
```
zuyu: Add the copyright in this file.

cli/tests/command_executor/D.test
```diff
@@ -0,0 +1,109 @@
...
+\d FOO2                   
+--
+ Table "FOO2"
+ Column                        | Type   
++------------------------------+-------+
+ col1                          | Int    
+ col2                          | Long   
+ col3                          | Double 
+ col4                          | Float  
+ averyverylongcolumnnamefortest| Char(5)
```
zuyu: Please print out one whitespace between the column name and `|`.

See @pateljm's example below:
```
jignesh=# \dt averylongtablenamethatseemstoneverend
                        List of relations
 Schema |                 Name                  | Type  |  Owner  
--------+---------------------------------------+-------+---------
 public | averylongtablenamethatseemstoneverend | table | jignesh
(1 row)
```

cli/tests/command_executor/D.test
```diff
@@ -0,0 +1,109 @@
...
+ Column| Type   
++------+-------+
+ col1  | Int    
+ col2  | Long   
+ col3  | Double 
+ col4  | Float  
+ col5  | Char(5)
+ Indexes
+  "FOO4_INDEX_2" CSB_TREE (col3, col4)
+  "FOO4_INDEX_1" CSB_TREE (col1, col2)
```
zuyu: Typo: `Indexes` -> `Indices`.

For `Indices`, could we also use the same format as a table?

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,182 @@
...
+      fprintf(out, "  \"%-*s\" %s", static_cast<int>(index_it->first.length()),
+              index_it->first.c_str(),
+              index_it->second.IndexSubBlockType_Name(
+                                  index_it->second.sub_block_type())
+                  .c_str());
+      fputc(' ', out);
+      fputc('(', out);
+      fprintf(out, "%s", relation->getAttributeById(index_it->second.indexed_attribute_ids(0))
+                               ->getDisplayName()
+                               .c_str());
```
zuyu: Refactor Line 166-170:

```
  const char *attribute_display_name = ...;
  fprintf(out, " (%s", attribute_display_name);
```

cramja:
@rogersjeffreyl

cramja:
Sorry @rogersjeffreyl I accidently closed the PR when I meant to cancel the comment I was writing.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,182 @@
...
+  // TODO(rogers): Add handlers for partitioning information.
+  if (relation->hasIndexScheme()) {
+    fprintf(out, "%*s\n", static_cast<int>(CommandExecutor::kInitMaxColumnWidth)+2, " Indexes");
+    const quickstep::IndexScheme &index_scheme = relation->getIndexScheme();
+    for (auto index_it = index_scheme.begin(); index_it != index_scheme.end();
+         ++index_it) {
+      fprintf(out, "  \"%-*s\" %s", static_cast<int>(index_it->first.length()),
+              index_it->first.c_str(),
+              index_it->second.IndexSubBlockType_Name(
+                                  index_it->second.sub_block_type())
```
zuyu: Please fix the indentation like this:

```
  index_it->second.IndexSubBlockType_Name(
      index_it->second.sub_block_type());
```

Note that we add four whitespace indentations when wrapping an expression, while two for a complete statement ended with `;`.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,182 @@
...
+                  .c_str());
+      fputc(' ', out);
+      fputc('(', out);
+      fprintf(out, "%s", relation->getAttributeById(index_it->second.indexed_attribute_ids(0))
+                               ->getDisplayName()
+                               .c_str());
+      for (std::size_t i = 1; i < index_it->second.indexed_attribute_ids_size(); ++i) {
+        fprintf(out, ", %s", relation->getAttributeById(index_it->second.indexed_attribute_ids(i))
+                               ->getDisplayName()
+                               .c_str());
```
zuyu: Suggest to extract the expression from `printf`:

```
  const char *attribute_display_name = ...;
  fprintf(out, ", %s", attribute_display_name);
```

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,182 @@
...
+      }
+      fputc(')', out);
+      fputc('\n', out);
```
zuyu: Prefer to combine two `fputc` into one `fprintf`, but it's up to you.

cli/CommandExecutor.hpp
```diff
@@ -0,0 +1,97 @@
...
+   * @param statement The parsed statement from the cli.
+   * @param catalog_database The catalog information about the current database.
+   * @param out The stream where the output of the command has to be redirected to.      
+   */
+  static void executeCommand(const ParseStatement &statement,
+                             const CatalogDatabase &catalog_database,
+                             FILE *out);
+
+ private:
+  static const std::size_t kInitMaxColumnWidth;
```
zuyu: Prefer to compilation-time constants, instead of run-time constants:

```
  // Add comments for why pick up the value.
  static constexpr std::size_t kInitMaxColumnWidth = 6;
```

https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

cli/CommandExecutor.hpp
```diff
@@ -0,0 +1,97 @@
...
+  CommandExecutor();
```
zuyu: Prefer to use global functions, instead of the current implementation:

https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions

With that being said, I suggest the following changes:
```C++
// .hpp file
#include <cstddef>  // for std::size_t

namespace quickstep {
namespace cli {

// Comments for why choose the value.
constexpr std::size_t kInitMaxColumnWidth = 6;

static const char *kDescribeDatabaseCommand;
static const char *kDescribeTableCommand;

/*
 * @brief ...
 */
void ExecuteCommand(...);

}  // namespace cli
}  // namespace quickstep
```

```C++
// .cpp file
namespace quickstep {
namespace cli {

const char *kDescribeDatabaseCommand = "\\d";
const char *kDescribeTableCommand = "\\dt";

namespace {

void ExecuteDescribeTable(...) {
  ...
}

void ExecuteDescribeTable() {
  ...
}

}  // namespace

void ExecuteCommand(...) {
  ...
}

}  // namespace cli
}  // namespace quickstep
```

cli/tests/command_executor/D.test
rogersjeffreyl: @zuyu i think indices and indexes are both correct usage. Postgres uses Indexes

postgres=# \d test
     Table "public.test"
 Column |  Type   | Modifiers 
--------+---------+-----------
 a      | integer | 
Indexes:
    "test_index" btree (a)

rogersjeffreyl:
@cramja @zuyu 
CI failed earlier due to the ref count issue. Have merged with master and started CI again.

rogersjeffreyl:
@cramja @zuyu  The CI has passed now.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+    const CatalogDatabase &catalog_database, FILE *out) {
+  // Column width initialized to 6 to take into account the header name
+  // and the column value table
+  int max_column_width = C::kInitMaxColumnWidth;
+  const CatalogRelation *relation;
+  if (arguments->size() == 0) {
+    for (const CatalogRelation &rel : catalog_database) {
+      int column_width = static_cast<int>(rel.getName().length());
+      max_column_width =
+          max_column_width < column_width ? column_width : max_column_width;
```
zuyu: How about the following refactor:

```
#include <algorithm>

  max_column_width =
      std::max(max_column_width,
               static_cast<int>(rel.getName().length()));
```

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+    const ParseString &table_name = arguments->front();
+    const std::string &table_name_val = table_name.value();
+    relation = catalog_database.getRelationByName(table_name_val);
+
+    if (relation == nullptr) {
+     THROW_SQL_ERROR_AT(&(arguments->front())) << " Unrecognized relation "  <<table_name_val;
+    }
+    int column_width = static_cast<int>(relation->getName().length());
+    max_column_width =
+        max_column_width < column_width ? column_width : max_column_width;
```
zuyu: Same refactor using `std::max` as above.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+      max_column_width =
+          max_column_width < column_width ? column_width : max_column_width;
+    }
+  } else {
+    const ParseString &table_name = arguments->front();
+    const std::string &table_name_val = table_name.value();
+    relation = catalog_database.getRelationByName(table_name_val);
+
+    if (relation == nullptr) {
+     THROW_SQL_ERROR_AT(&(arguments->front())) << " Unrecognized relation "  <<table_name_val;
```
zuyu: Please check the indentation before `THROW_SQL_ERROR_AT` and whitespace around `<< table_name_val`.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+void executeDescribeTable(
+    const PtrVector<ParseString> *arguments,
+    const CatalogDatabase &catalog_database, FILE *out) {
+  const ParseString &table_name = arguments->front();
+  const std::string &table_name_val = table_name.value();
+  const CatalogRelation *relation =
+      catalog_database.getRelationByName(table_name_val);
+  if (relation == nullptr) {
+     THROW_SQL_ERROR_AT(&(arguments->front())) << " Unrecognized relation "  <<table_name_val;
```
zuyu: Same as the above, the indentation and whitespace issues.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+  int max_type_column_width = C::kInitMaxColumnWidth;
+
+  for (const CatalogAttribute &attr : *relation) {
+    // Printed column needs to be wide enough to print:
+    //   1. The attribute name (in the printed "header").
+    //   2. Any value of the attribute's Type.
+    int attr_column_width = attr.getDisplayName().length();
+    int type_column_width = attr.getType().getName().length();
+    max_attr_column_width = std::max(max_attr_column_width, attr_column_width);
+    max_type_column_width = std::max(max_type_column_width, type_column_width);
```
zuyu: Please replace `attr_column_width` and `type_column_width` with their values in the assignment expressions for `max_attr_column_width` and `max_type_column_width`, respectively.

Otherwise, mark the former two `const`.

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+                                                 index_it->second.indexed_attribute_ids(i))
+                                                 ->getDisplayName().c_str();
+        fprintf(out, ", %s", attribute_display_name);
+      }
+      fputc(')', out);
+      fputc('\n', out);
+    }
+  }
+}
+}  // namespace
```
zuyu: Please add one empty line before and after Line 164, respectively. Thanks!

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+              index_it->second.IndexSubBlockType_Name(
+                  index_it->second.sub_block_type()).c_str());
+      fputc(' ', out);
+      fputc('(', out);
+      fprintf(out, "%s", relation->getAttributeById(index_it->second.indexed_attribute_ids(0))
+                             ->getDisplayName().c_str());
+      for (std::size_t i = 1; i < static_cast<std::size_t>(index_it->second.indexed_attribute_ids_size()); ++i) {
+        const char *attribute_display_name = relation->getAttributeById(
+                                                 index_it->second.indexed_attribute_ids(i))
+                                                 ->getDisplayName().c_str();
```
zuyu: Indentation issue: either add four more whitespace here, or refactor the whole statement as following:

```
  const char * attribute_display_name =
      relation->getAttributeById(index_it->second.indexed_attribute_ids(i))
          ->getDisplayName().c_str();
```

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+      }
+      fputc(')', out);
+      fputc('\n', out);
+    }
+  }
+}
+}  // namespace
+void executeCommand(const ParseStatement &statement,
+                                     const CatalogDatabase &catalog_database,
+                                     FILE *out) {
```
zuyu: Indentation issues:

```
void executeCommand(const ParseStatement &statement,
                    const CatalogDatabase &catalog_database,
                    FILE *out) {
```

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+  const ParseCommand &command = static_cast<const ParseCommand &>(statement);
+  const PtrVector<ParseString> *arguments = command.arguments();
+  const std::string command_str = command.command()->value();
```
zuyu: Could we use the reference (`&`) here to avoid a deep copy?

cli/CommandExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+    } else {
+      executeDescribeTable(arguments, catalog_database, out);
+    }
+  } else {
+    THROW_SQL_ERROR_AT(command.command()) << "Invalid Command";
+  }
+}
+}  // namespace cli
+}  // namespace quickstep
+
```
zuyu: Remove this empty line. Thanks!

cli/CommandExecutor.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+class CatalogDatabase;
+class CatalogAttribute;
+class CatalogRelation;
+
+namespace cli {
+/** \addtogroup CLI
+ *  @{
+ */
+// Adding the max column width as 6  as the default initializer
```
zuyu: Please add one empty line between two comments.

cli/CommandExecutor.hpp
```diff
@@ -0,0 +1,64 @@
...
+// as the length of the word Column is 6 characters.
+// This is used while describing the table.
+constexpr int kInitMaxColumnWidth = 6;
+
+static const char kDescribeDatabaseCommand[] = "\\dt";
+static const char kDescribeTableCommand[] = "\\d";
```
zuyu: Prefer to use `constexpr` instead of `const`.

https://google.github.io/styleguide/cppguide.html#Use_of_constexpr

cli/PrintToScreen.hpp
```diff
@@ -43,14 +43,13 @@ class PrintToScreen {
   static void PrintRelation(const CatalogRelation &relation,
                             StorageManager *storage_manager,
                             FILE *out);
-
+  static void printHBar(const std::vector<int> &column_widths,
+                        FILE *out);
```
zuyu: Please add one empty line before and after this function `printHBar`, respectively. Thanks!

Also, we need to modify this file as global functions within namespaces in a separate PR.

cli/QuickstepCli.cpp
```diff
@@ -1,6 +1,6 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
- *   Copyright 2015-2016 Pivotal Software, Inc.
+ *   Copyright 2015 Pivotal Software, Inc.
```
zuyu: Why deleted some copyright? Please reset it.

Also, you should add the university one.

cli/CommandExecutor.hpp
```diff
@@ -0,0 +1,64 @@
...
+  * @brief Executes the command by calling the command handler.
+  *        
+  * @param statement The parsed statement from the cli.
+  * @param catalog_database The catalog information about the current database.
+  * @param out The stream where the output of the command has to be redirected to.      
+*/
+void executeCommand(const ParseStatement &statement,
+                    const CatalogDatabase &catalog_database,
+                    FILE *out);
+/** @} */
```
zuyu: Please add one empty line between Line 58 and 59.

cli/tests/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+#   Copyright 2011-2015 Quickstep Technologies LLC.
+#   Copyright 2015-2016 Pivotal Software, Inc.
```
zuyu: Since this is a new file, you don't need these two copyright from `Quickstep Technologies LLC` and `Pivotal`. Instead, add the university one. Thanks!

cli/tests/CommandExecutorTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include <fstream>
+#include <memory>
+#include <iostream>
```
zuyu: Please sort the header file in the alphabet order. Thanks!

cli/tests/CommandExecutorTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+  std::unique_ptr<quickstep::CommandExecutorTestRunner>
+      test_runner(
+          new quickstep::CommandExecutorTestRunner(argv[3]));
+  test_driver.reset(
+      new quickstep::TextBasedTestDriver(&input_file, test_runner.get()));
+  test_driver->registerOption(
+      quickstep::CommandExecutorTestRunner::kResetOption);
+
+  ::testing::InitGoogleTest(&argc, argv);
+  int success = RUN_ALL_TESTS();
```
zuyu: Mark `success` `const`.

cli/tests/CommandExecutorTestRunner.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace quickstep {
+
+class CatalogRelation;
+
+namespace O = ::quickstep::optimizer;
+namespace P = ::quickstep::optimizer::physical;
+namespace L = ::quickstep::optimizer::logical;
+
+const std::string CommandExecutorTestRunner::kResetOption =
```
zuyu: Please use `char *` instead of `std::string`. Google C++ style explicitly says this type of static variables.

https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

And we should change the same issue in `/query_optimizer/tests/ExecutionGeneratorTestRunner` in another separate PR.

cli/tests/CommandExecutorTestRunner.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  MemStream output_stream;
+  sql_parser_.feedNextBuffer(new std::string(input));
+
+  while (true) {
+    ParseResult result = sql_parser_.getNextStatement();
+
+    O::OptimizerContext optimizer_context(0 /* query_id */,
+                                       test_database_loader_.catalog_database(),
+                                       test_database_loader_.storage_manager());
```
zuyu: Indentation issues for Lines 75 and 76: add two more whitespace for alignments.

cli/tests/CommandExecutorTestRunner.cpp
```diff
@@ -0,0 +1,131 @@
...
+          quickstep::cli::executeCommand(
+              *result.parsed_statement,
+              *(test_database_loader_.catalog_database()),
+              output_stream.file());
+        } else  {
+          QueryHandle query_handle(optimizer_context.query_id());
+          O::LogicalGenerator logical_generator(&optimizer_context);
+          O::PhysicalGenerator physical_generator;
+          O::ExecutionGenerator execution_generator(&optimizer_context,
+                                               &query_handle);
```
zuyu: Indentation issues: please add more whitespace to align with Line 95. Otherwise, merge into the previous line.

cli/tests/CommandExecutorTestRunner.cpp
```diff
@@ -0,0 +1,131 @@
...
+          const P::PhysicalPtr physical_plan =
+            physical_generator.generatePlan(
```
zuyu: Indentation issue for a wrapped line: please add two more whitespace.

cli/tests/CommandExecutorTestRunner.cpp
```diff
@@ -0,0 +1,131 @@
...
+                logical_generator.generatePlan(*result.parsed_statement));
+          execution_generator.generatePlan(physical_plan);
+          foreman_->setQueryPlan(
+            query_handle.getQueryPlanMutable()->getQueryPlanDAGMutable());
```
zuyu: Same indentation issue for a wrapped line.

cli/tests/CommandExecutorTestRunner.hpp
```diff
@@ -0,0 +1,117 @@
...
+  MessageBusImpl bus_;
+  std::unique_ptr<Foreman> foreman_;
+  std::unique_ptr<Worker> worker_;
+
+  std::unique_ptr<WorkerDirectory> workers_;
+
+  DISALLOW_COPY_AND_ASSIGN(CommandExecutorTestRunner);
+};
+
+
```
zuyu: Please remove this empty line.

cli/tests/command_executor/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+#   Copyright 2011-2015 Quickstep Technologies LLC.
+#   Copyright 2015 Pivotal Software, Inc.
```
zuyu: Since this is a new file, please remove unrelated copyrights, and add the university one.

cli/tests/CommandExecutorTestRunner.hpp
```diff
@@ -0,0 +1,117 @@
...
+}  // namespace quickstep
+
+#endif /* QUICKSTEP_CLI_TESTS_COMMAND_EXECUTOR_TEST_RUNNER_HPP_ */
```
zuyu: I know you get this comment from the existing file, but for new file, we should use the following style:

```
#endif  // QUICKSTEP_CLI_TESTS_COMMAND_EXECUTOR_TEST_RUNNER_HPP_
```

Note that two whitespace between `#endif` and `//`.

We should fix the existing files in a separate PR as well.

cli/tests/command_executor/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+         "${CMAKE_CURRENT_SOURCE_DIR}/Dt.test"
+         "${CMAKE_CURRENT_BINARY_DIR}/Dt.test"
+         "${CMAKE_CURRENT_BINARY_DIR}/Dt/")
+
+
+# Create the folders where the unit tests will store their data blocks for the
+# duration of their test.
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/D)
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Dt)
+
```
zuyu: Please remove this empty line.

cli/tests/command_executor/D.test
```diff
@@ -0,0 +1,109 @@
...
+                   col5 CHAR(5));
+CREATE INDEX FOO3_INDEX_1 on FOO3 (col1) USING CSBTREE;
+CREATE TABLE FOO4 (col1 INT,
+                   col2 LONG,
+                   col3 DOUBLE,
+                   col4 FLOAT,
+                   col5 CHAR(5));
+CREATE INDEX FOO4_INDEX_1 on FOO4 (col1, col2) USING CSBTREE; 
+CREATE INDEX FOO4_INDEX_2 on FOO4 (col3, col4) USING CSBTREE; 
+CREATE TABLE AVERYLONGTABLENAMETHATSEEMSTONEVEREND (col1 INT);
```
zuyu: Please fix the following `SQL` style issues. Thanks!
 * All the table names should be in the lower case.
 * `ON` should be capitalized.

cli/tests/command_executor/Dt.test
```diff
@@ -0,0 +1,57 @@
...
+                   col2 LONG,
+                   col3 DOUBLE,
+                   col4 FLOAT,
+                   col5 CHAR(5));
+CREATE TABLE FOO4 (col1 INT,
+                   col2 LONG,
+                   col3 DOUBLE,
+                   col4 FLOAT,
+                   col5 CHAR(5));
+CREATE TABLE AVERYLONGTABLENAMETHATSEEMSTONEVEREND (col1 INT);
```
zuyu: The same `SQL` style issues here.

zuyu:
@rogersjeffreyl I've done one pass. After addressing all the comments, we should merge this PR. Thanks!

cli/tests/command_executor/Dt.test
rogersjeffreyl: adding a note that existing existing ExecutionGenerator  tests do not follow sql styling in naming of tables and have to be refactored in another pr

cli/CommandExecutor.cpp
```diff
@@ -59,21 +59,19 @@ void executeDescribeDatabase(
...
 
     if (relation == nullptr) {
-     THROW_SQL_ERROR_AT(&(arguments->front())) << " Unrecognized relation "  <<table_name_val;
+      THROW_SQL_ERROR_AT(&(arguments->front())) << " Unrecognized relation " << table_name_val;
     }
-    int column_width = static_cast<int>(relation->getName().length());
-    max_column_width =
-        max_column_width < column_width ? column_width : max_column_width;
+    max_column_width = std::max(static_cast<int>(relation->getName().length()),
+                                    max_column_width);
```
zuyu: For your current line wrap fashion, we don't do indentation like this. Instead, align `max_column_width` to `static_cast`, as they are both arguments to `std::max`.

cli/CommandExecutor.cpp
```diff
@@ -117,10 +115,11 @@ void executeDescribeTable(
...
     //   2. Any value of the attribute's Type.
-    int attr_column_width = attr.getDisplayName().length();
-    int type_column_width = attr.getType().getName().length();
-    max_attr_column_width = std::max(max_attr_column_width, attr_column_width);
-    max_type_column_width = std::max(max_type_column_width, type_column_width);
+    max_attr_column_width =
+        std::max(max_attr_column_width,
+            static_cast<int>(attr.getDisplayName().length()));
+    max_type_column_width = std::max(max_type_column_width,
+                                static_cast<int>(attr.getType().getName().length()));
```
zuyu: To wrap this line, there are a few ways, but unfortunately, not using the current one. On the other hand, the way you did for `max_attr_column_width` is ok. You may want to do the same thing to `max_type_column_width`.

Basically, the goal of a line wrapping is for readability so that we could distinguish it with another one-line statement. The main idea to wrap a line is to add four whitespace indentation counting from the first letter of the previous line. If the wrapped line is a part of a function call involving other variables (in this case, two arguments in `std::max`), we typically align the variable with other variables (See my code piece below).

```
  max_type_column_width = std::max(max_type_column_width,
                                   static_cast<int>(attr.getType().getName().length()));
```

This following way is also acceptable, but it is not as good as the previous one, since it does not present in a clear way that the expression in the wrapped line has a strong relationship with the `max_type_column_width` above.
```
  max_type_column_width = std::max(max_type_column_width,
      static_cast<int>(attr.getType().getName().length()));
```

zuyu:
@rogersjeffreyl LGTM. Nice job!

PR #121, opened 2016-04-04
Added Work Order Deserialization.

zuyu:
This PR adds the support to deserialize `WorkOrder`s from proto, used by the distributed version only.

Note that we don't have a separate unit test to cover the serialization, but it has covered in #117.

There will be another follow-up PR to create `WorkOrder` proto in `RelationalOperator`s.


pateljm:
LTGM. @hbdeshmukh?

relational_operators/HashJoinOperator.hpp
```diff
@@ -19,6 +19,7 @@
 #define QUICKSTEP_RELATIONAL_OPERATORS_HASH_JOIN_OPERATOR_HPP_
 
 #include <cstddef>
+#include <utility>
 #include <vector>
 #include <memory>
```
hbdeshmukh: Need to sort the includes.

relational_operators/BuildHashOperator.hpp
```diff
@@ -148,6 +149,30 @@ class BuildHashWorkOrder : public WorkOrder {
...
+   * @param input_relation The relation to build hash table on.
+   * @param join_key_attributes The IDs of equijoin attributes in
+   *        input_relation.
+   * @param any_join_key_attributes_nullable If any attribute is nullable.
+   * @param build_block_id The block id.
+   * @param hash_table The JoinHashTable to use.
+   * @param storage_manager The StorageManager to use.
+   **/
+  BuildHashWorkOrder(const CatalogRelationSchema &input_relation,
+                     std::vector<attribute_id> &&join_key_attributes,
```
hbdeshmukh: I am not 100% sure how this will work. The caller site for this constructor is BuildHashOperator. Let's assume it has a class member called ``join_key_attributes_``. 

When 1st build hash work order is created the constructor will get called (?) as follows :
``
BuildHashWorkOrder(.., std::move(join_key_attributes_), ...<other params>);
``
What's the state of ``join_key_attributes_`` when second build hash work order needs to be created?

zuyu: @hbdeshmukh The caller site for this is `WorkOrderFactory`, which also reconstructs `join_key_attributes` from proto.

hbdeshmukh: Does the ``WorkOrderFactory`` reconstruct the ``join_key_attributes`` for every new ``BuildHashWorkOrder``?

zuyu: Yes, this is the idea of a reconstruction from proto.

zuyu:
@hbdeshmukh I've resolved the comments. Please take another look. Thanks!

relational_operators/WorkOrder.proto
```diff
@@ -0,0 +1,244 @@
...
+    optional uint32 join_hash_table_index = 112;
+  }
+}
+
+message DropTableWorkOrder {
+  extend WorkOrder {
+    // If set, CatalogDatabaseCache will drop the relation.
+    optional int32 relation_id = 128;
+    // Optional, and maybe empty.
+    repeated fixed64 block_ids = 129;
```
hbdeshmukh: What's the difference between ``block_ids`` not present and it being empty?

relational_operators/WorkOrder.proto
```diff
@@ -0,0 +1,244 @@
...
+  }
+}
+
+message FinalizeAggregationWorkOrder {
+  extend WorkOrder {
+    // All required.
+    optional uint32 aggr_state_index = 144;
+    optional int32 insert_destination_index = 145;
```
hbdeshmukh: What's the deal with the integers in this proto file? If tomorrow we have to add another parameter to any one of the WorkOrders, do we have to re-number all the fields below that WorkOrder? That seems too much of a work. Is there an alternative?

relational_operators/WorkOrder.proto
zuyu: I think it's the same.

relational_operators/WorkOrder.proto
zuyu: `insert_destination_index` represents as an `std::int32_t`.

We have reserved 16 numbers for each `WorkOrder` proto. If there is a new field, we need to assign a number for it. We may not need to re-number for a new field, but we may have to do so for a new `WorkOrder` as we keep the `WorkOrderType` sorted.

relational_operators/WorkOrderFactory.cpp
```diff
@@ -0,0 +1,602 @@
...
+             query_context.isValidPredicate(
+                 proto.GetExtension(serialization::UpdateWorkOrder::predicate_index)) &&
+             proto.HasExtension(serialization::UpdateWorkOrder::update_group_index) &&
+             query_context.isValidUpdateGroupId(
+                 proto.GetExtension(serialization::UpdateWorkOrder::update_group_index)) &&
+             proto.HasExtension(serialization::UpdateWorkOrder::operator_index) &&
+             proto.HasExtension(serialization::UpdateWorkOrder::block_id);
+    }
+    default:
+      return false;
```
hbdeshmukh: Should we output a log message suggesting that the type of WorkOrder is unknown?

zuyu: Not necessarily. The following code will print out the whole proto whenever invalid:

https://github.com/pivotalsoftware/quickstep/pull/121/files#diff-2c12385b34f03b5eb61a3f5df00505eeR65

relational_operators/WorkOrderFactory.cpp
```diff
@@ -0,0 +1,602 @@
...
+  }
```
hbdeshmukh: This is one giant file, great job in writing it! Honestly, I have no idea how to review it. I am going to trust you that there are no mistakes. 

Alternatively, can we think of any way in which we can simplify these functions? Perhaps something like moving them to individual WorkOrder files if it makes sense. That way, it might be easier to understand the code.

relational_operators/WorkOrderFactory.hpp
```diff
@@ -0,0 +1,93 @@
...
+   * @param catalog_database The database to resolve relation and attribute
+   *        references in.
+   * @param query_context A pointer to QueryContext.
+   * @param storage_manager The StorageManager to use.
+   * @param shiftboss_client_id The TMB client id of Shiftboss.
+   * @param bus A pointer to the TMB.
+   *
+   * @return A new WorkOrder reconstructed from the supplied Protocol Buffer.
+   **/
+  static WorkOrder* ReconstructFromProto(const serialization::WorkOrder &proto,
```
hbdeshmukh: I have some questions: 
1. Will this function be called by ShiftBoss?
2. Will this function be called once per WorkOrder? 
3. Are there any opportunities for serializing multiple WorkOrders in one proto and subsequently deserializing them all at once? It should be possible for the Foreman to dispatch multiple WorkOrders to one Shiftboss at once. Therefore, we should be able to gain some saving in bulk serializing and deserializing of WorkOrders.

relational_operators/WorkOrderFactory.cpp
zuyu: Thank you for reviewing. You've already done a great job on that!

If we split these code into different `WorkOrder`s, there seems to be a circle dependency between operators and work orders.

As I mentioned in the PR, #117 serves as a big test for several incoming smaller PRs, including this one.

relational_operators/WorkOrderFactory.hpp
zuyu: I have answers:
1. Yes, see [the code](https://github.com/pivotalsoftware/quickstep/pull/117/files#diff-9888f8df64eefe6ae7a4152508eec0f9R95) if you want.
2. Yes for now.
3. I think so, and that is a optimization for later, not in the first prototype of the distributed version. `ForemanDistributed` could group a set of `WorkOrder` proto from the same operator, and then `Shiftboss` could generate them in a batch.

relational_operators/WorkOrderFactory.cpp
hbdeshmukh: What worries me is that the file is difficult to read and it is very easy to make mistakes, which could potentially be dangerous.

relational_operators/WorkOrder.proto
hbdeshmukh: Is the sorting required?

relational_operators/WorkOrderFactory.cpp
zuyu: I was wonder why it is hard to read. I understand that there are too many conditions, but that's because some `WorkOrder` proto has the same number of fields to be serialized.

One way to check the correctness is to compare `WorkOrder` proto and its factory reconstruction method, though.

relational_operators/WorkOrder.proto
zuyu: Yes, sorting for [`WorkOrderType`](https://github.com/pivotalsoftware/quickstep/pull/121/files#diff-b215cf81d42ca18444b83ca67bf323eeR23) is required, as we rarely add a new `WorkOrder`.

hbdeshmukh: Is it a technical requirement or a matter of preference?

zuyu: For a matter of readability, not technical though.

hbdeshmukh: So if someone creates a (fictional) ``AaaWorkOrder``, they will have to renumber 16 to 341 for no good reason. I would strongly oppose the sorting requirement.

zuyu: Yes, you're right.  I've already discuss with @pateljm regarding this issue in the sorted `WorkOrderType`s.

relational_operators/WorkOrderFactory.cpp
hbdeshmukh: I am curious, what is the circular dependency that you mentioned in one of the above comments?

zuyu: Unfortunately, I didn't remember the exact circular dependency, but I did encounter when I implement in half a year ago.

zuyu:
@hbdeshmukh Although this PR is large, but it mainly contains a new `WorkOrderFactory` and `WorkOrder` proto definitions. So I think we should be good to merge this one, as #117 as a whole passes the CI. Finally, we could always refactor some code in any follow-up PR later if necessary.

hbdeshmukh:
Yes, we will merge it but I ought to raise my concerns as a reviewer. 

One potential way for refactoring that I can think of is to move ``ReconstructFromProto`` and ``ProtoIsValid`` from WorkOrderFactory to respective ``WorkOrder`` classes as static methods. It seems less likely that this would cause any circular dependencies. I think it's alright to leave the refactoring as a TODO.

zuyu:
Assume there is no circular dependency, without `WorkOrderFactory`, which do you think should have a big `switch` statement to choose the right  constructor from various `WorkOrder` ones for `serialization::WorkOrder`? It is gonna to be `Shiftboss`, right?

Even worse, we need to implement `ProtoIsValid`, which should also contain a big `switch` for various `WorkOrder::ProtoIsValid`s, right? It has to be `Shiftboss`, the only component to reconstruct `WorkOrder`s from its proto, if without `WorkOrderFactory`.

So do you think the above changes would make the code much better and clearer, or the opposite?

relational_operators/WorkOrder.proto
pateljm: I had mentioned that sorting does improve the readability so like it. Regenerating the numbers does not seem like a big deal at this point as we don't have legacy deployments. But @hbdeshmukh is right at in the future we may not want to strictly sort. For now, I would prefer keeping it sorted, but am ok with whatever the two of you can agree on.

hbdeshmukh:
Alright, looks good. Merging.

relational_operators/WorkOrder.proto
zuyu: @pateljm FYI, unlike `Catalog` proto, regenerating the numbers for `WorkOrder` proto is never a big issue (but a tedious job), as we don't store them permanently, but only serialize on the fly.

@hbdeshmukh Thanks for reviewing and merging.

PR #122, opened 2016-04-04
Factored out a commonly-called method of CatalogRelationSchema.

cramja:
This fixes a code style factoring issue from #115 . We refactored the getMaximumAttributeByteLengths() because the functionality was called multiple times in PackedRowStoreTupleStorageSubblock.

### DIFF
--- a/catalog/CatalogRelationSchema.cpp
+++ b/catalog/CatalogRelationSchema.cpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015-2016 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -25,6 +27,7 @@
 #include "catalog/Catalog.pb.h"
 #include "catalog/CatalogAttribute.hpp"
 #include "catalog/CatalogErrors.hpp"
+#include "catalog/CatalogTypedefs.hpp"
 #include "types/Type.hpp"
 #include "utility/PtrVector.hpp"
 #include "utility/StringUtil.hpp"
@@ -52,6 +55,7 @@ CatalogRelationSchema::CatalogRelationSchema(const serialization::CatalogRelatio
       min_variable_byte_length_(0),
       min_variable_byte_length_excluding_nullable_(0),
       estimated_variable_byte_length_(0),
+      max_byte_lengths_(proto.attributes_size()),
       current_nullable_attribute_index_(-1),
       current_variable_length_attribute_index_(-1) {
   DCHECK(ProtoIsValid(proto))
@@ -150,7 +154,7 @@ attribute_id CatalogRelationSchema::addAttribute(CatalogAttribute *new_attr) {
       }
       estimated_variable_byte_length_ += attr_type.estimateAverageByteLength();
     } else {
-      variable_length_attribute_indices_.push_back(-1);
+      variable_length_attribute_indices_.push_back(kInvalidCatalogId);
 
       fixed_length_attribute_offsets_.resize(new_attr->getID() + 1, fixed_byte_length_);
       fixed_byte_length_ += attr_type.maximumByteLength();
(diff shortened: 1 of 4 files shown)


catalog/CatalogRelationSchema.hpp
```diff
@@ -415,6 +417,16 @@ class CatalogRelationSchema {
     return attr_vec_.end_skip();
   }
 
+  /**
+   * @brief Get a vector of the maximum byte lengths for each attribute.
+   * @details An index in the vector corresponds with the attribute's id.
+   *
+   * @return A vector of the maximum byte lengths for each attribute.
+   */
+  const std::vector<std::size_t>& getMaximumAttributeByteLengths() const {
```
pateljm: inline this function.

zuyu: It is implicitly `inlined` for any function definitions in the header files.

catalog/CatalogRelationSchema.cpp
```diff
@@ -163,10 +166,16 @@ attribute_id CatalogRelationSchema::addAttribute(CatalogAttribute *new_attr) {
...
     } else {
-      nullable_attribute_indices_.push_back(-1);
+      nullable_attribute_indices_.push_back(kInvalidCatalogId);
     }
 
-    return new_attr->getID();
+    const attribute_id attr_id = new_attr->getID();
+    // Ensure that we will not overrun the vector by resizing and initializing
+    // new entries to zero.
+    max_byte_lengths_.resize(attr_id + 1, 0);
```
zuyu: Refactor to use the init list:

```
  ...
  estimated_variable_byte_length_(0),
  max_byte_lengths_(proto.attributes_size()),
  ... {
 ... 
}
```

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -193,6 +181,7 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
       const bool fast_copy =
           (accessor->getImplementationType() ==
               ValueAccessor::Implementation::kCompressedPackedRowStore);
+      const std::size_t attrs_total_size = relation_.getMaximumByteLength();
```
zuyu: This is a bug, right?

It should be the sum of all attributes' max lengths.

storage/PackedRowStoreTupleStorageSubBlock.cpp
```diff
@@ -147,20 +147,8 @@ tuple_id PackedRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *acc
...
-    // object in the following inner loops.
-    std::vector<std::size_t> attrs_max_size;
-    for (CatalogRelationSchema::const_iterator attr_it = relation_.begin();
-         attr_it != relation_.end();
-         ++attr_it) {
-      attrs_max_size.push_back(attr_it->getType().maximumByteLength());
-      attrs_total_size += attrs_max_size.back();
-    }
+    const std::vector<std::size_t> &attrs_max_size =
+        relation_.getMaximumAttributeByteLengths();
```
zuyu: Actually, I think the above two `const` variables could move outside of `ValueAccessor`, like `num_nullable_attrs`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
cramja: Check out the method `addAttribute` in CatalogRelationSchema. Internally it updates this value with
```c++
max_byte_length_ += attr_type.maximumByteLength();
```

catalog/CatalogRelationSchema.cpp
cramja: Good idea

storage/PackedRowStoreTupleStorageSubBlock.cpp
cramja: What's the comparative advantage that this change would make?

catalog/CatalogRelationSchema.cpp
cramja: I tried this change not realizing that there was a constructor in the header file which is also used. That constructor is not given a proto (or any other ahead of time knowledge of the size of the relation) so the `resize` is still necessary.

zuyu: No, we don't need two operations: `resize` and then assign a value.

As an alternative, we just need a `push_back`, like `nullable_attribute_indices_`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
zuyu: I am thinking of moving all `const` variables outside of a loop. In this case, `ValueAccessor`.

storage/PackedRowStoreTupleStorageSubBlock.cpp
zuyu: Oh, I missed that. Sorry.

storage/PackedRowStoreTupleStorageSubBlock.cpp
cramja: @pateljm and I talked - we decided it was okay to declare them inside the function so long as they were not being declared inside a tight loop.

catalog/CatalogRelationSchema.cpp
cramja: I used resize in combination with the element assignment operator `[]` here instead of `push_back` to ensure that when we set a particular index of the `max_byte_lengths_` vector, that it will set the correct mapping between [attribute_id]->attribute length (push_back may not ensure this if attributes were removed).

pateljm:
LTGM. Merging.

catalog/CatalogRelationSchema.cpp
zuyu: In `addAttribute`, we always append the new attribute to the end, although it may change in near future.

PR #124, opened 2016-04-05
Added worker_thread_id in TMB Messages Sent To Foreman.

zuyu:
This PR adds `worker_thread_id` in all TMB messages sent to Foreman. The `thread id-based map` mechanism used here is the one for passing the TMB client id of a worker thread to some `WorkOrder`s and `InsertDestinations`, originally written by @hbdeshmukh.

### DIFF
--- a/storage/InsertDestination.hpp
+++ b/storage/InsertDestination.hpp
@@ -35,6 +35,7 @@
 #include "storage/StorageBlockInfo.hpp"
 #include "storage/StorageBlockLayout.hpp"
 #include "threading/SpinMutex.hpp"
+#include "threading/ThreadIdBasedMap.hpp"
 #include "types/containers/Tuple.hpp"
 #include "utility/Macros.hpp"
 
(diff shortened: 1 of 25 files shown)


storage/InsertDestination.hpp
```diff
@@ -209,10 +210,29 @@ class InsertDestination : public InsertDestinationInterface {
...
+    // 2. Use thread-local storage: each worker saves its id in the local
+    //    storage.
+    // 3. Use a globally accessible map whose key is the caller thread's
+    //    process level id and value is the worker thread index.
+    //
+    // Option 1 involves modifying the signature of several functions across
+    // different modules. Option 2 was difficult to implement given that Apple's
+    // Clang doesn't allow C++11's thread_local keyword. Therefore we chose
+    // option 3.
+    const std::size_t worker_thread_id = WorkerThreadIdMap::Instance()->getValue();
```
hbdeshmukh: This ThreadIDMap is used for retrieving the worker thread ID. After our discussion yesterday, another ThreadIDMap will be needed for retrieving the TMB client ID to send the block filled TMB message to Foreman. I think we can get away with just using the ThreadIDMap for TMB client ID, as it was earlier. I will explain how. 

You require that each message sent to Foreman (both pipelining and work completion) should tell the Foreman who the sender is (i.e. its worker thread ID). 
For the work completion message : As it is sent from the Worker thread, we can just append the worker thread ID in the completion message. 

For the pipelining message: 
1. In the message from the InsertDestination to Foreman, we include the TMB client ID of the worker (as it was earlier). 
2. We modify the WorkerDirectory class to provide a mapping from TMB client ID to worker thread ID. (Right now it provides a mapping from worker thread ID to TMB client ID). 
3. Once the Foreman receives a pipelining message (obviously we can distinguish a pipelining message from work completion message), it uses the WorkerDirectory and gives the TMB client ID of the sender as input and gets the worker thread ID as a result. This worker thread ID can be used for the desired purpose.

This approach will require fewer changes and should be straightforward.

zuyu: Note that `worker_thread_id` refers to `worker_id_` (aka, the index) in `WorkerDirectory`.

Since we have introduced `agent_client_id` for sending messages to Foreman, we don't need the TMB client id of any `Worker` threads.

PR #125, opened 2016-04-05
Refactor AggregationHandle to avoid cyclic library dependency when using HashTable

jianqiao:
This PR refactors `AggregationHandle` into two classes: class  `AggregationHandle` to contain only pure virtual methods, and its helper subclass `AggregationConcreteHandle` to contain method implementations as well as helper methods that are shared by all the aggregation handle subclasses (e.g. `AggregationHandleSum`).

The reason for doing this is to avoid cyclic library dependency that occurs in a later DISTINCT aggregation PR. That is, we hope to add some methods in the base class `AggregationHandle` that are shared by all its subclasses. However, the methods use `HashTable` and `HashTableFactory`, which forms the following dependency cycle:
```
AggregationHandle -> HashTable -> StorageManager -> StorageBlock -> AggregationHandle
```

### DIFF
--- a/expressions/aggregation/AggregationConcreteHandle.hpp
+++ b/expressions/aggregation/AggregationConcreteHandle.hpp
@@ -1,0 +1,276 @@
+/**
+ *   Copyright 2011-2015 Quickstep Technologies LLC.
+ *   Copyright 2015 Pivotal Software, Inc.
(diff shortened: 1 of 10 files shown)


pateljm:
LTGM. Merging.

PR #126, opened 2016-04-05
Add parser and optimizer support for DISTINCT aggregation

jianqiao:
This PR adds parser and optimizer support for the DISTINCT aggregation feature.


parser/ParseBasicExpressions.hpp
```diff
@@ -392,6 +400,13 @@ class ParseFunctionCall : public ParseExpression {
   }
 
   /**
+   * @return Whether this function call contains the DISTINCT operator.
```
zuyu: Q: `distinct` will not become an actual operator, right?

jianqiao: Right. There are two scenarios for `DISTINCT`:

(1) Apply to `SELECT` items, e.g.
```
SELECT DISTINCT x, y + z
FROM a
```
This will be represented as an aggregation with no aggregate function but group by expressions `x`, `y + z`, which is not related to this PR.

(2) Apply inside aggregate function, e.g.
```
SELECT SUM(DISTINCT x)
FROM a
GROUP BY y;
```
This will be treated as the `DISTINCT` case of the sum aggregation. There will be methods inside `AggregationHandleSum` to handle this.

zuyu: So, how about we use `property` instead of `operator` when referring `distinct`?

jianqiao: Yes sounds good, will update it.

query_optimizer/strategy/tests/Aggregate_unittest.cpp
```diff
@@ -78,54 +82,59 @@ class AggregateTest : public StrategyTest {
...
-      logical_table_reference_0_,
-      {alias_0_} /* grouping_expressions */,
-      {sum_alias} /* aggregate_expressions */);
-  expected_physical_ = P::Aggregate::Create(
-      physical_table_reference_0_,
-      {alias_0_} /* grouping_expressions */,
-      {sum_alias} /* aggregate_expressions */,
-      nullptr /* filter_predicate */);
-  EXPECT_CORRECT_PHYSICAL();
+  for (bool is_distinct : {true, false}) {
```
zuyu: I think we could add `const` before `bool`.

query_optimizer/strategy/tests/Aggregate_unittest.cpp
```diff
@@ -78,54 +82,59 @@ class AggregateTest : public StrategyTest {
...
+    const E::AggregateFunctionPtr sum = createSum(add_literal_0_, true, is_distinct);
```
zuyu: Add comments for `true` using this style `/* some_value */`.

query_optimizer/strategy/tests/Aggregate_unittest.cpp
```diff
@@ -78,54 +82,59 @@ class AggregateTest : public StrategyTest {
...
+    const E::AliasPtr sum_alias = createAlias(sum,
+                                              "sum" /* attribute_name */,
+                                              "test_relation" /* relation_name */);
```
zuyu: I know this was from the existing code, but why the relation name is a fixed string, instead of a function call to get the name?

zuyu:
@jianqiao Please add more unit cases in the `/parser/tests/Select.test` regarding `distinct`.

jianqiao:
@zuyu Tests will be provided in a subsequent PR as `Distinct.test` when the full functionality is ready.

query_optimizer/strategy/tests/Aggregate_unittest.cpp
jianqiao: This is the output alias name, and this `sum_alias` will be directly referenced by both `actual_physical_` and `expected_physical_`, so can use any name here.

query_optimizer/strategy/tests/Aggregate_unittest.cpp
jianqiao: Updated.

query_optimizer/strategy/tests/Aggregate_unittest.cpp
jianqiao: Updated.

zuyu:
When we have the full functionality, the expected `Distinct.test` should be in multiple places:
 * `/parser/tests`
 * `/query_optimizer/tests/resolver`
 * `/query_optimizer/tests/logical_generator`
 * `/query_optimizer/tests/physical_generator`
 * `/query_optimizer/tests/execution_generator`

Since we have supports in `parser` and `/query_optimizer/resolver`, we should include the tests for the former two, right?

pateljm:
Stopping the Optimize builds on Travis and merging as these are parser and optimizer changes. We have a long queue of pending jobs in Travis, and we have a full run through Travis before the last set of minor changes. 

LTGM. Merging. Thanks @jianqiao!

jianqiao:
Will add two test files in 5 minutes.

pateljm:
Ok @jianqiao -- reopened the PR. Let me know when you are done.

jianqiao:
@pateljm @zuyu The commits to the `parser_optimizer_distinct_aggregation` branch cannot be reflected in this PR. Will close this one and send a new PR.

PR #127, opened 2016-04-05
Transaction Part 3: LockTable, TransactionTable, LockRequest

hakanmemisoglu:
This PR includes 3 new classes:

- LockTable: Maps each resource to the transaction that have a lock on that resource.

- TransactionTable: Inverted version of LockTable.

- LockRequest: Input for LockManager which will be added later.

### DIFF
--- a/transaction/LockRequest.hpp
+++ b/transaction/LockRequest.hpp
@@ -33,8 +33,8 @@ namespace transaction {
  * @brief Enum class for representing request types.
  **/
 enum class RequestType {
-  kACQUIRE_LOCK = 0,
-  kRELEASE_LOCKS
+  kAcquireLock = 0,
+  kReleaseLocks,
 };
 
 /**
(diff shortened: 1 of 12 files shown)


transaction/LockRequest.hpp
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * @brief Enum class for representing request types.
+ **/
+enum class RequestType {
+  kAcquireLock = 0,
+  kReleaseLocks,
+};
+
+/**
+ * @brief Class for encapsulate lock request put into
```
pateljm: for -> to

transaction/LockTable.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+namespace quickstep {
+namespace transaction {
+
+LockTableResult
+LockTable::putLock(const transaction_id tid,
+                   const ResourceId &rid,
+                   const AccessMode access_mode) {
+  // TODO(hakan): Lock upgrade is not supported.
+  lock_list_pair &lock_list_pair = internal_map_[rid];
```
pateljm: Is this function thread safe? What happens if two threads call this function at the same time. Race condition below?

transaction/LockRequest.hpp
hakanmemisoglu: I will change the documentation. Thanks Jignesh.

transaction/LockTable.cpp
hakanmemisoglu: I will change the documentation to inform about my assumptions. I will add some preconditions and postconditions. I will also add some TODOs to improve the reliability of the functions in case of that they are used by other people.

pateljm:
Sounds good. As discussed, we should have a mutex protect the calls to the lock table. But, we can do that in a separate PR. Merging.

Issue #73, opened 2016-03-08
Integration Test for quickstep_cli_shell


zuyu:
We should have the integration test for `quickstep_cli_shell` in the CI test pipeline.

pateljm:
@zuyu: Can you expand on this? Do you mean an end-to-end test? I'm working on one.

zuyu:
@pateljm I mean, in the `Travis` CI we run `quickstep_cli_shell` with some queries, and then compare the query results with the expected ones.

pateljm:
Good point. I am working on creating a few queries to do just that. Haven't finished it, but will try and do that soon.

PR #130, opened 2016-04-06
Drop catalog initialization related files

pateljm:
Housecleaning: These files are no longer needed since PR #116.


zuyu:
@pateljm There is a bug in the cli shell, but currently it cannot be found by the CI. And this is the reason why we need #73.

Please set [the default database init flag](https://github.com/pivotalsoftware/quickstep/blob/drop_catalog_init_files/cli/QuickstepCli.cpp#L144) to `true`, otherwise the cli crashes with the following log:

```
F0406 03:31:22.377707 1940639744 QuickstepCli.cpp:241] FATAL ERROR DURING STARTUP: UnableToReadCatalog: Unable to read catalog from file qsstor/catalog.pb.bin
*** Check failure stack trace: ***
    @        0x10ceb397a
    @        0x10ceb293b
    @        0x10ceb362a
    @        0x10ceba468
    @        0x10ceb3db5
    @        0x10ce585ea
Abort trap: 6
```

pateljm:
@zuyu All the checks in Travis went through. 

Where is the bug? Are you suggesting that we should initialize a database by default. That is really odd default behavior.

zuyu:
@pateljm Just run `./quickstep_cli_shell`, and you'll know what I was talking about.

pateljm:
@zuyu I know what you are talking about! Read my comment carefully -- it is odd to initialize a database by default!

zuyu:
As an alternative, check whether the catalog exists. If not, create one without crashing.

pateljm:
What you are saying does not make sense. If the qsstor/catalog.pb.bin exists, we already check it.

zuyu:
I am talking about the opposite. If the default catalog proto does not exist, we create one without crashing the system. It is ok that `-initialize_db` is `false` by default.

I just don't want to see that when a user first runs the system, it crashes due to no catalog.

pateljm:
I think that is ok.

zuyu:
As a part of this PR, we should remove this `TODO` comment:

https://github.com/pivotalsoftware/quickstep/blob/drop_catalog_init_files/cli/QuickstepCli.cpp#L197

pateljm:
I added an message to provide hint to the user when there is no database and that they should initialize one. We may come back and revisit it later, but this is ok for now.

Good point about removing the TODO. Will do that now.

zuyu:
I did not see any follow-up commit, until you told me in a comment later. It needs to fresh the page.

pateljm:
No worries. Thanks for your input on this @zuyu. 

Cheers,
Jignesh

zuyu:
FYI, the CI test does not cover changes to the CLI shell. We need to add a test to do so.

pateljm:
Yup -- need to add end-to-end tests. Its on a TODO list and what lead me to add the MODULO operator.

PR #132, opened 2016-04-06
Adds EXTRACT function for extracting fields from DATETIME

jianqiao:
This PR adds parser and resolver support for the `EXTRACT` function.

Since the backend `DateExtractOperation` and `DateExtractUncheckedOperator` are already there. So the `EXTRACT` function feature is complete. SQL tests are also added.


query_optimizer/tests/execution_generator/Select.test
```diff
@@ -632,6 +632,48 @@ ORDER BY group_col1,
 +--------------------+--------------------+--------------------+
 ==
 
+
+CREATE TABLE dates (value DATETIME);
```
zuyu: If we change one test file in `query_optimizer`, we need to update other three test files under different directories. In this case, they are

 * `resolver`
 * `logical_generator`
 * `physical_generator`

parser/ParseBasicExpressions.hpp
```diff
@@ -448,6 +448,69 @@ class ParseFunctionCall : public ParseExpression {
...
+   *
+   * @param line_number The line number of the token "extract" in the statement.
+   * @param column_number The column number of the token "extract in the statement.
+   * @param extract_field The field to extract.
+   * @param source_expression The expression to extract a field from.
+   */
+  ParseExtractFunction(const int line_number,
+                       const int column_number,
+                       ParseString *extract_field,
+                       ParseExpression *date_expression)
```
zuyu: Could we mark two pointers above `const`?

parser/ParseBasicExpressions.hpp
```diff
@@ -448,6 +448,69 @@ class ParseFunctionCall : public ParseExpression {
...
+      std::vector<std::string> *inline_field_names,
+      std::vector<std::string> *inline_field_values,
+      std::vector<std::string> *non_container_child_field_names,
+      std::vector<const ParseTreeNode*> *non_container_child_fields,
+      std::vector<std::string> *container_child_field_names,
+      std::vector<std::vector<const ParseTreeNode*>> *container_child_fields) const override;
+
+ private:
+  std::unique_ptr<ParseString> extract_field_;
+  std::unique_ptr<ParseExpression> date_expression_;
```
zuyu: Mark two pointers `const`.

parser/ParseExpression.hpp
```diff
@@ -37,10 +39,11 @@ class ParseExpression : public ParseTreeNode {
   enum ExpressionType {
     kAttribute,
     kBinaryExpression,
+    kExtract,
     kFunctionCall,
     kScalarLiteral,
     kSubqueryExpression,
-    kUnaryExpression,
+    kUnaryExpression
```
zuyu: Reset this change. `c++ 11` allows it.

zuyu:
@jianqiao In the title, replace `operator` with `function`.

query_optimizer/tests/execution_generator/Select.test
jianqiao: I updated a test for `resolver`. For `logical_generator` and `physical_generator`, I think the purpose  is to check query plans. But `EXTRACT` is an expression which is not changed during logical planning and physical planning. So the resolver test should be sufficient.

parser/ParseBasicExpressions.hpp
jianqiao: We can do that but all the other `ParseTreeNode` subclasses did not use `const` for `ParseTreeNode` objects. Perhaps a separate PR to update all if that is a concern?

parser/ParseBasicExpressions.hpp
jianqiao: Same concern as above.

parser/ParseExpression.hpp
jianqiao: Updated.

pateljm:
LTGM. Merging. I see the Travis error, but that is my bad as the line is over 120 characters and I'll fix it next in a separate PR. 

Thanks @jianqiao.

PR #137, opened 2016-04-06
Removed a reference usage in SelectWorkOrder.

zuyu:
It is particular useful in the distributed case, where a `WorkOrder` cannot access to its operator.


query_optimizer/ExecutionGenerator.cpp
```diff
@@ -477,14 +477,14 @@ void ExecutionGenerator::convertSelection(
...
+  std::vector<attribute_id> attributes;
   SelectOperator *op
-      = convertSimpleProjection(project_expressions_group_index, attributes.get())
+      = convertSimpleProjection(project_expressions_group_index, &attributes)
         ? new SelectOperator(*input_relation_info->relation,
                              *output_relation,
                              insert_destination_index,
                              execution_predicate_index,
-                             attributes.release(),
+                             move(attributes),
```
hbdeshmukh: Are there any other operators for which similar changes are needed?

zuyu: If you are talking about the shared reference between a operator and a work order, I think no. In fact, `SelectOperator` is the only one I need to change to run queries in the distributed version.

hbdeshmukh:
Looks good, merging.

PR #138, opened 2016-04-06
DO NOT MERGE: Explore faster GCC build times in Travis

pateljm:
GCC builds with `joinwithbinaryexpressions` turned on takes a very long time since the parallelism is set to 1. Explore if we can improve on this by moving up the gcc version that is more robust to template generation.


pateljm:
Thanks @hakanmemisoglu for PR#161. Closing this.

Issue #110, opened 2016-03-31
Support Schema Changes.


zuyu:
Adding support for schema changes include parser support and passing the [Alter.test](https://github.com/pivotalsoftware/quickstep/blob/master/parser/tests/Alter.test), and internal changes to `CatalogRelationSchema`.

We also need to copy `Alter.test` in other `/query_optimizer/tests/` places including `resolver`, `logical_generator`, `physical_generator`, and `execution_generator`.

PR #141, opened 2016-04-06
Avoided update CatalogDatabaseCache on the fly.

zuyu:
In distributed version, it is not safe to update cached `CatalogRelationSchema` on the fly, while some storage blocks still have reference to the existing schema.

So, disable updates in the cache for now.

### DIFF
--- a/catalog/CatalogDatabaseCache.cpp
+++ b/catalog/CatalogDatabaseCache.cpp
@@ -19,6 +19,7 @@
 #include <memory>
 #include <unordered_map>
 #include <utility>
+#include <vector>
 
 #include "catalog/Catalog.pb.h"
 #include "catalog/CatalogRelationSchema.hpp"
@@ -30,6 +31,7 @@
 
 using std::make_unique;
 using std::move;
+using std::vector;
 
 namespace quickstep {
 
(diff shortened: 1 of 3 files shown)


pateljm:
@zuyu: So what happens when the catalog changes?

catalog/CatalogDatabaseCache.cpp
```diff
@@ -61,15 +61,17 @@ void CatalogDatabaseCache::update(const serialization::CatalogDatabase &proto) {
 
   SpinSharedMutexExclusiveLock<false> lock(relations_mutex_);
   for (int i = 0; i < proto.relations_size(); ++i) {
-    auto relation_schema = make_unique<const CatalogRelationSchema>(proto.relations(i));
-    const relation_id rel_id = relation_schema->getID();
+    const serialization::CatalogRelationSchema &proto_relation = proto.relations(i);
+    const relation_id rel_id = proto_relation.relation_id();
```
pateljm: Just checking if there is a reason to use relation_id instead of relationID. Not a big deal, but just wanted to know the reason. Thanks!

hbdeshmukh: I am not sure how much performance intensive this function will be.  However, If performance is a concern we can refactor it as follows:

```
// No protection needed for this -
vector<rel_id> relation_ids;
// populate relation_ids from proto.

vector<relation_id> need_to_add_relation_id

// In a shared lock mode
for each rel_id in relation_ids {
  if (rel_id not found in rel_map_) {
    need_to_add_relation_id.empalce(rel_id)
  }
}
if (!need_to_add_relation_id.empty()) {
// Exclusive lock
// Add the necessary relation IDs to rel_map_.
}
```

hbdeshmukh:
I have one refactoring suggestion for better performance and echoing @pateljm's question on catalog changes.

zuyu:
@pateljm @hbdeshmukh As noted in #110, we does not support schema changes.

When we support schema changes, we may want to have a version number in `CatalogRelationSchema` to indicate whether we need to update `CatalogDatabaseCache`.

If the changed schema has been cached, a naive way is to flush all the blocks belonged to the schema to the disk, before we reset the schema. There are certain scheduling optimization could be done as well, i.e., postpone the schema changes unless no queries are using it.

catalog/CatalogDatabaseCache.cpp
zuyu: @pateljm Do you mean `relation_id` in `proto_relation.relation_id()`?

If so, it is the API generated by protobuf. Thanks!

zuyu: @hbdeshmukh This function is called once per query, before Foreman sends the first `WorkOrder` for the query.

You're right that we need to mutex protection once we have multiple query support (i.e., once `QueryManager` has merged), and your proposed refactor looks good. So, I prefer to do so in another PR. Thanks!

hbdeshmukh: I didn't actually think of concurrent queries for the refactoring, but thanks for pointing to that. There are three reasons:

1. You don't need mutex protection to read proto. Therefore line 64 and 65 can be moved out of any protected scope. 
2. Finding whether a rel_id is in rel_map_ only requires shared lock. Therefore line 67 - 71 can be moved to a shared lock scope. 
3. Finally when you want to modify rel_map_, you should grab an exclusive lock. 

I think it is fine to delay the refactoring until we have concurrent query support, but you can add a TODO for that.

zuyu: @hbdeshmukh Ah, now I got that you were talking about the big `ExclusiveLock`. I wish you could comment on Line 62 regarding this concern.

Actually, I could refactor it in this PR. Thanks!

zuyu:
@hbdeshmukh I've refactored using fine-grained locks to update `CatalogRelationSchema`s. Please take a look. Thanks!

hbdeshmukh:
Looks good, merging.

PR #143, opened 2016-04-07
Change to gold linker.

### LINKED ISSUE #155: Introduce gold linker for Travis to speedup overall linking phase
Gold linker is newer linker and provides faster linking phase for building the project.

hakanmemisoglu:
This PR makes Travis to use gold linker which is faster than original linker. Resolves #155

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ install:
       export CC="clang-3.7";
       export CXX="clang++-3.7";
     fi
+  - export CLINKER=`which gold`
   - export DEBUG_FLAGS="-g0";
   - export RELEASE_FLAGS="-O0 -DNDEBUG";
   - export LINKER_FLAGS="-s"
@@ -39,6 +40,7 @@ install:
 before_script:
   - $CC --version
   - $CXX --version
+  - $CLINKER --version
   - (cd build &&
      cmake -D CMAKE_BUILD_TYPE=$BUILD_TYPE
            -D CMAKE_C_FLAGS_DEBUG="$DEBUG_FLAGS"
@@ -48,6 +50,7 @@ before_script:
            -D CMAKE_EXE_LINKER_FLAGS="$LINKER_FLAGS"
            -D CMAKE_C_COMPILER=$CC
            -D CMAKE_CXX_COMPILER=$CXX
+           -D CMAKE_LINKER=$CLINKER
            -D USE_TCMALLOC=0
            -D VECTOR_COPY_ELISION_LEVEL=$VECTOR_COPY_ELISION_LEVEL ..)
 
@@ -77,6 +80,7 @@ addons:
       - gcc-5
       - g++-5
       - clang-3.7
+      - binutils-gold
       - libprotobuf-dev
       - protobuf-compiler
       - python-networkx


zuyu:
@hakanmemisoglu This is really cool. It saves 10+ minutes (~`30%` execution time) using `LLVM`.

pateljm:
Really neat!

hakanmemisoglu:
@pateljm, latest numbers are:

    Compiler: clang 
    BUILD_TYPE=Debug VECTOR_COPY_ELISION_LEVEL=none 18 min 25 sec
    BUILD_TYPE=Release VECTOR_COPY_ELISION_LEVEL=none 18 min 27 sec
    BUILD_TYPE=Debug VECTOR_COPY_ELISION_LEVEL=joinwithbinaryexpressions 21 min 23 sec
    BUILD_TYPE=Release VECTOR_COPY_ELISION_LEVEL=joinwithbinaryexpressions 27 min 56 sec
    
    Compiler: gcc
    BUILD_TYPE=Debug VECTOR_COPY_ELISION_LEVEL=none 17 min 44 sec
    BUILD_TYPE=Release VECTOR_COPY_ELISION_LEVEL=none 14 min 46 sec
    BUILD_TYPE=Debug VECTOR_COPY_ELISION_LEVEL=joinwithbinaryexpressions 47 min 26 sec
    BUILD_TYPE=Release VECTOR_COPY_ELISION_LEVEL=joinwithbinaryexpressions 24 min 43 sec

pateljm:
This is neat! Is there a minimal set of changes that would make this work that does not touch the third_party files?

hakanmemisoglu:
I think we should update glog library to fix couple of issues. I do not think we can do it without touching glog library.

pateljm:
Good idea about fixing the glog library. Let's discuss tomorrow as a group (and come back and document this discussion here). 

Also, could we just change the capture statement (and not all the formatting).We can have a separate PR for re-formatting. 

Thanks! This is really neat if we can get this working.

zuyu:
@hakanmemisoglu I think we could use as many as pre-built packages as possible to reduce the build time. For example, `gtest`.

See below for how to do so in `travis`:

https://github.com/easylogging/easyloggingpp/blob/master/.travis.yml#L6

hakanmemisoglu:
@zuyu The 3rd party libraries compilation times are not bottleneck in the build process. I am trying to upgrade to newest google test in #153 and related PR #156 to reduce compilation errors produced by gcc-5.3. The repos from Ubuntu 12.04 are generally old, (for example the package you mentioned is version 1.6.0,  and it seems it is  5-6 years old when I check [https://github.com/google/googletest/tree/release-1.6.0](url). To sum up, gcc-5.3 is more compilant to C++11 standard, and C++11 itself is more strict about certain warnings. Since we convert any warning to an error, it forces Travis to fail. This PR addresses some of the warnings. If this PR does not harm anybody's workflow, it should be merged into main.

zuyu:
@hakanmemisoglu Let me explain your logic flow here. We are trying to use gcc 5.3, but encountered some warnings (as errors for us) due to `gtest`, right?

If I understand correctly, using a pre-built `gtest` binary in `travis` should not generate any warnings, right?

Also, have you tried `gtest` 1.7, released on Sep 2013? Although it is still quite old.

hakanmemisoglu:
@zuyu Problem is with the macros, and macros are exposed with header files and these macros are compiled with our source code. Compilation of google test (in the beginning of build phase) is always OK, I did not encounter any error with that. 
Right now, I am trying to fix the error coming from the header files, By the way, whatever we have installed precompiled packages, or source code from github, it will give errors.

pateljm:
Nice work @hakanmemisoglu Imagine all the energy that we are now going to save. Thank you! 

LTGM. Merging.

PR #144, opened 2016-04-07
DO NOT MERGE: Try to debug the ref count problem on Travis CI.

jianqiao:
Print stacktrace information when `StorageBlockBase::ref()` and `StorageBlockBase::unref()` are called.


pateljm:
@jianqiao Thank you so much for starting this!

PR #146, opened 2016-04-08
Automatically set the buffer pool size based on installed system memory: Part II (final)

pateljm:
Use the new function to detect the installed memory. When we can detect successfully, use it to automatically set the buffer pool size. For systems with more than 32GB  of memory grab 90% for the buffer pool, otherwise only grab 80%. Some magic constants in `storage/StorageManager.cpp,` which are isolated just there. At some point when we have more of these, we may need to refactor all magic constants in one place.


cli/QuickstepCli.cpp
```diff
@@ -164,7 +164,9 @@ int main(int argc, char* argv[]) {
                                          : 1);
 
   if (real_num_workers > 0) {
-    printf("Starting Quickstep with %d worker thread(s)\n", real_num_workers);
+    printf("Starting Quickstep with %d worker thread(s) and a %.2f GB buffer pool\n",
+           real_num_workers,
+           (static_cast<double>(quickstep::FLAGS_buffer_pool_slots) * quickstep::kSlotSizeBytes)/(1<<30));
```
zuyu: Suggest to precompute the value:

```
  const double buffer_pool_size_in_gata_bytes =
      static_cast<double>(quickstep::FLAGS_buffer_pool_slots) * quickstep::kSlotSizeBytes / (1 << 30);
  printf("... %.2f ...", ..., buffer_pool_size_in_gata_bytes);
```

storage/StorageManager.cpp
```diff
@@ -96,23 +97,63 @@ DEFINE_int32(block_domain, 1,
...
+ *         and never returns false.
+ **/
+static bool SetOrValidateBufferPoolSlots(const char *flagname,
+                                         std::uint64_t value) {
+  // Magic constants which may change in over time. The key idea is to leave
+  //   more memory (20%) on "small" systems for other processes.
+  const std::uint64_t a_giga_byte = (1<<30);
+  const std::uint64_t threshold_for_a_large_memory_system = 32;  // in GB.
+  const std::uint64_t percentage_to_grab_for_small_systems = 80;
+  const std::uint64_t percentage_to_grab_for_large_systems = 90;
```
zuyu: @pateljm I've read your comments, but since `a_giga_byte` is also used in the CLI shell to compute the buffer pool size, we should define it in `/storage/StorageConstants.hpp`. Also, we need to change its name to `kAGigaByte`, according to [the C++ style for the constants](https://google.github.io/styleguide/cppguide.html#Constant_Names).

```
  constexpr std::uint64_t kAGigaByte = (1 << 30);
```

Note that we use `constexpr` to indicate a true constant in the compilation time based on [the C++ style](https://google.github.io/styleguide/cppguide.html#Use_of_constexpr).

So do these three:
 * `constexpr std::uint64_t kThresholdForALargeMemorySystem = 32;`
 * `kPercentageToGrabForSmallMemorySystems`
 * `kPercentageToGrabForLargeMemorySystems`

storage/StorageManager.cpp
```diff
@@ -96,23 +97,63 @@ DEFINE_int32(block_domain, 1,
...
+      }
+      return true;
+    }
+    // Could not calculate the installed memory. Use a default value of 1k slots.
+    LOG(INFO) << "Unable to determine an appropriate buffer pool size. "
+              << "Using a default value of 2GB.\n";
+    FLAGS_buffer_pool_slots = 1024;
     return true;
+  } else {
+    return true;  // User supplied value is > 0 and will be used.
```
zuyu: How about the following refactor, which saves one level indentation:

```
  if (value != 0) {
    return true;  // ...
  }

  // Code to automatically detect memory size.
  ...
  return true;
```

cli/QuickstepCli.cpp
pateljm: Why?

storage/StorageManager.cpp
pateljm: Good ideas.

pateljm: Will use const rather than constexpr as that is what is used in StorageConstants.hpp. We can have a separate PR to change all. But, in this case it won't matter.

cli/QuickstepCli.cpp
zuyu: This ups to you, but it looks more cleaner to me.

storage/StorageManager.cpp
zuyu: OK.

And FYI, `constexpr` is introduced in `c++`, seems before we wrote the code in `StorageConstants.hpp`.

storage/StorageManager.cpp
pateljm: Good idea!

storage/StorageConstants.hpp
```diff
@@ -34,6 +34,24 @@ namespace quickstep {
 // different large page sizes.
 const std::size_t kSlotSizeBytes = 0x200000;
 
+// A GigaByte.
+const std::uint64_t kAGigaByte = (1 << 30);
```
zuyu: Please include `cstdint`, as you use `std::uint64_t`.

storage/StorageManager.cpp
```diff
@@ -96,23 +97,60 @@ DEFINE_int32(block_domain, 1,
...
+ * @param flagname The name of the buffer pool flag. 
+ * @param value The value of this flag from the command line, or default (0)
+ * @return True if the value was set to a legimate value, false otherwise. 
+ *         Currently this method aims to always find some legitimate value,
+ *         and never returns false.
+ **/
+static bool SetOrValidateBufferPoolSlots(const char *flagname,
+                                         std::uint64_t value) {
+  if (value != 0) {
+    // TODO(jmp): Check if this value  is safe and warn the user if it is not.
```
zuyu: Remove the extra whitespace in `value  is`.

storage/StorageConstants.hpp
pateljm: done

storage/StorageManager.cpp
```diff
@@ -96,23 +97,60 @@ DEFINE_int32(block_domain, 1,
...
+    //   based on whether the system is large or small.
+    if (total_memory/kAGigaByte < kLargeMemorySystemThresholdInGB) {
+      // This is a "small" system. Leave a litte more memory for others.
+      FLAGS_buffer_pool_slots
+        = (total_memory*kPercentageToGrabForSmallSystems)/(kSlotSizeBytes*100);
+    } else {
+      // This is a "large" system. Grab nearly all of the installed memory.
+      FLAGS_buffer_pool_slots
+        = (total_memory*kPercentageToGrabForLargeSystems)/(kSlotSizeBytes*100);
+    }
```
zuyu: Please add two more whitespace indentations for two `FLAGS_buffer_pool_slots` assignments above:

```
  FLAGS_buffer_pool_slots
      = (total_memory* kPercentageToGrabForLargeSystems)/(kSlotSizeBytes*100);
```

Alternatively, we could refactor the `if-else` statement as the following:

```
  FLAGS_buffer_pool_slots =
      ((total_memory/kAGigaByte < kLargeMemorySystemThresholdInGB)
          ? kPercentageToGrabForSmallSystems
          : kPercentageToGrabForLargeSystems) *
       total_memory /
       (kSlotSizeBytes*100);
```

zuyu:
LGTM, except for the minor indentation issue for `FLAGS_buffer_pool_slots`.

storage/StorageManager.cpp
pateljm: Why 2 more whitespaces?

zuyu: Code style: totally four whitespace indentations when wrapping an expression into a new line.

storage/StorageConstants.hpp
```diff
@@ -19,6 +19,7 @@
 #define QUICKSTEP_STORAGE_STORAGE_CONSTANTS_HPP_
 
 #include <cstddef>
+#include <cstint>
```
zuyu: Typo: `cstint` -> `cstdint`.

storage/StorageManager.cpp
pateljm: Hmm.. Can you point to where it says that in the style guide.

zuyu: Mutiple place states the subsequent line for 4 whitespace indentation:

https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions
https://google.github.io/styleguide/cppguide.html#Constructor_Initializer_Lists

pateljm: None of those are what this line is about.

zuyu: Well, this is our own code convention, then.

zuyu: An example from `cpplint` unit test:

https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint_unittest.py#L3635

pateljm: Streching now are we ...

PR #149, opened 2016-04-10
Parser support for EXISTS subquery

jianqiao:
This PR implements parser support for the EXISTS subquery.

The `ParsePredicateExists` class is created as a new file. Because putting it inside `ParsePredicate.*` will introduce cyclic library dependency:
```
ParsePredicate -> ParseSubqueryExpression -> ParseSelect -> ParsePredicate
```


zuyu:
LGTM!

PR #150, opened 2016-04-11
Parser and resolver support for CASE expressions

jianqiao:
This PR adds parser and resolver support for `CASE` expressions. Meanwhile the backend support `ScalarCaseExpression` is already there. So the `CASE` expression feature is fully functional.


pateljm:
LTGM. Merging.

PR #151, opened 2016-04-11
Resolver support for the EXISTS subquery

jianqiao:
This PR adds resolver support for the `EXISTS` subquery.

Main changes include:
1. Adds `class Exists` and `class SubqueryExpression` in `query_optimizer/expressions`.
2. Updates `NameResolver` to support resolving attributes from outer scopes (i.e. in nested queries, the inner query may involve attributes from the outer query's table references).

### DIFF
--- a/query_optimizer/expressions/Exists.hpp
+++ b/query_optimizer/expressions/Exists.hpp
@@ -1,0 +1,118 @@
+/**
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#ifndef QUICKSTEP_QUERY_OPTIMIZER_EXPRESSIONS_EXISTS_HPP_
+#define QUICKSTEP_QUERY_OPTIMIZER_EXPRESSIONS_EXISTS_HPP_
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "query_optimizer/OptimizerTree.hpp"
+#include "query_optimizer/expressions/AttributeReference.hpp"
+#include "query_optimizer/expressions/Expression.hpp"
+#include "query_optimizer/expressions/ExpressionType.hpp"
+#include "query_optimizer/expressions/Predicate.hpp"
+#include "query_optimizer/expressions/SubqueryExpression.hpp"
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+namespace optimizer {
+namespace expressions {
+
+/** \addtogroup OptimizerExpressions
+ *  @{
+ */
+
+class Exists;
+typedef std::shared_ptr<const Exists> ExistsPtr;
+
+/**
+ * @brief EXISTS predicate expression.
+ */
+class Exists : public Predicate {
+ public:
+  ExpressionType getExpressionType() const override {
+    return ExpressionType::kExists;
+  }
+
+  std::string getName() const override {
+    return "Exists";
+  }
+
+  bool isConstant() const override {
+    return false;
+  }
+
+  /**
+   * @return The subquery expression for this EXISTS predicate.
+   */
+  const SubqueryExpressionPtr& exists_subquery() const {
+    return exists_subquery_;
+  }
+
+  ExpressionPtr copyWithNewChildren(
+      const std::vector<ExpressionPtr> &new_children) const override {
+    DCHECK_EQ(new_children.size(), 1u);
+    return Create(std::static_pointer_cast<const SubqueryExpression>(new_children[0]));
+  }
+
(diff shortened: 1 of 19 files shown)


pateljm:
LTGM; merging. This push was simply syncing the gen files, so not waiting for Travis as that came back green just before this last push.

query_optimizer/expressions/Exists.hpp
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @return The subquery expression for this EXISTS predicate.
+   */
+  const SubqueryExpressionPtr& exists_subquery() const {
+    return exists_subquery_;
+  }
+
+  ExpressionPtr copyWithNewChildren(
+      const std::vector<ExpressionPtr> &new_children) const override {
+    DCHECK_EQ(new_children.size(), 1u);
```
zuyu: Sorry for the late comment, but as a convention for `CHECK_EQ`, we put the expected value in the left hand side, while the checked one in the right.

For this case, it should look like `DCHECK_EQ(1u, new_children.size());`.

query_optimizer/expressions/ExpressionUtil.cpp
```diff
@@ -29,7 +33,20 @@ AttributeReferencePtr ToRef(const NamedExpressionPtr &expression) {
...
-                                    expression->getValueType());
+                                    expression->getValueType(),
+                                    AttributeReferenceScope::kLocal);
+}
+
+std::vector<AttributeReferencePtr> GetAttributeReferencesWithinScope(
+    const std::vector<AttributeReferencePtr> &attributes,
+    const AttributeReferenceScope scope) {
+  std::vector<AttributeReferencePtr> filtered_attributes;
+  for (const auto& attr_it : attributes) {
```
zuyu: If we apply the reference sigil standard here, it would look like:

`for (const auto &attr_it : attributes) {`.

query_optimizer/expressions/SubqueryExpression.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "query_optimizer/expressions/AttributeReference.hpp"
+
+#include "glog/logging.h"
+
+namespace quickstep {
+namespace optimizer {
+namespace expressions {
+
+::quickstep::Scalar* SubqueryExpression::concretize(
+    const std::unordered_map<ExprId, const CatalogAttribute*>& substitution_map) const {
```
zuyu: What do you think if we apply the reference sigil standard here?

`const std::unordered_map<ExprId, const CatalogAttribute*> &substitution_map`.

query_optimizer/expressions/SubqueryExpression.hpp
```diff
@@ -0,0 +1,126 @@
...
+  std::vector<AttributeReferencePtr> getReferencedAttributes() const override;
+
+  ExpressionPtr copyWithNewChildren(
+      const std::vector<ExpressionPtr> &new_children) const override {
+    DCHECK(new_children.empty());
+    return Create(subquery_);
+  }
+
+  ::quickstep::Scalar* concretize(
+      const std::unordered_map<ExprId, const CatalogAttribute*>& substitution_map) const override;
```
zuyu: What do you think if we apply the reference sigil standard here?

`const std::unordered_map<ExprId, const CatalogAttribute*> &substitution_map) ...`.

query_optimizer/expressions/SubqueryExpression.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * @brief Creates a subquery expression.
+   * @note This expression can only be used in a logical plan.
+   *
+   * @param subquery The logical subquery node.
+   * @return An immutable SubqueryExpression.
+   */
+  static SubqueryExpressionPtr Create(const logical::LogicalPtr& subquery) {
```
zuyu: What do you think if we apply the reference sigil standard here?

`const logical::LogicalPtr &subquery`.

query_optimizer/expressions/SubqueryExpression.hpp
```diff
@@ -0,0 +1,126 @@
...
+  void getFieldStringItems(
+      std::vector<std::string> *inline_field_names,
+      std::vector<std::string> *inline_field_values,
+      std::vector<std::string> *non_container_child_field_names,
+      std::vector<OptimizerTreeBaseNodePtr> *non_container_child_fields,
+      std::vector<std::string> *container_child_field_names,
+      std::vector<std::vector<OptimizerTreeBaseNodePtr>> *container_child_fields) const override;
+
+ private:
+  explicit SubqueryExpression(const logical::LogicalPtr& subquery)
```
zuyu: What do you think if we apply the reference sigil standard here?

`const logical::LogicalPtr &subquery`.

query_optimizer/expressions/SubqueryExpression.hpp
```diff
@@ -0,0 +1,126 @@
...
+      : subquery_(subquery),
+        output_attribute_(subquery->getOutputAttributes()[0]) {
+    DCHECK(!subquery->getOutputAttributes().empty());
+  }
+
+  logical::LogicalPtr subquery_;
```
zuyu: Please mark `const`. Thanks!

query_optimizer/resolver/NameResolver.hpp
```diff
@@ -47,8 +49,12 @@ class NameResolver {
  public:
   /**
    * @brief Constructor.
+   *
+   * @param parent_resolver The NameResolver inherited from the outer query.
+   *                        NULL if there is no outer query.
```
zuyu: FYI, as a convention, this line should align to the first letter of `parent_resolver`. See below for the more details:

```
  * @param parent_resolver The NameResolver inherited from the outer query.
  *        NULL if there is no outer query.
```

zuyu:
@jianqiao It generally looks good, but you may want to fix some style issues in other PR. Thanks!

PR #152, opened 2016-04-11
Query manager class and tests

hbdeshmukh:
This PR is about the class QueryManager. This is the first step in allowing concurrent (read-only) query execution. Its functionality includes:

- Generation of new work orders
- Storing pending (the ones not yet scheduled for execution) work orders, 
- Processing work order completion messages
- Allowing APIs to check the state of the execution of the query. 

Note that majority of the functions in this class were implemented in Foreman. There are some minor changes. The test for QueryManager also reuses majority of the Foreman unit test, however it has been simplified as follows:

- In Foreman unit test, we had to actually execute a WorkOrder (reference: https://github.com/pivotalsoftware/quickstep/blob/master/query_execution/tests/Foreman_unittest.cpp#L379) 
- In QueryManager unit test, the QueryManager just needs to process a message about WorkOrder completion, thereby eliminating the need of actual execution. This makes the test somewhat easier to read.
- In Foreman test, we kept checking about the state of the workers, where as in QueryManager unit test, we don't bother about workers at all. 

I apologize for the long PR (1800+ new lines), although bulk of the code is not new. As this PR is anyway a big change, I decided to create another PR (which should be manageable to review) for the following tasks:

- Use QueryManager in Foreman. 
- Make Foreman a long living thread (like workers). In which case, the main thread will be registered as a TMB client and new queries to Foreman will be sent as a message. 
- A hook for PolicyEnforcer (without its true functionalities) and use it in Foreman.


query_execution/QueryManager.cpp
```diff
@@ -0,0 +1,469 @@
...
+                           StorageManager *storage_manager,
+                           tmb::MessageBus *bus)
+      : foreman_client_id_(foreman_client_id),
+        query_id_(DCHECK_NOTNULL(query_handle)->query_id()),
+        catalog_database_(DCHECK_NOTNULL(catalog_database)),
+        storage_manager_(DCHECK_NOTNULL(storage_manager)),
+        bus_(DCHECK_NOTNULL(bus)) {
+  DCHECK_NOTNULL(query_handle->getQueryPlanMutable());
+  query_dag_ = query_handle->getQueryPlanMutable()->getQueryPlanDAGMutable();
+  DCHECK_NOTNULL(query_dag_);
```
zuyu: There are warnings (treated errors for us) for unused variables to use `DCHECK_NOTNULL` alone as a statement in the release mode.

On the other hand, we could have the following refactor in the init list:

```
  ...
  query_dag_(DCHECK_NOTNULL(
      DCHECK_NOTNULL(query_handle->getQueryPlanMutable())
          ->getQueryPlanDAGMutable()))
  ... {
```

query_execution/QueryManager.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+  const dag_node_index num_operators_in_dag = query_dag_->size();
+
+  output_consumers_.resize(num_operators_in_dag);
+  blocking_dependencies_.resize(num_operators_in_dag);
```
zuyu: This is optionally, but I prefer to construct them with explicit size, instead of doing a later `resize`.

So, I suggest the following refactor to move above two in the init list of the constructor:
```
  query_dag_(...),
  output_consumers_(query_dag_->size()),
  blocking_dependencies_(query_dag_->size()),
  ... {
```

zuyu:
Actually, I prefer you to include using `QueryManager` in `Foreman` in this PR. Thus, it shows how much portions of `Foreman` code moved into `QueryManager`, and we just will have more deletions. Thanks!

What do you think?

query_execution/QueryManager.cpp
```diff
@@ -0,0 +1,469 @@
...
+      // Possible message senders include InsertDestinations and some
+      // operators which modify existing blocks.
+      serialization::DataPipelineMessage proto;
+      CHECK(proto.ParseFromArray(tagged_message.message(),
+                                 tagged_message.message_bytes()));
+
+      op_index = proto.operator_index();
+      processDataPipelineMessage(proto.operator_index(),
+                                 proto.block_id(),
+                                 proto.relation_id());
```
zuyu: @hbdeshmukh I have a question here: do we lose the data locality information (i.e., the worker thread index) in `Foreman` when using `QueryManager` to process messages? Thanks.

query_execution/QueryManager.cpp
hbdeshmukh: I would rather not do that because it seems hard to read.

query_execution/QueryManager.cpp
hbdeshmukh: Because the query_dag_ isn't getting constructed in the initializer list, the vector initializations happen in the constructor body.

query_execution/QueryManager.cpp
hbdeshmukh: The information is available at this location. If the ``QueryManager`` wants to read which Worker sent this pipeline message, it can do that by reading the sender of the ``tagged_message``.

zuyu: I know `QueryManager` knows it, but my question was, does `Foreman` know the sender worker thread index for scheduling future related `WorkOrder`?

For example, a `WorkOrder` completion message for a `BuildHash` is received, how `Foreman` schedules a `HashJoinWorkOrder` to the same worker? In fact, I am particular interested in the distributed case where the same `worker` means the same `Shiftboss`. Thanks.

hbdeshmukh: Understood it now. In the next PR, I will add a function ``WorkerMessage::setRecipientHint()``, where the QueryManager can provide a hint to Foreman as to which worker should get that message. The Foreman can choose to use the hint or ignore it, depending on the dispatch strategy that it uses. 

In short, the sequence should be as follows:
1. QueryManager receives the pipeline message.
2. It reads the sender of the message from the message. 
3. Internally, it can maintain a map of block IDs to worker IDs (where the block was most likely produced).
4. When producing the work order for which the input is the pipelined block (from step 1), it can set the recipient hint as the sender of the message in step 1. 

For the current implementation, I won't actually provide support for all these. However, I will add these comments in the code for the ease of future implementation.

hbdeshmukh:
If I use this one single PR for all the changes, it may delay the review process and might make the review as more work. I should also mention that I am still working on the follow-up changes and the branch is not ready yet.

zuyu:
@hbdeshmukh Since I am the reviewer, I prefer to have `use QueryManager in Foreman` as a part of this PR.

If I understand correctly, doing so will just have more deletions, as most of the `Foreman` functionality has moved to `QueryManager`, right? Thanks!

hbdeshmukh:
Well, it's more than that. Here are the steps -

1. Add QueryManager class and tests. 
2. Add query ID field to WorkOrders.
3. Add query ID field to pretty much all the messages and query context. 
4. Add a placeholder for PolicyEnforcer class. This class receives messages from Foreman, reroutes them to appropriate QueryManager. On the other side, it collects WorkerMessages (consisting of WorkOrders) from QueryManagers and sends them to Foreman. 
5. Use PolicyEnforcer in Foreman, by dropping all the existing functionalities related to QueryProgress in Foreman.
6. Make Foreman a long-living thread by creating a message for new query that gets sent to Foreman from the main thread. Since we have some tests (query optimizer) with similar working setups, I will have to modify them as well. 

Steps 1, 2 and 3 (almost) are done already (You can check them [here](https://github.com/pivotalsoftware/quickstep/compare/query-manager-used-in-foreman). I am not sure why you would want to review all the changes at once, but if that's what you prefer - I can create a single PR. In that case, we can close this PR and create a new one when the other branch is ready.

zuyu:
@hbdeshmukh I suggest this PR should also allow `Foreman` to call `QueryManager` using the functions moved from `Foreman` to `QueryManager`, including `getNextWorkerMessage` and `processMessage`. In other words, `Foreman` should remove all the functions covered by `QueryManager` in the same PR.

No `query_id`, `PolicyEnforcer`, nor long-running `Foreman`.

query_execution/QueryManager.hpp
```diff
@@ -0,0 +1,369 @@
...
+   *
+   * @param foreman_client_id The TMB client ID of the foreman thread.
+   * @param num_numa_nodes The number of NUMA nodes used by the system.
+   * @param query_handle The QueryHandle object for this query.
+   * @param catalog_database The CatalogDatabse used by the query.
+   * @param storage_manager The StorageManager used by the query.
+   * @param bus The TMB used for communication.
+   **/
+  QueryManager(const tmb::client_id foreman_client_id,
+               const int num_numa_nodes,
```
zuyu: Please use `numa_node_id` instead of `int` for `num_numa_nodes`. Also, include `/catalog/CatalogTypedefs.hpp`.

query_execution/QueryManager.hpp
```diff
@@ -0,0 +1,369 @@
...
+   * @param start_operator_index Begin the search for the schedulable WorkOrder
+   *        with the operator at this index.
+   * @param numa_node The next WorkOrder should preferably have its input(s)
+   *        from this numa_node. This is a hint and not a binding requirement.
+   *
+   * @return A pointer to the WorkerMessage. If there's no WorkOrder to be
+   *         executed, return NULL.
+   **/
+  WorkerMessage* getNextWorkerMessage(
+      const dag_node_index start_operator_index, const int numa_node = -1);
```
zuyu: Ditto for `numa_node_id`.

query_execution/QueryManager.hpp
```diff
@@ -0,0 +1,369 @@
...
+   *
+   * @return QueryStatusCode as determined after the message is processed.
+   **/
+  QueryStatusCode processMessage(const TaggedMessage &tagged_message);
+
+  /**
+   * @brief Get the QueryExecutionState for this query.
+   **/
+  inline const QueryExecutionState& getQueryExecutionState() const {
+    return *(query_exec_state_.get());
```
zuyu: Refactor to `return *query_exec_state_;`.

FYI, `unique_ptr` supports `operator*`.

query_execution/tests/QueryManager_unittest.cpp
```diff
@@ -0,0 +1,933 @@
...
+
+  const MockOperator &op = static_cast<const MockOperator &>(
+      query_plan_->getQueryPlanDAG().getNodePayload(id));
+
+  constructQueryManager();
+
+  // op doesn't have any dependencies.
+  EXPECT_TRUE(op.getBlockingDependenciesMet());
+
+  for (int i = 0; i < 3; i++) {
```
zuyu: Change to `++i`.

query_execution/tests/QueryManager_unittest.cpp
```diff
@@ -0,0 +1,933 @@
...
+      query_plan_->addRelationalOperator(new MockOperator(true, false));
+  const QueryPlan::DAGNodeIndex id2 =
+      query_plan_->addRelationalOperator(new MockOperator(true, false));
+
+  // Create a blocking link.
+  query_plan_->addDirectDependency(id2, id1, true);
+
+  static_cast<MockOperator *>(
+      query_plan_->getQueryPlanDAGMutable()->getNodePayloadMutable(id1))
+      ->setOutputRelationID(0xdead);
```
zuyu: Please add four more whitespace indentations for this line.

See the line length part in https://github.com/pivotalsoftware/quickstep/wiki/Quickstep-Style-Guide.

query_execution/QueryManager.hpp
hbdeshmukh: ``num_numa_nodes`` is just a count of NUMA nodes used. I have changed it to ``size_t`` though, as it will always be non-negative.

query_execution/QueryManager.hpp
hbdeshmukh: Done.

zuyu:
LGTM. Nice job!

PR #154, opened 2016-04-11
Optimizer support for the EXISTS subquery.

jianqiao:
This PR adds the optimizer support for `EXISTS` subqueries. An `EXISTS` subquery will be converted into _left semi_ hash join or _left anti_ hash join by the `UnnestSubqueries` logical transformation.

### DIFF
--- a/query_optimizer/logical/Aggregate.cpp
+++ b/query_optimizer/logical/Aggregate.cpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -24,6 +26,7 @@
 #include "query_optimizer/expressions/AttributeReference.hpp"
 #include "query_optimizer/expressions/ExpressionUtil.hpp"
 #include "query_optimizer/expressions/NamedExpression.hpp"
+#include "query_optimizer/expressions/PatternMatcher.hpp"
 #include "utility/Cast.hpp"
 
 #include "glog/logging.h"
@@ -70,6 +73,35 @@ std::vector<E::AttributeReferencePtr> Aggregate::getReferencedAttributes() const
                                  referenced_attributes_in_expression.end());
   }
   return referenced_attributes;
+}
+
+LogicalPtr Aggregate::copyWithNewInputExpressions(
+      const std::vector<E::ExpressionPtr>& input_expressions) const {
+  DCHECK_EQ(grouping_expressions_.size() + aggregate_expressions_.size(),
+            input_expressions.size());
+
+  std::vector<E::NamedExpressionPtr> new_grouping_expressions;
+  for (std::vector<E::ExpressionPtr>::size_type index = 0;
+       index < grouping_expressions_.size();
+       ++index) {
+    E::NamedExpressionPtr grouping_expression;
+    E::SomeNamedExpression::MatchesWithConditionalCast(input_expressions[index],
+                                                       &grouping_expression);
+    DCHECK(grouping_expression != nullptr);
+    new_grouping_expressions.emplace_back(grouping_expression);
+  }
+
+  std::vector<E::AliasPtr> new_aggregate_expressions;
+  for (std::vector<E::ExpressionPtr>::size_type index = grouping_expressions_.size();
+       index < input_expressions.size();
+       ++index) {
+    E::AliasPtr aggregate_expression;
+    E::SomeAlias::MatchesWithConditionalCast(input_expressions[index], &aggregate_expression);
+    DCHECK(aggregate_expression != nullptr);
+    new_aggregate_expressions.emplace_back(aggregate_expression);
+  }
+
+  return Create(input_, new_grouping_expressions, new_aggregate_expressions);
 }
 
 void Aggregate::getFieldStringItems(
--- a/query_optimizer/logical/Aggregate.cpp
+++ b/query_optimizer/logical/Aggregate.cpp
@@ -76,7 +76,7 @@ std::vector<E::AttributeReferencePtr> Aggregate::getReferencedAttributes() const
 }
 
 LogicalPtr Aggregate::copyWithNewInputExpressions(
-      const std::vector<E::ExpressionPtr>& input_expressions) const {
+    const std::vector<E::ExpressionPtr> &input_expressions) const {
   DCHECK_EQ(grouping_expressions_.size() + aggregate_expressions_.size(),
             input_expressions.size());
 
--- a/query_optimizer/logical/Filter.hpp
+++ b/query_optimizer/logical/Filter.hpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -24,6 +26,7 @@
 
 #include "query_optimizer/OptimizerTree.hpp"
 #include "query_optimizer/expressions/AttributeReference.hpp"
+#include "query_optimizer/expressions/Expression.hpp"
 #include "query_optimizer/expressions/LogicalAnd.hpp"
 #include "query_optimizer/expressions/Predicate.hpp"
 #include "query_optimizer/logical/Logical.hpp"
@@ -70,6 +73,9 @@ class Filter : public Logical {
     DCHECK_EQ(new_children.size(), children().size());
     return Filter::Create(new_children[0], filter_predicate_);
   }
+
+  LogicalPtr copyWithNewInputExpressions(
+      const std::vector<expressions::ExpressionPtr>& input_expressions) const override;
 
   std::vector<expressions::AttributeReferencePtr> getOutputAttributes() const override {
     return input()->getOutputAttributes();
--- a/query_optimizer/logical/Filter.hpp
+++ b/query_optimizer/logical/Filter.hpp
@@ -75,7 +75,7 @@ class Filter : public Logical {
   }
 
   LogicalPtr copyWithNewInputExpressions(
-      const std::vector<expressions::ExpressionPtr>& input_expressions) const override;
+      const std::vector<expressions::ExpressionPtr> &input_expressions) const override;
 
   std::vector<expressions::AttributeReferencePtr> getOutputAttributes() const override {
     return input()->getOutputAttributes();
--- a/query_optimizer/logical/HashJoin.hpp
+++ b/query_optimizer/logical/HashJoin.hpp
@@ -54,7 +54,7 @@ typedef std::shared_ptr<const HashJoin> HashJoinPtr;
  */
 class HashJoin : public BinaryJoin {
  public:
-  enum JoinType {
+  enum class JoinType {
     kInnerJoin = 0,
     kLeftSemiJoin,
     kLeftAntiJoin
@@ -65,17 +65,16 @@ class HashJoin : public BinaryJoin {
 
   std::string getName() const override {
     switch (join_type_) {
-      case kInnerJoin:
+      case JoinType::kInnerJoin:
         return "HashJoin";
-      case kLeftSemiJoin:
+      case JoinType::kLeftSemiJoin:
         return "HashLeftSemiJoin";
-      case kLeftAntiJoin:
+      case JoinType::kLeftAntiJoin:
         return "HashLeftAntiJoin";
       default:
         LOG(FATAL) << "Invalid JoinType: "
                    << static_cast<typename std::underlying_type<JoinType>::type>(join_type_);
     }
-    QUICKSTEP_UNREACHABLE();
   }
 
   /**
--- a/query_optimizer/logical/HashJoin.hpp
+++ b/query_optimizer/logical/HashJoin.hpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -20,10 +22,12 @@
 
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include "query_optimizer/OptimizerTree.hpp"
 #include "query_optimizer/expressions/AttributeReference.hpp"
+#include "query_optimizer/expressions/Predicate.hpp"
 #include "query_optimizer/logical/BinaryJoin.hpp"
 #include "query_optimizer/logical/Logical.hpp"
 #include "query_optimizer/logical/LogicalType.hpp"
@@ -50,9 +54,29 @@ typedef std::shared_ptr<const HashJoin> HashJoinPtr;
  */
 class HashJoin : public BinaryJoin {
  public:
+  enum JoinType {
+    kInnerJoin = 0,
+    kLeftSemiJoin,
+    kLeftAntiJoin
+  };
+
+
   LogicalType getLogicalType() const override { return LogicalType::kHashJoin; }
 
-  std::string getName() const override { return "HashJoin"; }
+  std::string getName() const override {
+    switch (join_type_) {
+      case kInnerJoin:
+        return "HashJoin";
+      case kLeftSemiJoin:
+        return "HashLeftSemiJoin";
+      case kLeftAntiJoin:
+        return "HashLeftAntiJoin";
+      default:
+        LOG(FATAL) << "Invalid JoinType: "
+                   << static_cast<typename std::underlying_type<JoinType>::type>(join_type_);
+    }
+    QUICKSTEP_UNREACHABLE();
+  }
 
   /**
    * @brief Join attributes in the left logical 'left_'.
@@ -68,13 +92,29 @@ class HashJoin : public BinaryJoin {
     return right_join_attributes_;
   }
 
+  /**
+   * @brief The filtering predicate evaluated after join.
+   */
+  const expressions::PredicatePtr& residual_predicate() const {
+    return residual_predicate_;
+  }
+
+  /**
+   * @return Join type of this hash join.
+   */
+  JoinType join_type() const {
+    return join_type_;
+  }
+
   LogicalPtr copyWithNewChildren(
       const std::vector<LogicalPtr> &new_children) const override {
     DCHECK_EQ(new_children.size(), children().size());
     return HashJoin::Create(new_children[0],
                             new_children[1],
                             left_join_attributes_,
-                            right_join_attributes_);
+                            right_join_attributes_,
+                            residual_predicate_,
+                            join_type_);
   }
 
   std::vector<expressions::AttributeReferencePtr> getReferencedAttributes() const override {
@@ -98,13 +138,22 @@ class HashJoin : public BinaryJoin {
    * @param right The right input operator.
    * @param left_join_attributes The join attributes in the 'left'.
    * @param right_join_attributes The join attributes in the 'right'.
+   * @param Join type of this hash join.
    * @return An immutable HashJoin.
    */
   static HashJoinPtr Create(const LogicalPtr &left,
                             const LogicalPtr &right,
                             const std::vector<expressions::AttributeReferencePtr> &left_join_attributes,
-                            const std::vector<expressions::AttributeReferencePtr> &right_join_attributes) {
-    return HashJoinPtr(new HashJoin(left, right, left_join_attributes, right_join_attributes));
+                            const std::vector<expressions::AttributeReferencePtr> &right_join_attributes,
+                            const expressions::PredicatePtr &residual_predicate,
+                            const JoinType join_type) {
+    return HashJoinPtr(
+        new HashJoin(left,
+                     right,
+                     left_join_attributes,
+                     right_join_attributes,
+                     residual_predicate,
+                     join_type));
   }
 
  protected:
@@ -132,16 +181,26 @@ class HashJoin : public BinaryJoin {
   HashJoin(const LogicalPtr &left,
            const LogicalPtr &right,
            const std::vector<expressions::AttributeReferencePtr> &left_join_attributes,
-           const std::vector<expressions::AttributeReferencePtr> &right_join_attributes)
+           const std::vector<expressions::AttributeReferencePtr> &right_join_attributes,
+           const expressions::PredicatePtr &residual_predicate,
+           const JoinType join_type)
       : BinaryJoin(left, right),
         left_join_attributes_(left_join_attributes),
-        right_join_attributes_(right_join_attributes) {
+        right_join_attributes_(right_join_attributes),
+        residual_predicate_(residual_predicate),
+        join_type_(join_type) {
     DCHECK_EQ(left_join_attributes.size(), right_join_attributes.size());
     DCHECK(!left_join_attributes.empty());
+
+    if (residual_predicate_ != nullptr) {
+      addInputExpression(residual_predicate_);
+    }
   }
 
   std::vector<expressions::AttributeReferencePtr> left_join_attributes_;
   std::vector<expressions::AttributeReferencePtr> right_join_attributes_;
+  expressions::PredicatePtr residual_predicate_;
+  const JoinType join_type_;
 
   DISALLOW_COPY_AND_ASSIGN(HashJoin);
 };
--- a/query_optimizer/logical/Logical.hpp
+++ b/query_optimizer/logical/Logical.hpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -23,8 +25,11 @@
 
 #include "query_optimizer/OptimizerTree.hpp"
 #include "query_optimizer/expressions/AttributeReference.hpp"
+#include "query_optimizer/expressions/Expression.hpp"
 #include "query_optimizer/logical/LogicalType.hpp"
 #include "utility/Macros.hpp"
+
+#include "glog/logging.h"
 
 namespace quickstep {
 namespace optimizer {
@@ -40,7 +45,8 @@ typedef std::shared_ptr<const Logical> LogicalPtr;
 /**
  * @brief Base class for all logical operator/node.
  */
-class Logical : public OptimizerTree<Logical> {
+class Logical : public OptimizerTree<Logical>,
+                public std::enable_shared_from_this<const Logical> {
  public:
   /**
    * @brief Destructor.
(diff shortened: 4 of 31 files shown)


query_optimizer/logical/Aggregate.cpp
```diff
@@ -72,6 +75,35 @@ std::vector<E::AttributeReferencePtr> Aggregate::getReferencedAttributes() const
   return referenced_attributes;
 }
 
+LogicalPtr Aggregate::copyWithNewInputExpressions(
+      const std::vector<E::ExpressionPtr>& input_expressions) const {
```
zuyu: What do you think about the reference sigils standards in Quickstep? If it applies, it looks like `const std::vector<E::ExpressionPtr> &input_expressions`.

https://github.com/pivotalsoftware/quickstep/wiki/Quickstep-Style-Guide

query_optimizer/logical/HashJoin.hpp
```diff
@@ -50,9 +54,29 @@ typedef std::shared_ptr<const HashJoin> HashJoinPtr;
  */
 class HashJoin : public BinaryJoin {
  public:
+  enum JoinType {
```
zuyu: Could we use enum classes here?

query_optimizer/logical/Aggregate.cpp
pateljm: I like the nicer Zuyu :-) You are right @zuyu -- we should try to use the standard here

query_optimizer/physical/HashJoin.hpp
```diff
@@ -50,9 +53,28 @@ typedef std::shared_ptr<const HashJoin> HashJoinPtr;
  */
 class HashJoin : public BinaryJoin {
  public:
+  enum JoinType {
```
zuyu: Could we use the enum class here?

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+namespace E = ::quickstep::optimizer::expressions;
+namespace L = ::quickstep::optimizer::logical;
+
+L::LogicalPtr PushDownSemiAntiJoin::applyToNode(const L::LogicalPtr &input) {
+  L::HashJoinPtr hash_join;
+
+  const std::vector<L::LogicalPtr> children = input->children();
+  std::vector<L::LogicalPtr> new_children;
+  bool has_changes = false;
+  for (const L::LogicalPtr& child : children) {
```
zuyu: What do you think of how the reference sigil applies to here? `const L::LogicalPtr &child`.

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+namespace quickstep {
+namespace optimizer {
+
+namespace E = ::quickstep::optimizer::expressions;
+namespace L = ::quickstep::optimizer::logical;
+
+L::LogicalPtr PushDownSemiAntiJoin::applyToNode(const L::LogicalPtr &input) {
+  L::HashJoinPtr hash_join;
+
+  const std::vector<L::LogicalPtr> children = input->children();
```
zuyu: Could we use a reference here, instead of a deep copy?

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+  std::vector<L::LogicalPtr> new_children;
+  bool has_changes = false;
+  for (const L::LogicalPtr& child : children) {
+    if (L::SomeHashJoin::MatchesWithConditionalCast(child, &hash_join)) {
+      if (hash_join->join_type() == L::HashJoin::kLeftSemiJoin
+          || hash_join->join_type() == L::HashJoin::kLeftAntiJoin) {
```
zuyu: What do you think of being consistent in conjunctive / disjunctive expressions in a condition?

I like the following format from the code in Line 80 from the same file, could we also apply to here? Thanks!
```
  if (input_index < last_input_index &&
      semi_anti_join->residual_predicate() != nullptr) {
```

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+        break;
+      }
+      ++input_index;
+    }
+
+    if (input_index < last_input_index &&
+        semi_anti_join->residual_predicate() != nullptr) {
+      const std::vector<E::AttributeReferencePtr> input_child_output_attributes =
+          left_input_children[input_index]->getOutputAttributes();
+      const std::vector<E::AttributeReferencePtr> right_input_output_attributes =
```
zuyu: Could we mark both `input_child_output_attributes` and `right_input_output_attributes` as `reference`s, instead of doing deep copies? Thanks!

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+      E::UnorderedAttributeSet visible_attribute_set(input_child_output_attributes.begin(),
+                                                     input_child_output_attributes.end());
+      visible_attribute_set.insert(right_input_output_attributes.begin(),
+                                   right_input_output_attributes.end());
+
+      const std::vector<E::AttributeReferencePtr> referenced_attrs_in_residual =
+          E::GetAttributeReferencesWithinScope(
+              semi_anti_join->residual_predicate()->getReferencedAttributes(),
+              E::AttributeReferenceScope::kLocal);
+      for (const E::AttributeReferencePtr& referenced_attr : referenced_attrs_in_residual) {
```
zuyu: Could we think of the `reference sigil` style here?

`const E::AttributeReferencePtr &referenced_attr`.

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+      const std::vector<E::AttributeReferencePtr> input_child_output_attributes =
+          left_input_children[input_index]->getOutputAttributes();
+      const std::vector<E::AttributeReferencePtr> right_input_output_attributes =
+          semi_anti_join->right()->getOutputAttributes();
+      E::UnorderedAttributeSet visible_attribute_set(input_child_output_attributes.begin(),
+                                                     input_child_output_attributes.end());
+      visible_attribute_set.insert(right_input_output_attributes.begin(),
+                                   right_input_output_attributes.end());
+
+      const std::vector<E::AttributeReferencePtr> referenced_attrs_in_residual =
```
zuyu: Could we use a reference, instead of a deep copy?

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+#include "glog/logging.h"
+
+namespace quickstep {
+namespace optimizer {
+
+namespace E = ::quickstep::optimizer::expressions;
+namespace L = ::quickstep::optimizer::logical;
+
+struct CorrelatedQueryInfo {
+  enum QueryType {
```
zuyu: Could we use the enum class here? Also for consistency, could we rename `QueryType` to `JoinType`? Thanks!

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+        correlated_query(correlated_query_in),
+        probe_join_attributes(std::move(probe_join_attributes_in)),
+        build_join_attributes(std::move(build_join_attributes_in)),
+        non_hash_join_predicates(std::move(non_hash_join_predicates_in)) {}
+
+  QueryType query_type_;
+  L::LogicalPtr correlated_query;
+  std::vector<E::AttributeReferencePtr> probe_join_attributes;
+  std::vector<E::AttributeReferencePtr> build_join_attributes;
+  std::vector<E::PredicatePtr> non_hash_join_predicates;
```
zuyu: Could we mark these argument `const`?

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+};
+
+L::LogicalPtr UnnestSubqueries::apply(const L::LogicalPtr &input) {
+  DCHECK(L::SomeTopLevelPlan::Matches(input))
+      << "The input logical node must be of TopLevelPlan";
+
+  const L::TopLevelPlan* top_level_plan =
```
zuyu: What do you think of applying the pointer sigil style here?

`const L::TopLevelPlan *top_level_plan`.

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+                                 inner_attributes->end());
+        return L::Aggregate::Create(aggregate->input(),
+                                    group_expressions,
+                                    aggregate->aggregate_expressions());
+      }
+      return input_with_no_subqueries;
+    }
+    case L::LogicalType::kFilter: {
+      if (new_input_expressions.empty()) {
+        const L::Filter* filter = static_cast<const L::Filter*>(input_with_no_subqueries.get());
```
zuyu: What do you think of applying the pointer sigil style here?

`const L::Filter *filter ...`

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+      return input_with_no_subqueries;
+  }
+}
+
+E::ExpressionPtr UnnestSubqueriesForNonRootLogical::eliminateOuterAttributeReference(
+    const E::ExpressionPtr &expression,
+    E::UnorderedAttributeSet *inner_attributes,
+    std::vector<E::AttributeReferencePtr> *probe_join_attributes,
+    std::vector<E::AttributeReferencePtr> *build_join_attributes,
+    std::vector<E::PredicatePtr>* non_hash_join_predicates) {
```
zuyu: Could we be consistent in the pointer sigil style here? Thanks!

`std::vector<E::PredicatePtr> *non_hash_join_predicates`

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+  DCHECK(expression->getExpressionType() != E::ExpressionType::kExists);
+  DCHECK(expression->getExpressionType() != E::ExpressionType::kSubqueryExpression);
+
+  switch (expression->getExpressionType()) {
+    case E::ExpressionType::kPredicateLiteral:
```
zuyu: Could we add comments as the following:

```
  case E::ExpressionType::kPredicateLiteral:  // Fall through.
```

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+    case E::ExpressionType::kScalarLiteral:
+      return expression;
+    case E::ExpressionType::kLogicalAnd: {
+      const E::LogicalAnd* logical_and = static_cast<const E::LogicalAnd*>(expression.get());
```
zuyu: Ditto for the pointer sigil style.

`const E::LogicalAnd *logical_and ...`

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+    default: {
+      validateNonOuterAttributeReference(expression);
+      return expression;
+    }
+  }
+}
+
+void UnnestSubqueriesForNonRootLogical::validateNonOuterAttributeReference(
+    const E::ExpressionPtr &expression) {
+  const std::vector<E::AttributeReferencePtr> referenced_attributes =
```
zuyu: Could we use a reference here, instead of a deep copy?

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+      THROW_SQL_ERROR()
+          << "Outer attribute reference can only appear in a single-attribute equality predicate";
+    }
+  }
+}
+
+bool UnnestSubqueriesForNonRootLogical::isCorrelatedOuterAttribute(
+    const E::AttributeReferencePtr &attribute) const {
+  return attribute->scope() == E::AttributeReferenceScope::kOuter
+         && uncorrelated_subqueries_->find(attribute->id()) == uncorrelated_subqueries_->end();
```
zuyu: For the consistency, we put `&&` to the end of the line. So in this case:

```
  return attribute->scope() == E::AttributeReferenceScope::kOuter &&
         uncorrelated_subqueries_->find(attribute->id()) == uncorrelated_subqueries_->end();
```

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+}
+
+L::LogicalPtr UnnestSubqueriesForNonRootLogical::eliminateNestedScalarQueries(const L::LogicalPtr &node) {
+  const std::vector<E::ExpressionPtr> &input_expressions = node->input_expressions();
+  const std::vector<E::AttributeReferencePtr> input_attributes =
```
zuyu: Like Line 427, could we use the reference here, instead of a deep copy?

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+    new_input_expressions.emplace_back(new_input_expression);
+  }
+
+  if (has_changed_expression) {
+    // If there are correlated subquery expressions, add logical joins.
+    if (!correlated_query_info_vec.empty()) {
+      L::LogicalPtr new_child;
+
+      if (new_input_expressions[0] == nullptr
+          && node->getLogicalType() == L::LogicalType::kNestedLoopsJoin) {
```
zuyu: For the consistency, could we move `&&` to the end of Line 458? Thanks!

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+        new_child = node;
+      } else {
+        DCHECK_EQ(node->children().size(), 1u);
```
zuyu: Do you remember that we put the expected value in the left hand side, while the checked on in the right?

`DCHECK_EQ(1u, node->children().size());`

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+                                          correlated_query_info.probe_join_attributes,
+                                          correlated_query_info.build_join_attributes,
+                                          filter_predicate,
+                                          join_type);
+        }
+      }
+
+      // Special case for filter and nested loops join which may contain EXISTS and IN.
+      if (new_input_expressions[0] == nullptr) {
+        DCHECK_EQ(new_input_expressions.size(), 1u);
```
zuyu: As a convention for both `CHECK_EQ` and `DCHECK_EQ`, we put the expected value in the left hand side, while the checked one in the right.

So for this case, `DCHECK_EQ(1u, new_input_expressions.size());`.

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+    LOG_APPLYING_RULE(node, new_logical);
+    return new_logical;
+  }
+
+  LOG_IGNORING_RULE(node);
+  return node;
+}
+
+E::ExpressionPtr UnnestSubqueriesForExpession::applyInternal(
+    bool allow_exists_or_in,
```
zuyu: Could we mark `const` here?

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+                                           std::move(probe_join_attributes),
+                                           std::move(build_join_attributes),
+                                           std::move(non_hash_join_predicates));
+}
+
+E::ExpressionPtr DeOuterAttributeReference::applyToNode(const E::ExpressionPtr &input) {
+  E::AttributeReferencePtr attr;
+  if (E::SomeAttributeReference::MatchesWithConditionalCast(input, &attr)
+      && attr->scope() == E::AttributeReferenceScope::kOuter
+      && uncorrelated_subqueries_.find(attr->id()) == uncorrelated_subqueries_.end()) {
```
zuyu: As a convention in our codebase, we put `&&` or `||` at the end of line.

So it looks like:

```
  if (E::SomeAttributeReference::MatchesWithConditionalCast(input, &attr) &&
      attr->scope() == E::AttributeReferenceScope::kOuter &&
      uncorrelated_subqueries_.find(attr->id()) == uncorrelated_subqueries_.end()) {
```

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -0,0 +1,720 @@
...
+                                          filter_predicate,
+                                          join_type);
+        }
+      }
+
+      // Special case for filter and nested loops join which may contain EXISTS and IN.
+      if (new_input_expressions[0] == nullptr) {
+        DCHECK_EQ(new_input_expressions.size(), 1u);
+        DCHECK(node->getLogicalType() == L::LogicalType::kFilter
+               || node->getLogicalType() == L::LogicalType::kNestedLoopsJoin) << node->toString();
```
zuyu: As a convention in our codebase, we put `&&` or `||` at the end of line. So it looks like

```
  DCHECK(node->getLogicalType() == L::LogicalType::kFilter ||
         node->getLogicalType() == L::LogicalType::kNestedLoopsJoin) << node->toString();
```

query_optimizer/rules/UnnestSubqueries.hpp
```diff
@@ -0,0 +1,219 @@
...
+  expressions::ExpressionPtr apply(const expressions::ExpressionPtr &node) override {
+    return applyInternal(true /* allow_exists_or_in */,
+                         node);
+  }
+
+ private:
+  const expressions::UnorderedNamedExpressionSet &visible_attributes_from_outer_query_;
+
+  expressions::ExpressionPtr applyInternal(
+      bool allow_exists_or_in,
```
zuyu: Could we mark `const` here?

query_optimizer/rules/UnnestSubqueries.hpp
```diff
@@ -0,0 +1,219 @@
...
+ private:
+  OptimizerContext *context_;
+
+  DISALLOW_COPY_AND_ASSIGN(UnnestSubqueries);
+};
+
+class UnnestSubqueriesForNonRootLogical : public Rule<logical::Logical> {
+ public:
+  UnnestSubqueriesForNonRootLogical(
+      bool scalar_query,
```
zuyu: Could we mark `const` here?

query_optimizer/rules/UnnestSubqueries.hpp
```diff
@@ -0,0 +1,219 @@
...
+  logical::LogicalPtr eliminateNestedScalarQueries(const logical::LogicalPtr &node);
+
+  // Eliminate outer attribute references in this node, add predicates with
+  // outer attribute references into join attributes or non_hash_join_predicates.
+  expressions::ExpressionPtr eliminateOuterAttributeReference(
+      const expressions::ExpressionPtr &expression,
+      expressions::UnorderedAttributeSet *inner_attributes,
+      std::vector<expressions::AttributeReferencePtr> *probe_join_attributes,
+      std::vector<expressions::AttributeReferencePtr> *build_join_attributes,
+      std::vector<expressions::PredicatePtr>* non_hash_join_predicates);
```
zuyu: What do you think if the pointer sigil standard applies to here?

`std::vector<expressions::PredicatePtr> *non_hash_join_predicates`.

jianqiao:
@zuyu Thanks for the comments! Actually the `UnnestSubqueries.*` files are somehow directly pulled from the old quickstep repo since the functionality is standalone, and I didn't look too much into it styles. Seems that the file has quite a lot of style issues. Can you help fix that in an overall sense? I will address the rest files.

query_optimizer/rules/UnnestSubqueries.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+  void validateNonOuterAttributeReference(const expressions::ExpressionPtr &expression);
+
+  bool isCorrelatedOuterAttribute(const expressions::AttributeReferencePtr &attribute) const;
+
+  bool scalar_query_;
```
zuyu: Could we mark it `const` here? Or we may change it later?

zuyu:
@jianqiao Sure. I'll let you know once I've done one pass.

query_optimizer/logical/Filter.hpp
```diff
@@ -71,6 +74,9 @@ class Filter : public Logical {
     return Filter::Create(new_children[0], filter_predicate_);
   }
 
+  LogicalPtr copyWithNewInputExpressions(
+      const std::vector<expressions::ExpressionPtr>& input_expressions) const override;
```
zuyu: If we apply the reference sigil standard here, it should look like:

`const std::vector<expressions::ExpressionPtr> &input_expressions) const override;`

query_optimizer/logical/HashJoin.hpp
```diff
@@ -50,9 +54,29 @@ typedef std::shared_ptr<const HashJoin> HashJoinPtr;
...
+        return "HashJoin";
+      case kLeftSemiJoin:
+        return "HashLeftSemiJoin";
+      case kLeftAntiJoin:
+        return "HashLeftAntiJoin";
+      default:
+        LOG(FATAL) << "Invalid JoinType: "
+                   << static_cast<typename std::underlying_type<JoinType>::type>(join_type_);
+    }
+    QUICKSTEP_UNREACHABLE();
```
zuyu: I think we could delete this line, as we either return a value in all valid conditions, or abort the program by default.

query_optimizer/logical/HashJoin.hpp
```diff
@@ -132,16 +181,26 @@ class HashJoin : public BinaryJoin {
...
     DCHECK(!left_join_attributes.empty());
+
+    if (residual_predicate_ != nullptr) {
+      addInputExpression(residual_predicate_);
+    }
   }
 
   std::vector<expressions::AttributeReferencePtr> left_join_attributes_;
   std::vector<expressions::AttributeReferencePtr> right_join_attributes_;
+  expressions::PredicatePtr residual_predicate_;
```
zuyu: Could we mark `const` here?

query_optimizer/logical/Logical.hpp
```diff
@@ -64,12 +70,62 @@ class Logical : public OptimizerTree<Logical> {
...
+  }
+
+  /**
+   * @return The input attributes, which are the union set of the output attributes of
+   *         all children.
+   */
+  std::vector<expressions::AttributeReferencePtr> getInputAttributes() const {
+    std::vector<expressions::AttributeReferencePtr> input_attributes;
+    for (const LogicalPtr &child : children()) {
+      const std::vector<expressions::AttributeReferencePtr> child_output_attributes =
```
zuyu: Could we use a reference here, instead of a deep copy?

query_optimizer/logical/TopLevelPlan.hpp
```diff
@@ -124,6 +141,8 @@ class TopLevelPlan : public Logical {
   // Stored in the topological ordering based on dependencies.
   std::vector<LogicalPtr> shared_subplans_;
 
+  std::unordered_map<expressions::ExprId, int> uncorrelated_subquery_map_;
```
zuyu: Could we mark `const` here?

query_optimizer/physical/HashJoin.hpp
```diff
@@ -50,9 +53,28 @@ typedef std::shared_ptr<const HashJoin> HashJoinPtr;
...
+        return "HashJoin";
+      case kLeftSemiJoin:
+        return "HashLeftSemiJoin";
+      case kLeftAntiJoin:
+        return "HashLeftAntiJoin";
+      default:
+        LOG(FATAL) << "Invalid JoinType: "
+                   << static_cast<typename std::underlying_type<JoinType>::type>(join_type_);
+    }
+    QUICKSTEP_UNREACHABLE();
```
zuyu: I think we could remove this line, as we either return a value, or abort the program.

query_optimizer/physical/TopLevelPlan.hpp
```diff
@@ -142,6 +164,7 @@ class TopLevelPlan : public Physical {
   PhysicalPtr plan_;
   // Stored in the topological ordering based on dependencies.
   std::vector<PhysicalPtr> shared_subplans_;
+  std::unordered_map<expressions::ExprId, int> uncorrelated_subquery_map_;
```
zuyu: Could we mark `const` here?

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+  if (has_changes) {
+    return input->copyWithNewChildren(new_children);
+  } else {
+    return input;
+  }
+}
+
+L::LogicalPtr PushDownSemiAntiJoin::pushDownSemiAntiJoin(
+    const L::HashJoinPtr &semi_anti_join) {
+  const L::LogicalPtr left_input = semi_anti_join->left();
```
zuyu: Could we use a reference here for `left_input`?

query_optimizer/rules/PushDownSemiAntiJoin.cpp
```diff
@@ -0,0 +1,115 @@
...
+  std::vector<L::LogicalPtr> left_input_children = left_input->children();
+
+  if (!left_input_children.empty()) {
+    // Cannot push down a Filter down the right child of LeftOuterJoin.
+    std::vector<L::LogicalPtr>::size_type last_input_index = left_input_children.size();
```
zuyu: Could we mark `const` here?

zuyu:
@jianqiao I've done one pass, and once we fix these style issues, we are good to merge. Thanks!

jianqiao:
@zuyu You can push the changes to `UnnestSubqueries.*` directly into this branch and I will merge them together with other style fixes.

query_optimizer/logical/HashJoin.hpp
jianqiao: In fact all member variables of `Logical` subclasses can be `const` because the only way to access a `Logical` node is via `std::shard_ptr<const SubclassOfLogical>`. So in terms of readability, we can just avoid the redundancy by decorating none of them. Or we may declare all member variables of all subclasses of `Logical` to be `const`.

query_optimizer/logical/Logical.hpp
jianqiao: `child->getOutputAttributes()` is not returning a reference but a stack object.

zuyu: Then we may want to refactor as the following, to take advantage of the implicit `move` on a temporary variable provided by the compiler, instead of a deep copy resulted from the assign operator. Thanks!

```
  const std::vector<expressions::AttributeReferencePtr> child_output_attributes(child->getOutputAttributes());
```

You may need to wrap to a new line for readability, though.

jianqiao: I think typically there are three scenarios.

Senario 1:
```
A a;
A b = a;
```
In this case, `A b = a` is equivalent to `A b(a)` which will invoke `b`'s **copy constructor** with `a` as parameter.

Scenario 2:
```
A a;
A b = std::move(a);
```
In this case, `A b = std::move(a)` is equivalent to `A b(std::move(a))` which will invoke `b`'s **move constructor** with `std::move(a)` as parameter.

Scenario 3:
```
A f() {
  A a;
  return a;
}

void main() {
  A b = f();
}
```
In this scenario, typically if the `A`'s constructor has no side effect, then there will be **neither copy construction nor move construction**, known as the Named Return Value Optimization (NRVO).

zuyu: I think I am proposing the fourth scenario using the `move` constructor:

```
A f() {
  A a;
  return a;
}

A a(f());
```

jianqiao: `A a(f())` is equivalent to `A a = f()` because here it is construction instead of assignment. I.e.
```
A a = f();  // construction
```
```
A a;
a = f();  // assignment
```
You can check the examples in
http://en.cppreference.com/w/cpp/language/copy_constructor
and 
http://en.cppreference.com/w/cpp/language/copy_assignment

jianqiao:
Style comments addresses except for some of `const` and "deep copy" comments.

query_optimizer/logical/Logical.hpp
zuyu: `A a(f())` could be an implicit `move` constructor, instead of a copy constructor.

zuyu: The goal here is to avoid a deep copy whenever possible. Instead use either `std::move` or a reference.

jianqiao: I think we need to have an agreement on the semantics of `A a = b'. In other words, in your idea what is the output for the following C++ program?

```
#include <vector>
#include <iostream>
using namespace std;

class A {
 public:
  A() {
  }

  A(const A& a) {
    std::cerr << "Copy constructor\n";
  }

  A(A&& a) {
    std::cerr << "Move constructor\n";
  }

  A& operator= (const A& a) {
    std::cerr << "Copy assignment\n";
    return *this;
  }

  A& operator= (A&& a) {
    std::cerr << "Move assignment\n";
    return *this;
  }
};

A f() {
  A a;
  return a;
}

int main(int argc, char *argv[]) {
  A a;

  // 1
  A b = a;

  // 2
  A c = f();

  // 3
  A d = std::move(f());

  return 0;
}
```

zuyu: Alright, you could ignore my comments for this part.

I do know that `A a = f();` is a constructor, but I just learnt that it is actually a `move` constructor.

To answer your question:
```
// 1
Copy constructor

// 2

//3
Move constructor
```

zuyu:
LGTM. We should mark the arguments `const` in another separate PR.

PR #159, opened 2016-04-12
Set up CatalogDatabaseCache proto in optimizer.

zuyu:
This small PR set up the `CatalogDatabaseCache` proto in the optimizer.

Note that it does not affect the single-node version, as we use a macro to ensure that it is only used in the distributed version.

### DIFF
--- a/query_optimizer/ExecutionGenerator.hpp
+++ b/query_optimizer/ExecutionGenerator.hpp
@@ -1,6 +1,6 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
- *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2015-2016 Pivotal Software, Inc.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -21,6 +21,11 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+
+#ifdef QUICKSTEP_DISTRIBUTED
+#include <unordered_set>
+#endif
+
 #include <vector>
 
 #include "catalog/CatalogTypedefs.hpp"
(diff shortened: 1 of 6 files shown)


query_optimizer/QueryHandle.hpp
```diff
@@ -81,6 +82,20 @@ class QueryHandle {
   }
 
   /**
+    * @return The catalog database cache in the protobuf format.
+    */
```
hbdeshmukh: Do we need the ``QUICKSTEP_DISTRIBUTED`` guards around these new functions?

query_optimizer/ExecutionGenerator.hpp
```diff
@@ -368,6 +384,13 @@ class ExecutionGenerator {
   QueryPlan *execution_plan_;  // A part of QueryHandle.
   serialization::QueryContext *query_context_proto_;  // A part of QueryHandle.
 
+#ifdef QUICKSTEP_DISTRIBUTED
+  serialization::CatalogDatabase *catalog_database_cache_proto_;  // A part of QueryHandle.
+
+  // Used to bookkeep relation ids for 'catalog_database_cache_proto_'.
+  std::unordered_set<relation_id> referenced_relation_ids_;
```
hbdeshmukh: I am trying to understand if there's any reason to use ``unordered_set`` over other container choices such as ``vector`` or ``unordered_map``?

hbdeshmukh:
Hi @zuyu 

The PR looks good. I have couple of comments, mostly to clarify things.

query_optimizer/ExecutionGenerator.hpp
zuyu: I try to remove duplicate `relation_id`s, if any, i.e., a self-join query. So `vector` is not a choice. Also, we don't need a key-value map, so no `unordered_map`.

query_optimizer/QueryHandle.hpp
zuyu: We could, but since `serialization::CatalogDatabase` is defined without macros, I think we don't need for now.

We could always add within the macro scope later.

PR #167, opened 2016-04-15
Adds support for IN predicate

jianqiao:
This PR adds support for the `IN` predicate.

There are two use patterns:
(1) `IN` followed by a list of scalar expressions, e.g.
```
SELECT *
FROM R AS R(x, y)
WHERE x IN (1, 2, y+1)
```
In this case, the `IN` predicate will be transformed into a disjunction of equal comparisons. E.g. `x IN (1, 2, y+1)` will be transformed to `(x = 1 OR x = 2 OR x = y+1)`.

(2) `IN` followed by a subquery expression, e.g.
```
SELECT *
FROM R
WHERE x IN (SELECT y FROM S);
```
In this case, the `IN` predicate will be transformed into _hash semi join_ between `R` and `S` on `R.x = S.y` (also _hash anti join_ for `NOT IN`).


parser/ParsePredicateInTableQuery.hpp
```diff
@@ -0,0 +1,110 @@
...
+  std::unique_ptr<ParseSubqueryExpression> table_query_;
+
+  DISALLOW_COPY_AND_ASSIGN(ParsePredicateInTableQuery);
+};
+
+/** @} */
+
+}  // namespace quickstep
+
+#endif /* QUICKSTEP_PARSER_PARSE_PREDICATE_IN_TABLE_QUERY_HPP_ */
```
zuyu: As a convention, we use this style `#endif  // QUICKSTEP_PARSER_PARSE_PREDICATE_IN_TABLE_QUERY_HPP_`.

I know some files in the optimizer code written by @qiangzeng uses the style `/**/`, but for new file I suggest to follow the convention. Thanks!

parser/tests/TPCH.test
```diff
@@ -1302,9 +1657,9 @@ WHERE
 ORDER BY
   s_name
 --
-ERROR: syntax error (8 : 13)
-  s_suppkey in (
-            ^
+ERROR: syntax error (23 : 9)
+        SELECT
```
zuyu: Question: does this parser error result from missing the sub-query support? Thanks!

query_optimizer/expressions/InTableQuery.cpp
```diff
@@ -0,0 +1,55 @@
...
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "query_optimizer/expressions/InTableQuery.hpp"
+
+#include <vector>
+#include <string>
```
zuyu: Please sort the header files in alphabet order.

query_optimizer/expressions/InTableQuery.cpp
```diff
@@ -0,0 +1,55 @@
...
+    std::vector<std::vector<OptimizerTreeBaseNodePtr>> *container_child_fields) const {
+  non_container_child_field_names->push_back("test_expression");
+  non_container_child_fields->push_back(test_expression_);
+
+  non_container_child_field_names->push_back("table_query");
+  non_container_child_fields->push_back(table_query_);
+}
+
+
+
```
zuyu: Please remove two unnecessary empty lines. Thanks!

query_optimizer/expressions/InTableQuery.hpp
```diff
@@ -0,0 +1,143 @@
...
+  DISALLOW_COPY_AND_ASSIGN(InTableQuery);
+};
+
+/** @} */
+
+}  // namespace expressions
+}  // namespace optimizer
+}  // namespace quickstep
+
+#endif /* QUICKSTEP_QUERY_OPTIMIZER_EXPRESSIONS_IN_TABLE_QUERY_HPP_ */
```
zuyu: Ditto for `//` instead of `/**/`.

query_optimizer/expressions/InValueList.cpp
```diff
@@ -0,0 +1,72 @@
...
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "query_optimizer/expressions/InValueList.hpp"
+
+#include <vector>
+#include <string>
```
zuyu: Please sort the header files. Thanks!

query_optimizer/expressions/InValueList.hpp
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * @return Expressions to search for the value of the test_expression.
+   */
+  const std::vector<ScalarPtr>& match_expressions() const {
+    return match_expressions_;
+  }
+
+  ExpressionPtr copyWithNewChildren(
+      const std::vector<ExpressionPtr> &new_children) const override {
+    DCHECK_EQ(new_children.size(), children().size());
```
zuyu: Please swap two operands, as `new_children` should be the checked value. Thanks!

query_optimizer/expressions/InValueList.hpp
```diff
@@ -0,0 +1,158 @@
...
+};
+
+/** @} */
+
+}  // namespace expressions
+}  // namespace optimizer
+}  // namespace quickstep
+
+
+#endif /* QUICKSTEP_QUERY_OPTIMIZER_EXPRESSIONS_IN_VALUE_LIST_HPP_ */
```
zuyu: Please remove a unnecessary empty line above. Also, ditto for `//` instead of `/**/`. Thanks!

query_optimizer/resolver/Resolver.cpp
```diff
@@ -2523,7 +2536,74 @@ E::PredicatePtr Resolver::resolvePredicate(
...
+          resolveExpression(*in_table_query.test_expression(),
+                            nullptr /* type_hint */,
+                            &test_expr_resolution_info);
+      if (test_expr_resolution_info.hasAggregate() && !expression_resolution_info->hasAggregate()) {
+        expression_resolution_info->parse_aggregate_expression =
+            test_expr_resolution_info.parse_aggregate_expression;
+      }
+
+      ExpressionResolutionInfo table_query_resolution_info(*expression_resolution_info);
+      std::vector<const Type*> type_hints = { &test_expression->getValueType() };
```
zuyu: Could we mark `const`?

query_optimizer/rules/UnnestSubqueries.cpp
```diff
@@ -300,6 +302,7 @@ E::ExpressionPtr UnnestSubqueriesForNonRootLogical::eliminateOuterAttributeRefer
     std::vector<E::AttributeReferencePtr> *probe_join_attributes,
     std::vector<E::AttributeReferencePtr> *build_join_attributes,
     std::vector<E::PredicatePtr> *non_hash_join_predicates) {
+  DCHECK(expression->getExpressionType() != E::ExpressionType::kInTableQuery);
```
zuyu: Could we use `DCHECK_NE(expression->getExpressionType(), E::ExpressionType::kInTableQuery);` instead?

The style checker should suggest it whenever possible.

https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint_unittest.py#L1838

FYI, we don't need to consider the order of `checked` and `expected` values for anything but `CHECK_EQ` and `DCHECK_EQ`.

query_optimizer/rules/UnnestSubqueries.hpp
```diff
@@ -180,6 +181,7 @@ class UnnestSubqueriesForExpession : public Rule<expressions::Expression> {
       const expressions::ExpressionPtr &node);
 
   void transformExists(const expressions::Exists &exists_predicate);
+  void transformInTableQuery(const expressions::InTableQuery &in_table_query);
```
zuyu: Please add an empty line between lines `183` and `184` to separate two methods. Also, add `doxygen` comments for them. Thanks!

zuyu:
@jianqiao I've done one pass. There are just minor style comments. Thanks!

pateljm:
LTGM and I can merge once the style and documentation issues that @zuyu points out are fixed.

parser/tests/TPCH.test
jianqiao: Yes. This will be fixed with scalar subquery support.

query_optimizer/rules/UnnestSubqueries.cpp
jianqiao: It appears that DCHECK_NE cannot apply to `enum class` values because it uses the `<<` operator to print out the checked value -- however there is no `<<` overloading for the enum class.

jianqiao:
Just addressed all the comments.

query_optimizer/rules/UnnestSubqueries.cpp
zuyu: OK, we'd better ignore it for the `enum class` case.

zuyu:
LGTM, and will merge once the CI passes.

PR #168, opened 2016-04-15
DO NOT MERGE: Built in Travis using two jobs for all.

zuyu:


hakanmemisoglu:
This one always crashed in gcc debug build. I thing gcc (whether it is 4 or 5) consumes a lot of memory and eventually gives an internal error (as far as I remember in the phase of compilation of EqualityOperator). I tried to remove debug symbols from build but it still gave the same error. I have also tried `ggc-min-heapsize` and `ggc-min-expand`, again it was no luck.

zuyu:
@hakanmemisoglu Thank you for your input. Let's see how it goes. Thanks!

zuyu:
@hakanmemisoglu I see. It's a OOM issue. We have to enlarge the swap capacity.

pateljm:
Thanks for trying @zuyu and @hakanmemisoglu. Really nice work that @hakanmemisoglu and and @zuyu did and we have shaved off 25% time from a week ago. Sweet!

PR #169, opened 2016-04-15
Added hook (not actual code) to BitWeaving.

cramja:
This pull request adds the ability to create a bitweaving index in QS. The actual bitweaving code is in a [private repo](https://github.com/UWQuickstep/bitweaving), and currently can only be accessed by UW students. Instructions for installation can be found there.

This PR adds parser support for the creation of a BitWeaving index. However, without the Index installed, a query to `CREATE INDEX USING bitweaving;` will result in an SQL error message, but will not crash Quickstep. 

If a version of Quickstep which has been compiled with Bitweaving creates bitweaving blocks, and a version of Quickstep without Bitweaving accesses those blocks, a FATALERROR with a proper message will be thrown. I feel that if this is a serious issue (unlikely due to the fact that bitweaving is internal at this point) it can be addressed in another PR.

Instructions:

# BitWeaving
A BitWeaving Index for Quickstep.

This is an optional submodule for Quickstep. It **must be kept private** because of restrictions related to a patent. That said, if you are a member of UWQuickstep, feel free to try it. It installs like any submodule. Just be careful to remove it the correct way (described below) otherwise git will be angry with you.

## Installing
This assumes you have a `build/` directory with quickstep built in it.
```bash
cd quickstep_root

git submodule add https://github.com/UWQuickstep/bitweaving.git storage/bitweaving
# Triggers a rebuild of the CMakefile
touch CMakeLists.txt
cd build && make -j4
```
## Usage
The Bitweaving index can be created like any other index. However, it must be made on a compressed block.
``` SQL
create table foo (id int, name varchar(20)) with
  blockproperties (type compressed_columnstore, compress all, sort id);

create index bwfoo on foo(id) using bitweaving;

copy foo from 'foo.csv' with (delimiter ',');

-- A bitweaving index has now been created on foo.

```

## Uninstalling
The removal process is involved.
``` bash
cd quickstep_root

git submodule deinit -f storage/bitweaving
git rm storage/bitweaving --cached
rm storage/bitweaving -rf
rm -rf .git/modules/storage -rf
git checkout .gitmodules
touch CMakeLists.txt
cd build && make -j4

```


cramja:
This is something I wasn't expecting from the Travis scripts:
```
0.13s$ ./validate_cmakelists.py
Traceback (most recent call last):
  File "./validate_cmakelists.py", line 489, in <module>
    if main(sys.argv[1:]) > 0:
  File "./validate_cmakelists.py", line 469, in main
    process_cmakelists_file(cmakelists_filename, qs_module_dirs))
  File "./validate_cmakelists.py", line 365, in process_cmakelists_file
    generated_targets)
  File "./validate_cmakelists.py", line 249, in process_add_library
    qs_module_dirs))
  File "./validate_cmakelists.py", line 133, in get_dependency_set_from_cpp_src
    with open(src_filename, "r") as src_file:
IOError: [Errno 2] No such file or directory: './storage/bitweaving/BitWeavingIndexSubBlock.cpp'
```

cramja:
The check script is having trouble because, in `storage/CMakeLists.txt` we added a conditional library:
```cmake
if(QUICKSTEP_HAVE_BITWEAVING)
  add_library(quickstep_storage_BitWeavingIndexSubBlock
              bitweaving/BitWeavingIndexSubBlock.cpp
              bitweaving/BitWeavingIndexSubBlock.hpp)
  add_library(quickstep_storage_BitWeavingHIndexSubBlock
              bitweaving/BitWeavingHIndexSubBlock.cpp
              bitweaving/BitWeavingHIndexSubBlock.hpp)
  add_library(quickstep_storage_BitWeavingVIndexSubBlock
              bitweaving/BitWeavingVIndexSubBlock.cpp
              bitweaving/BitWeavingVIndexSubBlock.hpp)
endif()
```
Even if the conditional is not met, it will be checked. 

I guess there's 2 ways to fix this:
1. Try to change the CMake file to move the library creation to a separate CMakeLists.txt file in storage/bitweaving
2. Modify the check script to ignore certain patterns when reading libraries: for example, ignore `*BitWeaving*` regex.

I prefer 2 because even if I modify to remove cmake files, then we will get the following warnings from the `validate` script:
```
Missing target_link_libraries() for quickstep_storage_StorageBlock:
	quickstep_storage_bitweaving_BitWeavingHIndexSubBlock
	quickstep_storage_bitweaving_BitWeavingIndexSubBlock
	quickstep_storage_bitweaving_BitWeavingVIndexSubBlock
Possibly superfluous target_link_libraries() for quickstep_storage_StorageBlock:
	quickstep_storage_BitWeavingHIndexSubBlock
	quickstep_storage_BitWeavingIndexSubBlock
	quickstep_storage_BitWeavingVIndexSubBlock
```

@pateljm What do you think?

pateljm:
@cramja: Agreed -- Go with the second option. Thanks!

cramja:
@pateljm I tried moving the CMake files to bitweaving to see if I could ignore that entire subdirectory (very clean solution). I ran into trouble, so instead I added a new directive to the `validate` script.

In sections of CMake code you'd like to ignore, you can surround them with:
```cmake
# CMAKE_VALIDATE_IGNORE_BEGIN
...
# CMAKE_VALIDATE_IGNORE_END
```

and the `validate` script will ignore everything in the comments and then resume parsing in the previous state as before entering the IGNORE state.

pateljm:
Neat! LTGM. Merging.

Issue #51, opened 2016-02-29
Optimize pre-loading the buffer pool


pateljm:
When starting Quickstep with the `-preload_buffer_pool` option, if the buffer pool size is less than the size of the DB on disk, then stop loading the buffer pool when evictions start. Warn the user about this situation when that happens.

pateljm:
PR #172 closes this issue.

PR #172, opened 2016-04-16
Stop preloading the buffer pool when it is full.

pateljm:
This PR addressed issue #51.

When the database is larger than the buffer pool, a message is printed, and the preloading is stopped saving unnessary IOs. 

Also, the time to preload the buffer pool is now printed. 

Since this is a feature that is largely used for performance testing, no tests are added. To test this feature, create a database that has more than 10 blocks, and try the following: 
`./quickstep_cli_shell -preload_buffer_pool=true` and
`./quickstep_cli_shell -preload_buffer_pool=true -buffer_pool_slots=10`

### DIFF
--- a/cli/QuickstepCli.cpp
+++ b/cli/QuickstepCli.cpp
@@ -169,7 +169,7 @@ int main(int argc, char* argv[]) {
            real_num_workers,
            (static_cast<double>(quickstep::FLAGS_buffer_pool_slots) * quickstep::kSlotSizeBytes)/quickstep::kAGigaByte);
   } else {
-    LOG(FATAL) << "Quickstep needs at least one worker thread";
+    LOG(FATAL) << "Quickstep needs at least one worker thread to run";
   }
 
 #ifdef QUICKSTEP_HAVE_FILE_MANAGER_HDFS
@@ -262,14 +262,19 @@ int main(int argc, char* argv[]) {
       DefaultsConfigurator::GetNumNUMANodesCoveredByWorkers(worker_cpu_affinities);
 
   if (quickstep::FLAGS_preload_buffer_pool) {
+    std::chrono::time_point<std::chrono::steady_clock> preLoadStart, preLoadEnd;
+    preLoadStart = std::chrono::steady_clock::now();
+    printf("Preloading the buffer pool ... ");
     quickstep::PreloaderThread preloader(*query_processor->getDefaultDatabase(),
                                          query_processor->getStorageManager(),
                                          worker_cpu_affinities.front());
-    printf("Preloading buffer pool... ");
+
     fflush(stdout);
     preloader.start();
     preloader.join();
-    printf("DONE\n");
+    preLoadEnd = std::chrono::steady_clock::now();
+    printf("in %g seconds\n",
+           std::chrono::duration<double>(preLoadEnd - preLoadStart).count());
   }
 
   Foreman foreman(&bus,
--- a/cli/QuickstepCli.cpp
+++ b/cli/QuickstepCli.cpp
@@ -262,19 +262,19 @@ int main(int argc, char* argv[]) {
       DefaultsConfigurator::GetNumNUMANodesCoveredByWorkers(worker_cpu_affinities);
 
   if (quickstep::FLAGS_preload_buffer_pool) {
-    std::chrono::time_point<std::chrono::steady_clock> preLoadStart, preLoadEnd;
-    preLoadStart = std::chrono::steady_clock::now();
+    std::chrono::time_point<std::chrono::steady_clock> preload_start, preload_end;
+    preload_start = std::chrono::steady_clock::now();
     printf("Preloading the buffer pool ... ");
+    fflush(stdout);
     quickstep::PreloaderThread preloader(*query_processor->getDefaultDatabase(),
                                          query_processor->getStorageManager(),
                                          worker_cpu_affinities.front());
 
-    fflush(stdout);
     preloader.start();
     preloader.join();
-    preLoadEnd = std::chrono::steady_clock::now();
+    preload_end = std::chrono::steady_clock::now();
     printf("in %g seconds\n",
-           std::chrono::duration<double>(preLoadEnd - preLoadStart).count());
+           std::chrono::duration<double>(preload_end - preload_start).count());
   }
 
   Foreman foreman(&bus,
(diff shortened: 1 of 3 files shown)


cli/QuickstepCli.cpp
```diff
@@ -262,14 +262,19 @@ int main(int argc, char* argv[]) {
       DefaultsConfigurator::GetNumNUMANodesCoveredByWorkers(worker_cpu_affinities);
 
   if (quickstep::FLAGS_preload_buffer_pool) {
+    std::chrono::time_point<std::chrono::steady_clock> preLoadStart, preLoadEnd;
```
zuyu: For variable names, we use the `preload_start` style.

https://google.github.io/styleguide/cppguide.html#Variable_Names

cli/QuickstepCli.cpp
```diff
@@ -262,14 +262,19 @@ int main(int argc, char* argv[]) {
...
+    preLoadStart = std::chrono::steady_clock::now();
+    printf("Preloading the buffer pool ... ");
     quickstep::PreloaderThread preloader(*query_processor->getDefaultDatabase(),
                                          query_processor->getStorageManager(),
                                          worker_cpu_affinities.front());
-    printf("Preloading buffer pool... ");
+
     fflush(stdout);
```
zuyu: You may want to move this line, along with the `printf("Preloading ...")`, so that the output will be printed immediately.

storage/PreloaderThread.cpp
```diff
@@ -34,12 +34,28 @@ void PreloaderThread::run() {
     ThreadUtil::BindToCPU(cpu_id_);
   }
 
+  const std::size_t numBufferPoolSlots = storage_manager_->getMaxBufferPoolSlots();
```
zuyu: Ditto for the variable name `num_buffer_pool_slots`.

storage/PreloaderThread.cpp
```diff
@@ -34,12 +34,28 @@ void PreloaderThread::run() {
...
+  std::size_t blocksLoaded = 0;
```
zuyu: Ditto for the variable names `num_buffer_pool_slots` and `blocks_loaded`.

storage/PreloaderThread.cpp
```diff
@@ -34,12 +34,28 @@ void PreloaderThread::run() {
...
     std::vector<block_id> blocks = relation.getBlocksSnapshot();
     for (block_id current_block_id : blocks) {
-      BlockReference current_block = storage_manager_->getBlock(current_block_id, relation);
+      try {
+        BlockReference current_block = storage_manager_->getBlock(current_block_id, relation);
+      } catch (...) {
+        LOG(INFO) << "Error after loading " << blocksLoaded << "blocks\n";
+        throw;
+      }
+      blocksLoaded++;
```
zuyu: Use `++i` instead for the performance reason.

storage/PreloaderThread.cpp
```diff
@@ -34,12 +34,28 @@ void PreloaderThread::run() {
...
+      if (blocksLoaded == numBufferPoolSlots) {
+        // The buffer pool has filled up. But, some database blocks are not loaded.
+        printf(" The database is larger than the buffer pool. Only %lu blocks were loaded ",
```
zuyu: Please remove the whitespace before `The database`. Also, replace the whitespace with `\n` in the end of the string, to flush the output. Thanks!

storage/PreloaderThread.cpp
```diff
@@ -34,12 +34,28 @@ void PreloaderThread::run() {
...
+               blocksLoaded);
+        return;
+      }
     }
   }
+  printf(" Loaded %lu blocks ", blocksLoaded);
```
zuyu: Please replace the last whitespace with `\n` to flush the output.

Question: are you intended to add the first whitespace to distinguish the outputs belong to `preloading`? If so, I suggest to add one more whitespace. Thanks!

cli/QuickstepCli.cpp
pateljm: Will do.

cli/QuickstepCli.cpp
pateljm: Good point.

storage/PreloaderThread.cpp
pateljm: ok

storage/PreloaderThread.cpp
pateljm: I'm guessing you mean ++blocksLoaded;

storage/PreloaderThread.cpp
pateljm: Need that whitespace as it separates a previous printed part. If you run the code, you will see what I mean.

storage/PreloaderThread.cpp
pateljm: Same as comment above. This is what the intent is.

storage/PreloaderThread.cpp
zuyu: Yes, I use `i` because `blocksLoaded` will be replaced.

storage/PreloaderThread.cpp
zuyu: I see. Then the thing is that `printf` prints out the context only if the buffer is full or encounters `\n`. That's why sometimes we need `fflush`.

Do you want to add a `\n` in the end, or add a `fflush` statement? Thanks!

storage/PreloaderThread.cpp
```diff
@@ -34,12 +34,28 @@ void PreloaderThread::run() {
...
+  std::size_t blocks_loaded = 0;
+
   for (const CatalogRelation &relation : database_) {
     std::vector<block_id> blocks = relation.getBlocksSnapshot();
     for (block_id current_block_id : blocks) {
-      BlockReference current_block = storage_manager_->getBlock(current_block_id, relation);
+      try {
+        BlockReference current_block = storage_manager_->getBlock(current_block_id, relation);
+      } catch (...) {
+        LOG(INFO) << "Error after loading " << blocks_loaded << "blocks\n";
```
zuyu: We may want to use `LOG(ERROR)` instead of `LOG(INFO)`.

pateljm: It's really informational, so INFO is preferred.

zuyu:
LGTM!

Issue #173, opened 2016-04-16
Graceful error handling when loading malformed data.


pateljm:
When loading data using the COPY command if the input file is malformed, then we abort printing the following message: 

```
libc++abi.dylib: terminating with uncaught exception of type quickstep::TextScanFormatError: TextScanFormatError: Malformed data encountered during text scan (Row has too many fields)
```
Handle such errors gracefully by warning the user and aborting the load, and not crashing the system.

cramja:
There are a lot of `throw` statements in the file. I can replace them with SQL_ERRORS which will be caught by the CLI.

cramja:
Oh I see, specifically, the errors defined by Glob.h:

```
    The glob() function shall fail and return the corresponding value if:
    GLOB_ABORTED
        The scan was stopped because GLOB_ERR was set or (*errfunc()) returned non-zero.
    GLOB_NOMATCH
        The pattern does not match any existing pathname, and GLOB_NOCHECK was not set in flags.
    GLOB_NOSPACE
        An attempt to allocate memory failed.
```

From [glob.h](http://pubs.opengroup.org/onlinepubs/009695399/functions/glob.html)

PR #174, opened 2016-04-17
Better error messages for failed unix glob.

cramja:
In relation to #173 this is a small PR aimed at not fully fixing the issue but to better report Unix glob errors.

It also crash the system instead of having the system hang as it does now when a file pattern creates no file matches.


pateljm:
Nice! LTGM. Merging.

PR #175, opened 2016-04-17
Partition aware selection using NUMA-awareness

adalbertgerald:
The select operator has been made partition aware.


adalbertgerald:
Hi @hbdeshmukh, As we discussed, I have made the select operator partition aware. Can you please review this PR?

relational_operators/SelectOperator.cpp
```diff
@@ -54,43 +54,93 @@ bool SelectOperator::getAllWorkOrders(
...
+                input_relation_block_ids_[num_workorders_generated_],
+                predicate,
+                simple_projection_,
+                simple_selection_,
+                selection,
+                output_destination,
+                storage_manager),
+            op_index_);
+        ++num_workorders_generated_;
+      }
```
hbdeshmukh: This function has become a giant function with lots of if/else blocks. Can we refactor it so that the calls when partitioning is present and it is absent are dealt in different functions? That will make the code easier to read.

relational_operators/SelectOperator.cpp
```diff
@@ -54,43 +54,93 @@ bool SelectOperator::getAllWorkOrders(
...
     }
     return done_feeding_input_relation_;
   }
 }
 
 void SelectWorkOrder::execute() {
   BlockReference block(
-      storage_manager_->getBlock(input_block_id_, input_relation_));
+      storage_manager_->getBlock(input_block_id_, input_relation_, getPreferredNUMANodes()[0]));
```
hbdeshmukh: This may segfault if getPreferredNUMANodes() returns an empty vector. Error handling is needed to avoid such a case. Also, adding a note on why we pick only the first NUMA node from the vector should be helpful.

relational_operators/SelectOperator.hpp
```diff
@@ -87,7 +89,26 @@ class SelectOperator : public RelationalOperator {
         num_workorders_generated_(0),
         simple_projection_(false),
         input_relation_is_stored_(input_relation_is_stored),
-        started_(false) {}
+        placement_scheme_(input_relation.getNUMAPlacementScheme()),
+        started_(false) {
+    if (input_relation.hasPartitionScheme()) {
+      const PartitionScheme &part_scheme = input_relation.getPartitionScheme();
+      const PartitionSchemeHeader &part_scheme_header = part_scheme.getPartitionSchemeHeader();
+      int num_partitions = part_scheme_header.getNumPartitions();
```
hbdeshmukh: The type for ``num_partitions`` can be ``std::size_t``, as it will always be non-negative.

relational_operators/SelectOperator.hpp
```diff
@@ -124,7 +145,26 @@ class SelectOperator : public RelationalOperator {
         num_workorders_generated_(0),
         simple_projection_(true),
         input_relation_is_stored_(input_relation_is_stored),
-        started_(false) {}
+        placement_scheme_(input_relation.getNUMAPlacementScheme()),
+        started_(false) {
+    if (input_relation.hasPartitionScheme()) {
+      const PartitionScheme &part_scheme = input_relation.getPartitionScheme();
+      const PartitionSchemeHeader &part_scheme_header = part_scheme.getPartitionSchemeHeader();
+      int num_partitions = part_scheme_header.getNumPartitions();
```
hbdeshmukh: The type for num_partitions can be std::size_t, as it will always be non-negative.

relational_operators/SelectOperator.hpp
```diff
@@ -163,12 +202,15 @@ class SelectOperator : public RelationalOperator {
   const std::vector<attribute_id> simple_selection_;
 
   std::vector<block_id> input_relation_block_ids_;
+  std::vector<std::vector<block_id>> input_relation_block_ids_in_partition_;
```
hbdeshmukh: A description like below should be helpful for the readers:
```
A vector of vectors V where V[i] indicates the list of block IDs of the input relation that belong to the partition i
```

relational_operators/SelectOperator.hpp
```diff
@@ -163,12 +202,15 @@ class SelectOperator : public RelationalOperator {
...
 
   // A single workorder is generated for each block of input relation.
   std::vector<block_id>::size_type num_workorders_generated_;
+  std::vector<std::size_t> num_workorders_generated_in_partition_;
```
hbdeshmukh: Similar description as ``input_relation_block_ids_in_partition_`` should be helpful.

relational_operators/SelectOperator.cpp
adalbertgerald: Sure Harshad! I'll refactor this function.

relational_operators/SelectOperator.hpp
```diff
@@ -205,15 +247,18 @@ class SelectWorkOrder : public WorkOrder {
...
       : input_relation_(input_relation),
         input_block_id_(input_block_id),
         predicate_(predicate),
         simple_projection_(simple_projection),
         simple_selection_(simple_selection),
         selection_(selection),
         output_destination_(DCHECK_NOTNULL(output_destination)),
-        storage_manager_(DCHECK_NOTNULL(storage_manager)) {}
+        storage_manager_(DCHECK_NOTNULL(storage_manager)) {
+    preferred_numa_nodes_.push_back(numa_node);
```
hbdeshmukh: Should we allow negative values for ``numa_node``? 
Also, please use ``numa_node_id`` type defined in ``CatalogTypedefs.hpp``, instead of ``int``.

relational_operators/SelectOperator.cpp
adalbertgerald: Thanks for pointing this out. I'll correct this. 
One related question: Why would we need multiple NUMA nodes for a single workorder? If they are not needed, shall we just have one numa node associated with one workorder?

relational_operators/SelectOperator.hpp
```diff
@@ -241,15 +286,18 @@ class SelectWorkOrder : public WorkOrder {
...
       : input_relation_(input_relation),
         input_block_id_(input_block_id),
         predicate_(predicate),
         simple_projection_(simple_projection),
         simple_selection_(std::move(simple_selection)),
         selection_(selection),
         output_destination_(DCHECK_NOTNULL(output_destination)),
-        storage_manager_(DCHECK_NOTNULL(storage_manager)) {}
+        storage_manager_(DCHECK_NOTNULL(storage_manager)) {
+    preferred_numa_nodes_.push_back(numa_node);
```
hbdeshmukh: Should we allow negative values for ``numa_node``? 
Also, please use ``numa_node_id`` type defined in CatalogTypedefs.hpp, instead of ``int``.

relational_operators/SelectOperator.cpp
hbdeshmukh: You are right that in this class, there will always be a single preferred NUMA node. However for joins, we can have multiple preferred NUMA nodes. Now that I think of it, the note that I mentioned is kind of obvious. You can ignore it.

hbdeshmukh:
Hi @adalbertgerald 

Completed one pass over the code. 
I think, the functions ``feedInputBlock`` and ``feedInputBlocks`` should also be made partition aware, right? I didn't see changes in those functions.

relational_operators/SelectOperator.cpp
```diff
@@ -54,43 +54,93 @@ bool SelectOperator::getAllWorkOrders(
...
-                                simple_projection_,
-                                simple_selection_,
-                                selection,
-                                output_destination,
-                                storage_manager),
-            op_index_);
+      if (input_relation_.hasPartitionScheme()) {
+        const PartitionScheme &part_scheme =
+            input_relation_.getPartitionScheme();
+        int num_partitions = part_scheme.getPartitionSchemeHeader().getNumPartitions();
```
zuyu: Please mark `num_partitions  const`.

Actually, we could refactor it as the following, since you don't use `part_scheme` elsewhere.

```
  const std::size_t num_partitions =
      input_relation_.getPartitionScheme().getPartitionSchemeHeader().getNumPartitions();
```

relational_operators/SelectOperator.hpp
```diff
@@ -87,7 +89,26 @@ class SelectOperator : public RelationalOperator {
         num_workorders_generated_(0),
         simple_projection_(false),
         input_relation_is_stored_(input_relation_is_stored),
-        started_(false) {}
+        placement_scheme_(input_relation.getNUMAPlacementScheme()),
```
zuyu: @adalbertgerald This method is valid only if `QUICKSTEP_HAVE_LIBNUMA`. Therefore, we need to use `#ifdef` for this statement as well.

relational_operators/SelectOperator.hpp
```diff
@@ -163,12 +202,15 @@ class SelectOperator : public RelationalOperator {
...
   std::vector<block_id> input_relation_block_ids_;
+  std::vector<std::vector<block_id>> input_relation_block_ids_in_partition_;
 
   // A single workorder is generated for each block of input relation.
   std::vector<block_id>::size_type num_workorders_generated_;
+  std::vector<std::size_t> num_workorders_generated_in_partition_;
 
   const bool simple_projection_;
   const bool input_relation_is_stored_;
+  const NUMAPlacementScheme &placement_scheme_;
```
zuyu: We need to add `QUICKSTEP_HAVE_LIBNUMA` around this statement:

```
#ifdef QUICKSTEP_HAVE_LIBNUMA
  const NUMAPlacementScheme &placement_scheme_;
#endif
```

relational_operators/SelectOperator.hpp
```diff
@@ -241,15 +286,18 @@ class SelectWorkOrder : public WorkOrder {
                   std::vector<attribute_id> &&simple_selection,
                   const std::vector<std::unique_ptr<const Scalar>> *selection,
                   InsertDestination *output_destination,
-                  StorageManager *storage_manager)
+                  StorageManager *storage_manager,
+                  const int numa_node = 0)
```
zuyu: @adalbertgerald In the initial distributed version, we don't have NUMA support. So, please remove this for now. Thanks!

zuyu:
@adalbertgerald Do you know the existing unit tests cover the changes you made in this PR? Also, the CI builds have the NUMA support, have you tried to build it on a machine without `libnuma`? Thanks.

relational_operators/SelectOperator.cpp
adalbertgerald: I just noticed that getPreferredNUMANodes() will not return an empty vector for SelectOperator since we explicitly add a NUMA node in the SelectOperator's constructor. So, I think we need make this check. A default NUMA node hint of zero will be added as the preferred_numa_node_.

relational_operators/SelectOperator.hpp
adalbertgerald: @zuyu the purpose of this PR is to make the select operator partition aware. i.e. The select work units will be executed on the NUMA nodes where the partitions are present. So, if we remove this, then this PR would have no effect.

adalbertgerald:
Hi @hbdeshmukh, I have addressed the first set of review comments. Can you please take a look?

relational_operators/CMakeLists.txt
```diff
@@ -263,6 +263,8 @@ target_link_libraries(quickstep_relationaloperators_SelectOperator
                       glog
                       quickstep_catalog_CatalogRelation
                       quickstep_catalog_CatalogTypedefs
+                      quickstep_catalog_NUMAPlacementScheme
```
zuyu: @adalbertgerald Please add the macro `QUICKSTEP_HAVE_LIBNUMA` around `quickstep_catalog_NUMAPlacementScheme`.

zuyu:
@adalbertgerald I've created a dedicate PR to test your code w/o `libnuma`. Please fix the build errors before merging this PR. Thanks!

relational_operators/SelectOperator.hpp
```diff
@@ -135,13 +175,28 @@ class SelectOperator : public RelationalOperator {
...
+    }
   }
 
   void feedInputBlocks(const relation_id rel_id, std::vector<block_id> *partially_filled_blocks) override {
-    input_relation_block_ids_.insert(input_relation_block_ids_.end(),
-                                     partially_filled_blocks->begin(),
-                                     partially_filled_blocks->end());
+    if (input_relation_.hasPartitionScheme() && input_relation_.hasNUMAPlacementScheme()) {
+      const partition_id part_id =
+          input_relation_.getPartitionScheme().getPartitionForBlock((*partially_filled_blocks)[0]);
```
hbdeshmukh: It seems that you found the partition ID for only the first partially filled block. What about the partition IDs of the remaining blocks? Do you assume that it will be the same for the rest of them? I am not sure if I follow that assumption.

hbdeshmukh: Also, if partially_filled_blocks is empty, line 190 may segfault. Even though this situation is rare, we should safeguard it.

adalbertgerald:
Hi @hbdeshmukh, I have addressed your comment about feedInputBlocks and also made sure that the code works properly on non-NUMA boxes. Can you please take another look?

relational_operators/SelectOperator.cpp
```diff
@@ -35,6 +35,93 @@ namespace quickstep {
...
+  } else {
+    for (std::size_t part_id = 0; part_id < num_partitions; ++part_id) {
+      while (num_workorders_generated_in_partition_[part_id] <
+             input_relation_block_ids_in_partition_[part_id].size()) {
+        container->addNormalWorkOrder(
+            new SelectWorkOrder(
+                input_relation_,
+                input_relation_block_ids_in_partition_
+                    [part_id]
+                    [num_workorders_generated_in_partition_[part_id]],
```
hbdeshmukh: The term ``input_relation_block_ids_in_partition_[part_id][num_workorders_generated_in_partition_[part_id]]``, is kind of difficult to read. Can you please store it to another variable before the ``addNormalWorkOrder`` call?

relational_operators/SelectOperator.hpp
```diff
@@ -87,7 +93,28 @@ class SelectOperator : public RelationalOperator {
         num_workorders_generated_(0),
         simple_projection_(false),
         input_relation_is_stored_(input_relation_is_stored),
-        started_(false) {}
+        started_(false) {
+#ifdef QUICKSTEP_HAVE_LIBNUMA
+    placement_scheme_ = input_relation.getNUMAPlacementSchemePtr();
```
hbdeshmukh: Can we have a DCHECK to ensure that ``placement_scheme_`` is not null?

relational_operators/SelectOperator.hpp
```diff
@@ -124,7 +151,28 @@ class SelectOperator : public RelationalOperator {
         num_workorders_generated_(0),
         simple_projection_(true),
         input_relation_is_stored_(input_relation_is_stored),
-        started_(false) {}
+        started_(false) {
+#ifdef QUICKSTEP_HAVE_LIBNUMA
+    placement_scheme_ = input_relation.getNUMAPlacementSchemePtr();
```
hbdeshmukh: Can we have a DCHECK to ensure that ``placement_scheme_`` is not null?

relational_operators/SelectOperator.hpp
```diff
@@ -135,13 +183,27 @@ class SelectOperator : public RelationalOperator {
...
   }
 
   void feedInputBlocks(const relation_id rel_id, std::vector<block_id> *partially_filled_blocks) override {
-    input_relation_block_ids_.insert(input_relation_block_ids_.end(),
-                                     partially_filled_blocks->begin(),
-                                     partially_filled_blocks->end());
+    if (input_relation_.hasPartitionScheme()) {
+      for (auto it = partially_filled_blocks->begin(); it != partially_filled_blocks->end(); ++it) {
+        const partition_id part_id = input_relation_.getPartitionScheme().getPartitionForBlock((*it));
+        input_relation_block_ids_in_partition_[part_id].insert(input_relation_block_ids_in_partition_[part_id].end(),
```
hbdeshmukh: I think there is a potential performance optimization possible here. Each call to ``getPartitionForBlock()`` involves grabbing shared locks on each partition's mutex, checking if the block belongs to the partition. Instead, we can provide a method ``getPartitionsForBlocks()`` which accepts a list of blocks and returns corresponding list of their partition IDs. Therefore, once we grab a lock for a partition, we search for all the blocks and then release the lock. 

I think It is fine to leave this suggestion as a TODO for now, but we may have to implement it if performance of this function is an issue.

hbdeshmukh:
Hi @adalbertgerald 

I have given my last set of comments. We can merge the PR once you address them.

adalbertgerald:
Hi @hbdeshmukh, I have addressed your latest comments. Please take a look.

hbdeshmukh:
LGTM, merging.

PR #178, opened 2016-04-20
Adds support for left/right outer join

jianqiao:
This PR adds both optimizer and backend support for left/right outer joins. Also adds SQL tests for `resolver`/`logical_generator`/`physical_generator`/`execution_generator` tests.

### DIFF
--- a/relational_operators/WorkOrder.proto
+++ b/relational_operators/WorkOrder.proto
@@ -29,20 +29,21 @@ enum WorkOrderType {
   DESTROY_HASH = 6;
   DROP_TABLE = 7;
   FINALIZE_AGGREGATION = 8;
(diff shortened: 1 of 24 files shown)


pateljm:
LTGM. Merging.

relational_operators/WorkOrder.proto
```diff
@@ -151,6 +152,21 @@ message HashSemiJoinWorkOrder {
   }
 }
 
+message HashOuterJoinWorkOrder {
+  extend WorkOrder {
+    // All required.
+    optional int32 build_relation_id = 370;
```
zuyu: Please recompute the field assignments in this file!

PR #179, opened 2016-04-20
Deleted Blobs After Use in Parallel Loading.

zuyu:
This PR fixed the memory leak bug in a parallel load case where `TextScanWorkOrder` does not clean up the blobs produced by `TextSplitWorkOrder`.

### DIFF
--- a/relational_operators/TextScanOperator.cpp
+++ b/relational_operators/TextScanOperator.cpp
@@ -288,7 +288,7 @@ void TextScanWorkOrder::execute() {
 
     std::fclose(file);
   } else {
-    BlobReference blob = storage_manager_->getBlob(text_blob_);
+    MutableBlobReference blob = storage_manager_->getBlobMutable(text_blob_);
     const char *blob_pos = static_cast<const char*>(blob->getMemory());
     const char *blob_end = blob_pos + text_size_;
     bool have_row = false;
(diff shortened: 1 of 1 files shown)


relational_operators/TextScanOperator.cpp
```diff
@@ -288,7 +288,7 @@ void TextScanWorkOrder::execute() {
 
     std::fclose(file);
   } else {
-    BlobReference blob = storage_manager_->getBlob(text_blob_);
+    MutableBlobReference blob = storage_manager_->getBlobMutable(text_blob_);
```
pateljm: I get the code below. Not sure about why we also need to change to MutableBlobReference. I'm sure there is a reason, and would be good to state that here.

zuyu: Because the `release` method below modifies `blob` by doing some cleanups, i.e., destroying the `CountedReference`.

pateljm: Got it. Thanks!

pateljm:
LGTM. Merging.

relational_operators/TextScanOperator.cpp
zuyu: Actually, you're right. Although `release` is not a `const` method, as long as `blob` is not `const`, we are fine.

In other words, we don't need to change to `MutableBlobReference`. I'll fix it in another PR.

PR #180, opened 2016-04-21
DO NOT MERGE: Deadlock Fix Approach 1

zuyu:
In shared, sharded buffer pool, to prevent deadlock each thread holds at most one exclusive lock on a shard, instead of two exclusive locks (one on the loading shard, one on the eviction shard).

Cons: under-eviction (we evict less slots than it should).


zuyu:
Replaced by #183.

PR #182, opened 2016-04-21
Removed optional bitweaving submodule from submodule file

cramja:
This was an oversight from #169 Adding the bitweaving index repo to the submodule file will be confusing to users who do not have access to the bitweaving repo.


zuyu:
LGTM!

PR #183, opened 2016-04-21
Fixed deadlocks when loading a block or blob while evicting another.

zuyu:
In the shared, sharded buffer pool, prevent deadlocks by avoid holding an exclusive lock (for an eviction) on the same shard on which already has an exclusive lock (for a load from disk).


storage/StorageManager.cpp
```diff
@@ -517,33 +506,16 @@ MutableBlockReference StorageManager::getBlockInternal(
...
-    //   "eviction_lock(*lock_manager_.get(block_index))"
-    //   This situation could lead to a self-deadlock as block_index could
-    //   hash to the same position in the "ShardedLockManager" as "block."
-    //   To deal with this case, we pass the block information for "block"
-    //   though the call chain, and check for a collision in the the
-    //   "ShardedLockManager" in the function "makeRoomForBlock."
-    //   If a collision is detected we avoid a self-deadlock.
     ret = MutableBlockReference(loadBlock(block, relation, numa_node), eviction_policy_.get());
   }
+  // To be safe, release the shard for the block after its write lock destructs.
```
pateljm: Would it be clearer to replace "shard" with "lock held in the sharded hash table?"

storage/StorageManager.cpp
```diff
@@ -556,10 +528,12 @@ MutableBlobReference StorageManager::getBlobInternal(const block_id blob,
     SpinSharedMutexSharedLock<false> read_lock(blocks_shared_mutex_);
     std::unordered_map<block_id, BlockHandle>::iterator it = blocks_.find(blob);
     if (it != blocks_.end()) {
-      DCHECK(it->second.block->isBlob());
+      DEBUG_ASSERT(it->second.block->isBlob());
       ret = MutableBlobReference(static_cast<StorageBlob*>(it->second.block), eviction_policy_.get());
     }
   }
+  // To be safe, release the shard for the blob after its read lock destructs.
```
pateljm: Same comment: Would it be clearer to replace "shard" with "lock held in the sharded hash table?"

storage/StorageManager.cpp
```diff
@@ -487,23 +481,18 @@ MutableBlockReference StorageManager::getBlockInternal(
     const block_id block,
```
pateljm: Given the complexity of this method, could we have a comment for this whole function to explain the flow?

pateljm:
Excellent work @zuyu! See minor comments above.

storage/StorageManager.cpp
zuyu: I think I may need to revise the comment, but not the same as you proposed.

The new `release` method in `SharedLockManager` has nothing to do with the locks held, but removes the referenced shard from a set of shard references (used to detect collisions) so that `makeRoomForBlock` could evict a block from the same shard, if chosen.

cramja:
LGTM

PS For record keeping, here is a [gist containing the test code](https://gist.github.com/cramja/3b4d9683d95b80e1852402f81978bbe6) I used, in case we ever need to come back to this (and I have deleted the branches on github containing it).

zuyu:
@cramja We have a list of tasks to do followed by this fix, including function renaming. Could you please lead on that efforts? Thanks!

storage/StorageManager.cpp
zuyu: Most changes here was a part of reverting a merged PR. We could add comments in a separate PR.

pateljm:
LGTM. Merging.

cramja:
@zuyu Yes. I will create a PR today for this.

zuyu:
@cramja Thanks. I've done one for `makeRoomForBlock` in PR #192.

PR #184, opened 2016-04-21
Honor the buffer pool slots flag.

zuyu:
This PR honored the buffer pool slots flag in #146.


pateljm:
Nice catch @zuyu. LGTM. Merging.

PR #189, opened 2016-04-25
Fixed the lock problem with AggregationHandleSum::mergeStates()

jianqiao:
This PR is a one-line fix to the lock problem with `AggregationHandleSum::mergeStates()`. The fix is that we create a named variable for `SpinMutexLock`. The original line of code is somehow a typo that the lock is non-named and would be released immediately after creation, and causes non-deterministic results for TPC-H query 06.


pateljm:
LGTM. Merging.

PR #190, opened 2016-04-25
Introducing PolicyEnforcer and long lived Foreman thread

hbdeshmukh:
This PR introduces basic skeleton of PolicyEnforcer and makes the Foreman thread understand concurrent queries, as well as allows the Foreman thread to live throughout the lifetime of the Quickstep process. Some key highlights are described below: 
- Creation of a sekeleton PolicyEnforcer class.
- Foreman (C++ class) is now primarily responsible for dispatching work orders to workers and collect messages coming from worker threads during query execution. 
- PolicyEnforcer can talk to multiple QueryManager instances to collect WorkOrders to be dispatched. 
- PolicyEnforcer receives messages from Foreman, figures out the recipient of the message (using query ID in the message) and relays it to the appropriate QueryManager instance. 
- PolicyEnforcer can also perform admission control. 
- Foreman thread now doesn't die after receipt of a new query, unlike earlier. Instead it stays alive throughout the lifetime of Quickstep process. 
- The main thread is now capable of sending and receiving messages through TMB. 
- A new query sent via command line is sent from the main thread to the Foreman thread via a TMB message. 
- Moving some commonly used code in utility class.


cli/QuickstepCli.cpp
```diff
@@ -374,14 +385,20 @@ int main(int argc, char* argv[]) {
...
+                                           foreman.getBusClientID(),
+                                           std::move(admit_tagged_message));
         try {
-          start = std::chrono::steady_clock::now();
-          foreman.start();
-          foreman.join();
+          const AnnotatedMessage annotated_msg =
+              bus.Receive(main_thread_client_id, 0, true);
+          const TaggedMessage &tagged_message = annotated_msg.tagged_message;
+          DCHECK(tagged_message.message_type() == kWorkloadCompletionMessage);
```
zuyu: How about using `DCHECK_EQ`?

cli/tests/CommandExecutorTestRunner.cpp
```diff
@@ -97,14 +102,20 @@ void CommandExecutorTestRunner::runTestCase(
...
+          QueryExecutionUtil::SendTMBMessage(&bus_,
+                                             main_thread_client_id_,
+                                             foreman_->getBusClientID(),
+                                             std::move(admit_tagged_message));
+
+          // Receive workload completion message from Foreman.
+          const AnnotatedMessage annotated_msg =
+              bus_.Receive(main_thread_client_id_, 0, true);
+          const TaggedMessage &tagged_message = annotated_msg.tagged_message;
+          DCHECK(tagged_message.message_type() == kWorkloadCompletionMessage);
```
zuyu: How about using `DCHECK_EQ`?

query_execution/AdmitRequestMessage.hpp
```diff
@@ -0,0 +1,74 @@
...
+   **/
+  explicit AdmitRequestMessage(const std::vector<QueryHandle*> &query_handles)
+      : query_handles_(query_handles) {}
+
+  /**
+   * @brief Constructor for requesting single query admission.
+   *
+   * @param query_handle The handle of the query requesting to be admitted.
+   **/
+  explicit AdmitRequestMessage(QueryHandle* query_handle) {
```
zuyu: According to the code style, `*` should be near `query_handle`. Thanks!

query_execution/AdmitRequestMessage.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+  DISALLOW_COPY_AND_ASSIGN(AdmitRequestMessage);
+};
+
+/** @} */
+
+}  // namespace quickstep
+
+#endif  // QUICKSTEP_QUERY_EXECUTION_ADMIT_REQUEST_MESSAGE_HPP_
+
```
zuyu: Please remove this empty line!

query_execution/CMakeLists.txt
```diff
@@ -40,37 +42,38 @@ add_library(quickstep_queryexecution_WorkerMessage ../empty_src.cpp WorkerMessag
 add_library(quickstep_queryexecution_WorkerSelectionPolicy ../empty_src.cpp WorkerSelectionPolicy.hpp)
 
 # Link dependencies:
+target_link_libraries(quickstep_queryexecution_ForemanLite
+                      quickstep_utility_Macros)
```
zuyu: I was wondering why moving `ForemanLite`?

query_execution/CMakeLists.txt
```diff
@@ -40,37 +42,38 @@ add_library(quickstep_queryexecution_WorkerMessage ../empty_src.cpp WorkerMessag
...
                       tmb)
 target_link_libraries(quickstep_queryexecution_ForemanLite
                       glog
                       quickstep_threading_Thread
                       quickstep_utility_Macros
                       tmb)
+target_link_libraries(quickstep_queryexecution_PolicyEnforcer
+                      glog
+                      quickstep_queryexecution_QueryExecutionTypedefs
+                      quickstep_queryexecution_QueryExecutionMessages_proto
```
zuyu: Please arrange the above two according to the alphabet order.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
 using std::vector;
 
 namespace quickstep {
 
-void Foreman::initialize() {
+Foreman::Foreman(const tmb::client_id main_thread_client_id,
+                 WorkerDirectory *worker_directory,
+                 tmb::MessageBus *bus,
+                 CatalogDatabaseLite *catalog_database,
+                 StorageManager *storage_manager,
```
zuyu: I was wondering if we could arrange the four pointers above in some order, i.e., in the alphabet or else?

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
+                 const int cpu_id,
+                 const size_t num_numa_nodes)
+    : ForemanLite(bus, cpu_id),
+      main_thread_client_id_(main_thread_client_id),
+      worker_directory_(DCHECK_NOTNULL(worker_directory)),
+      catalog_database_(DCHECK_NOTNULL(catalog_database)),
+      storage_manager_(DCHECK_NOTNULL(storage_manager)),
+      min_load_per_worker_(2) {  // TODO(harshad) - Make this field configurable.
+  std::vector<QueryExecutionMessageType> sender_message_types{
```
zuyu: Please mark `const`.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
+      kPoisonMessage,
+      kRebuildWorkOrderMessage,
+      kWorkOrderMessage,
+      kWorkloadCompletionMessage};
+
+  for (auto message_type : sender_message_types) {
```
zuyu: Please add `const` as following:

```
  for (const auto message_type : sender_message_types) {
```

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
+    bus_->RegisterClientAsSender(foreman_client_id_, message_type);
+  }
+
+  std::vector<QueryExecutionMessageType> receiver_message_types{
```
zuyu: Please mark `const`.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
+      kPoisonMessage,
+      kRebuildWorkOrderCompleteMessage,
+      kWorkOrderFeedbackMessage,
+      kWorkOrdersAvailableMessage,
+      kWorkOrderCompleteMessage};
+
+  for (auto message_type : receiver_message_types) {
+    bus_->RegisterClientAsReceiver(foreman_client_id_, message_type);
+  }
+  policy_enforcer_.reset(new PolicyEnforcer(
```
zuyu: Please add an empty line before resetting `policy_enforcer_`.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
+      kAdmitRequestMessage,
+      kCatalogRelationNewBlockMessage,
+      kDataPipelineMessage,
+      kPoisonMessage,
+      kRebuildWorkOrderCompleteMessage,
+      kWorkOrderFeedbackMessage,
+      kWorkOrdersAvailableMessage,
+      kWorkOrderCompleteMessage};
+
+  for (auto message_type : receiver_message_types) {
```
zuyu: Please add `const`, similarly to `sender_message_types`.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
-        break;
-      }
-      case kRebuildWorkOrderCompleteMessage: {
-        serialization::WorkOrderCompletionMessage proto;
-        CHECK(proto.ParseFromArray(tagged_message.message(), tagged_message.message_bytes()));
-
-        processRebuildWorkOrderCompleteMessage(proto.operator_index(), proto.worker_thread_index());
+    const tmb::message_type_id message_type = tagged_message.message_type();
+    switch (message_type) {
+      case kCatalogRelationNewBlockMessage:
```
zuyu: Please add `  // Fall through` in the end. Thanks!

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
-        sendWorkerMessage(curr_worker, *msg);
-        workers_->incrementNumQueuedWorkOrders(curr_worker);
-      } else {
-        // No schedulable workorder at this point.
-        ++done_workers_count;
-      }
-    } else {
-      // curr_worker already has been assigned max_msgs_per_worker workorders.
-      ++done_workers_count;
+    if (canCollectNewMessages(message_type)) {
```
zuyu: Please add an empty line before the `if` statement.

query_execution/QueryExecutionTypedefs.hpp
```diff
@@ -65,7 +65,9 @@ enum QueryExecutionMessageType : message_type_id {
                               // their corresponding RelationalOperators.
   kRebuildWorkOrderMessage,  // From Foreman to Worker.
   kRebuildWorkOrderCompleteMessage,  // From Worker to Foreman.
-  kPoisonMessage,  // From the CLI shell to Foreman, then from Foreman to Workers.
+  kPoisonMessage,  // From the main thread to Foreman and workers.
+  kAdmitRequestMessage,  // Requesting a query (or queries) to be admitted.
+  kWorkloadCompletionMessage  // From Foreman to main thread.
```
zuyu: FYI, I sort these message types in the order of query life cycle. So, I think `kAdmitRequestMessage` could be in the first item, and `kWorkloadCompletionMessage` could put between `kRebuildWorkOrderCompleteMessage` and `kPoisonMessage`. Thoughts?

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
-void Foreman::run() {
-  // Initialize before for Foreman eventloop.
-  initialize();
 
   // Event loop
-  while (!query_exec_state_->hasQueryExecutionFinished()) {
+  for (;;) {
     // Receive() causes this thread to sleep until next message is received.
-    AnnotatedMessage annotated_msg = bus_->Receive(foreman_client_id_, 0, true);
+    AnnotatedMessage annotated_msg =
```
zuyu: Please mark `const`.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
-      }
-    } else {
-      // curr_worker already has been assigned max_msgs_per_worker workorders.
-      ++done_workers_count;
+    if (canCollectNewMessages(message_type)) {
+      vector<unique_ptr<WorkerMessage>> new_messages;
+      policy_enforcer_->getWorkerMessages(&new_messages);
+      dispatchWorkerMessages(&new_messages);
+    }
+    // We check again, as some queries may produce zero work orders and finish
```
zuyu: I suggest to add one empty line before the comment.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
+  if (QUICKSTEP_EQUALS_ANY_CONSTANT(message_type,
+                                    kCatalogRelationNewBlockMessage,
+                                    kWorkOrderFeedbackMessage)) {
+    return false;
+  } else if (worker_directory_->getLeastLoadedWorker().second <= min_load_per_worker_) {
+    // If the least loaded worker has only one pending work order, we should
+    // collect new messages and dispatch them.
+    return true;
+  } else {
+    return false;
```
zuyu: How about this following refactor for the `else-if-else` part:

```
  return worker_directory_->getLeastLoadedWorker().second <= min_load_per_worker_;
```

Actually, I also suggest to have a new function called `HasAnyAvaliableWorker` to wrap the expression.

query_execution/Foreman.cpp
```diff
@@ -47,330 +37,168 @@
...
-    // Either no workorder found on the given NUMA node, or numa_node is -1.
-    // Try to get a normal WorkOrder from other NUMA nodes.
-    work_order = workorders_container_->getNormalWorkOrder(index);
-    if (work_order != nullptr) {
-      query_exec_state_->incrementNumQueuedWorkOrders(index);
-      return WorkerMessage::WorkOrderMessage(work_order, index);
+void Foreman::dispatchWorkerMessages(vector<unique_ptr<WorkerMessage>> *messages) {
+  for (auto message_it = messages->begin();
+       message_it != messages->end();
+       ++message_it) {
```
zuyu: I was wondering why not use `for (const auto &message : *messages) {`.

Also, have you modified `messages` in this method? If not, why not use a `const` reference? Thanks.

query_execution/Foreman.hpp
```diff
@@ -22,46 +22,36 @@
...
 
 /** \addtogroup QueryExecution
  *  @{
  */
 
 /**
- * @brief The Foreman scans the query DAG, requests each operator to produce
- *        workorders. It also pipelines the intermediate output it receives to
- *        the relational operators which need it.
+ * @brief The Foreman receives queries from the main thread, messages from the
```
zuyu: Please add comments to state that this `Foreman` is for the single-node version only. Thanks!

query_execution/Foreman.hpp
```diff
@@ -71,233 +61,38 @@ class Foreman final : public ForemanLite {
...
-
-  /**
-   * @brief Set the maximum number of messages that should be allocated to each
-   *        worker during a single round of WorkOrder dispatch.
-   *
-   * @param max_msgs_per_worker Maximum number of messages.
-   **/
-  void setMaxMessagesPerWorker(const std::size_t max_msgs_per_worker) {
-    max_msgs_per_worker_ = max_msgs_per_worker;
+  tmb::client_id getBusClientID() const {
```
zuyu: `ForemanLite` has this function, so please remove it.

query_execution/Foreman.hpp
```diff
@@ -306,156 +101,27 @@ class Foreman final : public ForemanLite {
...
-
-  const int num_numa_nodes_;
-
-  WorkerDirectory *workers_;
+  std::unique_ptr<PolicyEnforcer> policy_enforcer_;
 
-  friend class ForemanTest;
-  FRIEND_TEST(ForemanTest, TwoNodesDAGPartiallyFilledBlocksTest);
+  // Every worker should have at least these many pending work orders.
+  const std::size_t min_load_per_worker_;
```
zuyu: Since you'd like `min_load_per_worker_` to be configurable, how about adding a `gflag` for this purpose?

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+  if (admitted_queries_.size() < kMaxConcurrentQueries) {
+    // Ok to admit the query.
+    const std::size_t query_id = query_handle->query_id();
+    if (admitted_queries_.find(query_id) == admitted_queries_.end()) {
+      admitted_queries_[query_id].reset(
+          new QueryManager(foreman_client_id_, num_numa_nodes_, query_handle,
+                           catalog_database_, storage_manager_, bus_));
+      return true;
+    } else {
+      LOG(ERROR) << "Query with same ID " << query_id << " exists";
```
zuyu: Gramma: `Query with the same ID`

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+  std::size_t query_id;
+  switch (tagged_message.message_type()) {
+    case kWorkOrderCompleteMessage: {
+      serialization::WorkOrderCompletionMessage proto;
+      CHECK(proto.ParseFromArray(tagged_message.message(),
+                                 tagged_message.message_bytes()));
+      query_id = proto.query_id();
+      break;
+    }
+    case kRebuildWorkOrderCompleteMessage: {
```
zuyu: We could combine the `kRebuildWorkOrderCompleteMessage` case with `kWorkOrderCompleteMessage`, right?

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+    }
+    case kWorkOrderFeedbackMessage: {
+      // TODO(harshad) Add query ID to FeedbackMessage.
+      WorkOrder::FeedbackMessage msg(const_cast<void *>(tagged_message.message()), tagged_message.message_bytes());
+      query_id = msg.header().query_id;
+      break;
+    }
+    default:
+      LOG(FATAL) << "Unknown message type found in PolicyEnforcer";
+  }
```
zuyu: Do you think it is a waste to serialize a proto only for `query_id`? We could take advantages of `QueryManager::processMessage` to get `query_id`, while postpone the `DCHECK` in Line 106 later.

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+  DCHECK(admitted_queries_.find(query_id) != admitted_queries_.end());
+  QueryManager::QueryStatusCode return_code =
```
zuyu: Please mark `const`.

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+  } else {
+    LOG(WARNING) << "Requesting WorkerMessages when no query is running";
+    return;
+  }
+  DCHECK_GT(per_query_share, 0u);
+  std::vector<std::size_t> finished_queries_ids;
+
+  for (auto query_manager_it = admitted_queries_.begin();
+       query_manager_it != admitted_queries_.end();
+       ++query_manager_it) {
```
zuyu: Could we use `for (const auto &admitted_query_info : admitted_queries_)`?

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+    QueryManager *curr_query_manager = query_manager_it->second.get();
+    DCHECK(curr_query_manager != nullptr);
+    std::size_t messages_collected_curr_query = 0;
+    while (messages_collected_curr_query < per_query_share) {
```
zuyu: Q: are we trying to implement a fair-based scheduling policy here?

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+void PolicyEnforcer::removeQuery(const std::size_t query_id) {
+  DCHECK(admitted_queries_.find(query_id) != admitted_queries_.end());
+  if (!admitted_queries_[query_id]->getQueryExecutionState().hasQueryExecutionFinished()) {
+    LOG(WARNING) << "Removing query with ID " << query_id
+                 << " that hasn't finished its execution";
+  }
+  admitted_queries_[query_id].reset(nullptr);
+  admitted_queries_.erase(query_id);
```
zuyu: We don't need to call `reset` in Line 172, as an `erase` automatically destroys the unique pointer.

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+bool PolicyEnforcer::admitQueries(
+    const std::vector<QueryHandle *> &query_handles) {
+  bool result = true;
+  for (QueryHandle* curr_query : query_handles) {
+    result = result && admitQuery(curr_query);
+  }
+  return result;
```
zuyu: Q: if `result` has been set to `false`, do you intend not to accept any more queries?

If yes, I have the following refactor:
```
  for (const QueryHandle *query_handle : query_handles) {
    if (!admitQuery(query_handle)) {
      return false;
    }
  }
  return true;
```

Otherwise, I have another refactor:
```
  bool result = true;
  for (const QueryHandle *query_handle : query_handles) {
    if (!admitQuery(query_handle) && result) {
      result = false;
    }
  }
  return result;
```

query_execution/QueryContext.proto
```diff
@@ -52,4 +52,6 @@ message QueryContext {
 
   // NOTE(zuyu): For UpdateWorkOrder only.
   repeated UpdateGroup update_groups = 9;
+
+  required uint64 query_id = 10;
```
zuyu: You may want to set `query_id` to `1`, and reset all others.

query_execution/QueryExecutionMessages.proto
```diff
@@ -31,14 +32,17 @@ message CatalogRelationNewBlockMessage {
...
 message DataPipelineMessage {
   required uint64 operator_index = 1;
   required fixed64 block_id = 2;
   required int32 relation_id = 3;
+  required uint64 query_id = 4;
 }
 
 message WorkOrdersAvailableMessage {
   required uint64 operator_index = 1;
+  required uint64 query_id = 2;
```
zuyu: I suggest to set `query_id` to `1` for all the cases, as this entry will not change in future.

query_execution/QueryExecutionUtil.hpp
```diff
@@ -60,6 +62,26 @@ class QueryExecutionUtil {
...
+    address.All(true);
+    std::unique_ptr<WorkerMessage> poison_message(WorkerMessage::PoisonMessage());
+    TaggedMessage poison_tagged_message(poison_message.get(),
+                                        sizeof(*poison_message),
+                                        kPoisonMessage);
+
+    const tmb::MessageBus::SendStatus send_status = bus->Send(
+        sender_id, address, style, std::move(poison_tagged_message));
+    CHECK(send_status == tmb::MessageBus::SendStatus::kOK) <<
+       "Broadcast message from Foreman to workers failed";
```
zuyu: We need to edit this message, as the sender and the receivers changed.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -46,10 +48,14 @@ class WorkOrdersContainer {
    *
    * @param num_operators Number of operators in the query DAG.
    * @param num_numa_nodes Number of NUMA nodes in the system.
+   * @param query_id The ID of the query.
    **/
   WorkOrdersContainer(const std::size_t num_operators,
-                      const std::size_t num_numa_nodes)
-      : num_operators_(num_operators), num_numa_nodes_(num_numa_nodes) {
+                      const std::size_t num_numa_nodes,
+                      const std::size_t query_id)
```
zuyu: I'd suggest to move `query_id` as the first argument.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -76,7 +82,7 @@ class WorkOrdersContainer {
    * @return If there are pending WorkOrders.
    **/
   inline bool hasNormalWorkOrder(const std::size_t operator_index) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please change to `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -92,9 +98,9 @@ class WorkOrdersContainer {
    **/
   inline bool hasNormalWorkOrderForNUMANode(
       const std::size_t operator_index, const int numa_node_id) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
-    DEBUG_ASSERT(numa_node_id >= 0);
-    DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+    DCHECK(operator_index < num_operators_);
+    DCHECK_GE(numa_node_id, 0);
+    DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please change to `DCHECK_LT` and in Line 101 as well.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -108,7 +114,7 @@ class WorkOrdersContainer {
    * @return If there are pending rebuild WorkOrders.
    **/
   inline bool hasRebuildWorkOrder(const std::size_t operator_index) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please change to `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -124,9 +130,9 @@ class WorkOrdersContainer {
    **/
   inline bool hasRebuildWorkOrderForNUMANode(
       const std::size_t operator_index, const int numa_node_id) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
-    DEBUG_ASSERT(numa_node_id >= 0);
-    DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+    DCHECK(operator_index < num_operators_);
+    DCHECK_GE(numa_node_id, 0);
+    DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please change to `DCHECK_LT` and in Line 133 as well.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -144,9 +150,9 @@ class WorkOrdersContainer {
    **/
   WorkOrder* getNormalWorkOrderForNUMANode(const std::size_t operator_index,
                                            const int numa_node_id) {
-    DEBUG_ASSERT(operator_index < num_operators_);
-    DEBUG_ASSERT(numa_node_id >= 0);
-    DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+    DCHECK(operator_index < num_operators_);
+    DCHECK_GE(numa_node_id, 0);
+    DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please change to `DCHECK_LT` and in Line 153 as well.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -182,9 +188,9 @@ class WorkOrdersContainer {
    **/
   WorkOrder* getRebuildWorkOrderForNUMANode(const std::size_t operator_index,
                                             const int numa_node_id) {
-    DEBUG_ASSERT(operator_index < num_operators_);
-    DEBUG_ASSERT(numa_node_id >= 0);
-    DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+    DCHECK(operator_index < num_operators_);
+    DCHECK_GE(numa_node_id, 0);
+    DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please change to `DCHECK_LT` and in Line 191 as well.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -164,7 +170,7 @@ class WorkOrdersContainer {
    **/
   WorkOrder* getNormalWorkOrder(const std::size_t operator_index,
                                 const bool prefer_single_NUMA_node = true) {
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please change to `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -220,8 +226,9 @@ class WorkOrdersContainer {
    * @param operator_index The index of the operator in the query DAG.
    **/
   void addNormalWorkOrder(WorkOrder *workorder, const std::size_t operator_index) {
-    DEBUG_ASSERT(workorder != nullptr);
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(workorder != nullptr);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please change to `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -202,7 +208,7 @@ class WorkOrdersContainer {
    **/
   WorkOrder* getRebuildWorkOrder(const std::size_t operator_index,
                                  const bool prefer_single_NUMA_node = true) {
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please change to `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -238,8 +245,9 @@ class WorkOrdersContainer {
    **/
   void addRebuildWorkOrder(WorkOrder *workorder,
                            const std::size_t operator_index) {
-    DEBUG_ASSERT(workorder != nullptr);
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(workorder != nullptr);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please change to `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -254,9 +262,9 @@ class WorkOrdersContainer {
    **/
   inline std::size_t getNumNormalWorkOrdersForNUMANode(
       const std::size_t operator_index, const int numa_node_id) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
-    DEBUG_ASSERT(numa_node_id >= 0);
-    DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+    DCHECK(operator_index < num_operators_);
+    DCHECK_GE(numa_node_id, 0);
+    DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please change to `DCHECK_LT` and in Line 265 as well.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -286,9 +294,9 @@ class WorkOrdersContainer {
    **/
   inline std::size_t getNumRebuildWorkOrdersForNUMANode(
       const std::size_t operator_index, const int numa_node_id) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
-    DEBUG_ASSERT(numa_node_id >= 0);
-    DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+    DCHECK(operator_index < num_operators_);
+    DCHECK_GE(numa_node_id, 0);
+    DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please change to `DCHECK_LT` and in Line 297 as well.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -271,7 +279,7 @@ class WorkOrdersContainer {
    **/
   inline std::size_t getNumNormalWorkOrders(
       const std::size_t operator_index) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please use `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -303,7 +311,7 @@ class WorkOrdersContainer {
    **/
   inline std::size_t getNumRebuildWorkOrders(
       const std::size_t operator_index) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please use `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -418,8 +426,8 @@ class WorkOrdersContainer {
     void addWorkOrder(WorkOrder *workorder);
 
     bool hasWorkOrderForNUMANode(const int numa_node_id) const {
-      DEBUG_ASSERT(numa_node_id >= 0);
-      DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+      DCHECK_GE(numa_node_id, 0);
+      DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please use `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -440,8 +448,8 @@ class WorkOrdersContainer {
 
     std::size_t getNumWorkOrdersForNUMANode(
         const int numa_node_id) const {
-      DEBUG_ASSERT(numa_node_id >= 0);
-      DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+      DCHECK_GE(numa_node_id, 0);
+      DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please use `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -463,8 +471,8 @@ class WorkOrdersContainer {
     }
 
     WorkOrder* getWorkOrderForNUMANode(const int numa_node_id) {
-      DEBUG_ASSERT(numa_node_id >= 0);
-      DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+      DCHECK_GE(numa_node_id, 0);
+      DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please use `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -220,8 +226,9 @@ class WorkOrdersContainer {
    * @param operator_index The index of the operator in the query DAG.
    **/
   void addNormalWorkOrder(WorkOrder *workorder, const std::size_t operator_index) {
-    DEBUG_ASSERT(workorder != nullptr);
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(workorder != nullptr);
+    DCHECK(operator_index < num_operators_);
+    workorder->setQueryID(query_id_);
```
zuyu: I was wondering if we could set `query_id` for a `WorkOrder` during its construction.

query_execution/tests/WorkOrdersContainer_unittest.cpp
```diff
@@ -57,7 +57,7 @@ class MockNUMAWorkOrder : public WorkOrder {
 
 TEST(WorkOrdersContainerTest, ZeroNUMANodesTest) {
   // A container for one operator and no NUMA nodes.
-  WorkOrdersContainer w(1, 0);
+  WorkOrdersContainer w(1, 0, 0);
```
zuyu: Could we add comments for the magic numbers, i.e., `0 /* query_id */`?

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -1527,5 +1527,11 @@ void ExecutionGenerator::convertTableGenerator(
   temporary_relation_info_vec_.emplace_back(tablegen_index, output_relation);
 }
 
+S::InsertDestination* ExecutionGenerator::addNewInsertDestinationToQueryContext() {
+  std::unique_ptr<S::InsertDestination> insert_destination_proto(query_context_proto_->add_insert_destinations());
```
zuyu: I was wondering why use a unique pointer here?

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -1527,5 +1527,11 @@ void ExecutionGenerator::convertTableGenerator(
...
+  insert_destination_proto->set_query_id(query_context_proto_->query_id());
```
zuyu: Since you have set `query_id` in `QueryContext` proto, I was wondering why you have to do so in `InsertDestination` proto as well.

query_optimizer/ExecutionGenerator.hpp
```diff
@@ -101,10 +101,10 @@ class ExecutionGenerator {
         query_handle_(DCHECK_NOTNULL(query_handle)),
         execution_plan_(DCHECK_NOTNULL(query_handle->getQueryPlanMutable())),
         query_context_proto_(DCHECK_NOTNULL(query_handle->getQueryContextProtoMutable())) {
+    query_context_proto_->set_query_id(query_handle->query_id());
```
zuyu: Since the `QueryHandle` has `query_id`, I was wondering why you need to add it in `QueryContext` proto?

query_optimizer/tests/TestDatabaseLoader.cpp
```diff
@@ -122,6 +122,7 @@ void TestDatabaseLoader::loadTestRelation() {
                                          nullptr,
                                          &storage_manager_,
                                          0 /* dummy op index */,
+                                         0,  // dummy query ID.
```
zuyu: How about `0 /* dummy query id */,`?

query_execution/WorkOrdersContainer.hpp
```diff
@@ -238,8 +245,9 @@ class WorkOrdersContainer {
    **/
   void addRebuildWorkOrder(WorkOrder *workorder,
                            const std::size_t operator_index) {
-    DEBUG_ASSERT(workorder != nullptr);
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(workorder != nullptr);
+    DCHECK(operator_index < num_operators_);
+    workorder->setQueryID(query_id_);
```
zuyu: IIRC, `RebuildWorkOrder` has set `query_id` in its constructor.

zuyu:
@hbdeshmukh I have a suggestion: could we have a separate PR for adding `query_id` merged before this PR? Thanks!

cli/QuickstepCli.cpp
hbdeshmukh: Changed.

cli/tests/CommandExecutorTestRunner.cpp
hbdeshmukh: Changed.

query_execution/AdmitRequestMessage.hpp
hbdeshmukh: Changed.

query_execution/AdmitRequestMessage.hpp
hbdeshmukh: Removed.

query_execution/CMakeLists.txt
hbdeshmukh: Removed. Not sure why that came up.

query_execution/CMakeLists.txt
hbdeshmukh: Changed.

query_execution/Foreman.cpp
hbdeshmukh: Why? How will that make a difference? I feel that we should not get obsessed by style preferences.

query_execution/Foreman.cpp
hbdeshmukh: Changed, for all the comments in the file.

query_execution/Foreman.cpp
hbdeshmukh: I would rather let it be like this as it is more readable. The compiler should be smart enough to take care of the optimizations.

query_execution/Foreman.hpp
hbdeshmukh: Added.

query_execution/PolicyEnforcer.cpp
hbdeshmukh: No. The incoming message can belong to any of the concurrent queries getting executed. Based on the message we have to find out which QueryManager should be used. ``admitted_queries_[query_id]`` gives us the appropriate QueryManager, which in turn processes the incoming message.

query_execution/PolicyEnforcer.cpp
hbdeshmukh: Done.

query_execution/PolicyEnforcer.cpp
hbdeshmukh: Changed.

query_execution/PolicyEnforcer.cpp
hbdeshmukh: No, this is a very simple round-robin policy. I will add more sophisticated (like fair, priority) policies later.

query_execution/PolicyEnforcer.cpp
hbdeshmukh: Changed.

query_execution/PolicyEnforcer.cpp
hbdeshmukh: Changed.

query_execution/QueryContext.proto
hbdeshmukh: Why is that?

query_execution/QueryExecutionMessages.proto
hbdeshmukh: Why does it matter?

query_execution/QueryExecutionTypedefs.hpp
hbdeshmukh: I really think this is unnecessary. A description for each message type is enough for the reader.

query_execution/QueryExecutionUtil.hpp
hbdeshmukh: Changed.

query_execution/WorkOrdersContainer.hpp
hbdeshmukh: How does it matter?

query_execution/WorkOrdersContainer.hpp
hbdeshmukh: I prefer ``DCHECK``, as it is more readable. Also, the cpplint doesn't prefer ``DCHECK_LT`` over ``DCHECK``.

query_execution/WorkOrdersContainer.hpp
hbdeshmukh: That's a good point. RebuildWorkOrder has query_id in its constructor because in the execute method, it needs the ``query_id`` for sending a pipelining message, which in turn needs the query ID. 

Adding the query ID to all WorkOrder constructors requires modifying base relational operator, base WorkOrder and all the constructor calls to the derived WorkOrder classes. I can mark this task as a TODO and return to it once this PR gets merged. Let me know if that sounds alright.

query_execution/tests/WorkOrdersContainer_unittest.cpp
hbdeshmukh: Why is that needed?

query_optimizer/ExecutionGenerator.cpp
hbdeshmukh: Because that makes it easier to pass on the pointer from this function at the call site. Do you have any alternative in mind?

query_execution/Foreman.cpp
zuyu: I like the way you did for `PolicyEnforcer` constructor. Could we do the similar for `Foreman`?

query_execution/QueryExecutionTypedefs.hpp
zuyu: No, when a reader encounters so many types, it would be good to have some logical relations behind the order.

query_execution/Foreman.cpp
zuyu: The compiler would, but I am more talking about adding the readability by introducing a new method for this.

query_execution/PolicyEnforcer.cpp
zuyu: The main thing here is to serialize every message twice. One in here, and the other in `QueryManager`. Could we do so once?

query_execution/PolicyEnforcer.cpp
zuyu: I was thinking of adding a `virtual` method for this, so that each policy would have its own implementation. Are we going to this direction when adding more sophisticated policies?

query_execution/QueryContext.proto
zuyu: `query_id` will not change in future, but some fields may add or remove. So setting `query_id` to `1` reduce the chances to change the assigning numbers.

query_execution/QueryExecutionMessages.proto
zuyu: It is also related to how we order things. It is nature for a message that first states the `query_id`, and then some detailed info within this query, i.e., `operator_index`.

query_execution/WorkOrdersContainer.hpp
zuyu: I think we have a container per query, so `query_id` is the most significant argument, then some info for this query, i.e., `num_operators`.

query_execution/WorkOrdersContainer.hpp
zuyu: I know that, so I suggest to add `query_id` as a separate PR before merging this one.

query_execution/tests/WorkOrdersContainer_unittest.cpp
zuyu: For code readability.

query_optimizer/ExecutionGenerator.cpp
zuyu: Yes, just remove the `unique_ptr`, and use the raw pointer. Because we do not manage the memory space for proto, although I don't know whether internally proto's `add_xxx` method uses a `new` operator.

query_execution/PolicyEnforcer.cpp
hbdeshmukh: Derived classes of PolicyEnforcer for each policy implementation.

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,185 @@
...
+        // Check if the query's execution is over.
+        if (curr_query_manager->getQueryExecutionState().hasQueryExecutionFinished()) {
+          // If the query has been executed, remove it.
+          finished_queries_ids.push_back(query_manager_it->first);
+        }
+        break;
+      }
+    }
+  }
+  for (std::size_t finished_qid : finished_queries_ids) {
```
zuyu: Please add `const` for `finished_query_id`

query_optimizer/ExecutionGenerator.cpp
hbdeshmukh: InsertDestination needs the query ID because the ID needs to be put in the pipelining message sent by the InsertDestination. Now InsertDestination can be constructed from its proto representation. Therefore it is good to have it in the InsertDestination proto.

query_optimizer/ExecutionGenerator.hpp
hbdeshmukh: Because the query ID in QueryContext proto helps identifying which query does the proto belong to.

hbdeshmukh:
Thanks @zuyu. I believe that the presence of the query ID in the query context is necessary for the distributed version to distinguish one context for an active query from another. Similarly in the InsertDestination, it is necessary as it needs to send messages which should have query ID as a field. 

I think there are two approaches we can take here: 

1. As you suggested, we separate a PR for changing the relational operators and work orders. Merge that first and then merge this one. The amount of change required for this PR is fairly large (but not complicated) which includes:

- Adding query ID field in the constructor of all the relational operators.
- Adding query ID field in all the getAllWorkOrders() calls. 
- Modify the base class WorkOrder constructor to include the query ID field as a parameter. 
- Each derived WorkOrder constructor calls the base WorkOrder class constructor. 
- Modifying all the affected tests. 

2. Alternatively, we can let the current PR merge first. In parallel, I can work on making the changes on adding query ID in relational operators and work orders. That way, we can remove some code from this PR (e.g. right now work orders are assigned a query ID in the WorkOrderContainer class. That arrangement can be removed, once each WorkOrder gets the query ID right when it is constructed).

I am more inclined towards option 2, as it may help move things faster.

zuyu:
@hbdeshmukh I understand your proposal. For the approach 2, the main drawback is to add some hack code regarding `query_id`, and then fix it. I am afraid that the fix later may not as clean as what do it right in the first place using the approach 1.

Do we target merging the scheduling feature for the paper experiments?

zuyu:
Replaced by https://github.com/apache/incubator-quickstep/pull/14

PR #191, opened 2016-04-26
Refactored how min and max BLOCKSIZEMBs are calculated.

cramja:
This corrects an issue found by @pateljm . The original code essentially used magic numbers for max/min blocksize and used Mbi instead of Mb. The changes to the code correct both these issues and update corresponding tests.


pateljm:
LGTM. Merging.

PR #192, opened 2016-04-26
Refactored makeRoomForBlock.

zuyu:
This refactor includes:
 * Renamed to `makeRoomForBlockOrBlob`.
 * Removed a unnecessary variable.
 * Simplified the `if-else` statement.
 * Added more comments about this method.

### DIFF
--- a/storage/StorageManager.cpp
+++ b/storage/StorageManager.cpp
@@ -397,7 +397,7 @@ void* StorageManager::allocateSlots(const std::size_t num_slots,
       = MAP_PRIVATE | MAP_ANONYMOUS | MAP_ALIGNED_SUPER;
 #endif
 
-  makeRoomForBlock(num_slots);
+  makeRoomForBlockOrBlob(num_slots);
   void *slots = nullptr;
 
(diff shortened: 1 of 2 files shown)


cramja:
@zuyu did we run the TPCH query 8/9 on this?

cramja:
@zuyu Also, when thinking about refactoring all the method names in EvictionPolicy.hpp from `blockPastTenseVerb` to `blockOrBlobPastTenseVerb` I think it seems too wordy.

Maybe we could make the naming more succinct to instead name them all of the form `notifyPastTenseVerb(block_id block_or_blob)`.

What do you think?

cramja:
BTW `pastTenseVerb` is one of `{created, evicted, unreferenced, referenced, deleted}`

zuyu:
@cramja I did not run TPC-H over this PR. This just refactors a method internally.

zuyu:
@cramja Alternatively for `EvictionPolicy`, we could add a comment before the class definition: `block refers to a block or blob.`

cramja:
@zuyu I think the simplest thing here is to use you approach. I will make a PR for this.

pateljm:
Merging.

PR #193, opened 2016-04-26
Better comments in eviction policy.

cramja:
@zuyu This will help explain better the meaning of 'block' in the Eviction Policy classes.


zuyu:
LGTM!

PR #195, opened 2016-04-27
Enable semi-join optimization for left-deep trees through bloom filters

saketj:
This PR adds support for semi-join optimization for left-deep trees through the use of bloom filters. The motivation behind this PR is that often the bottom-most join in a left-deep tree returns maximum number of tuples from the fact table, a majority of which are then discarded by joins up the tree. If somehow, the selectivity information from up the tree could be passed down to the bottom-most operators, we could avoid the cost of potentially passing and storing huge number of unnecessary tuples during query execution. 

Bloom filters are lightweight data structures that can help us push down this selectivity information down the operator tree. Therefore, the key idea is to delay the processing of the bottom-most probe operator on the fact table, until bloom filters on all the build operators on all the dimension tables are built. Then pass the bloom filters sideways to this probe operator, which can use these bloom filters to weed out the unnecessary tuples.

This feature is controlled by GFLAGs and is turned off by default. Pass the `--optimize_joins` flag from the command-line to explicitly enable this optimization.

On a typical left-deep join query like SSB Query 4, this optimization reduces the query execution time by almost 40%.

### DIFF
--- a/query_execution/QueryContext.cpp
+++ b/query_execution/QueryContext.cpp
@@ -83,34 +83,8 @@ QueryContext::QueryContext(const serialization::QueryContext &proto,
   for (int i = 0; i < proto.join_hash_tables_size(); ++i) {
     join_hash_tables_.emplace_back(
         JoinHashTableFactory::CreateResizableFromProto(proto.join_hash_tables(i),
-                                                       storage_manager));
-
-    // Check if there are any build side bloom filter defined on the hash table.
-    if (proto.join_hash_tables(i).build_side_bloom_filter_id_size() > 0) {
-      join_hash_tables_[i]->enableBuildSideBloomFilter();
-      bloom_filter_id build_bloom_filter_id = proto.join_hash_tables(i).build_side_bloom_filter_id(0);
-      join_hash_tables_[i]->setBuildSideBloomFilter(bloom_filters_[build_bloom_filter_id].get());
-    }
-
-    // Check if there are any probe side bloom filters defined on the hash table.
-    if (proto.join_hash_tables(i).probe_side_bloom_filters_size() > 0) {
-      join_hash_tables_[i]->enableProbeSideBloomFilter();
-      // Add as many probe bloom filters as defined by the proto.
-      for (int j = 0; j < proto.join_hash_tables(i).probe_side_bloom_filters_size(); ++j) {
-        // Add the pointer to the probe bloom filter within the list of probe bloom filters to use.
-        const auto probe_side_bloom_filter = proto.join_hash_tables(i).probe_side_bloom_filters(j);
-        const bloom_filter_id probe_bloom_filter_id = probe_side_bloom_filter.probe_side_bloom_filter_id();
-        join_hash_tables_[i]->addProbeSideBloomFilter(bloom_filters_[probe_bloom_filter_id].get());
-
-        // Add the attribute ids corresponding to this probe bloom filter.
-        std::vector<attribute_id> probe_attribute_ids;
-        for (int k = 0; k < probe_side_bloom_filter.probe_side_attr_ids_size(); ++k) {
-          const attribute_id probe_attribute_id = probe_side_bloom_filter.probe_side_attr_ids(k);
-          probe_attribute_ids.push_back(probe_attribute_id);
-        }
-        join_hash_tables_[i]->addProbeSideAttributeIds(std::move(probe_attribute_ids));
-      }
-    }
+                                                       storage_manager,
+                                                       bloom_filters_));
   }
 
   for (int i = 0; i < proto.insert_destinations_size(); ++i) {
--- a/query_execution/QueryContext.cpp
+++ b/query_execution/QueryContext.cpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015-2016 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -37,6 +39,7 @@
 #include "storage/InsertDestination.pb.h"
 #include "types/TypedValue.hpp"
 #include "types/containers/Tuple.hpp"
+#include "utility/BloomFilter.hpp"
 #include "utility/SortConfiguration.hpp"
 
 #include "glog/logging.h"
@@ -65,6 +68,13 @@ QueryContext::QueryContext(const serialization::QueryContext &proto,
                                                         storage_manager));
   }
 
+  for (int i = 0; i < proto.bloom_filters_size(); ++i) {
+    const serialization::BloomFilter &bloom_filter_proto = proto.bloom_filters(i);
+    bloom_filters_.emplace_back(new BloomFilter(bloom_filter_proto.bloom_filter_seed(),
+                                                bloom_filter_proto.number_of_hashes(),
+                                                bloom_filter_proto.bloom_filter_size()));
+  }
+
   for (int i = 0; i < proto.generator_functions_size(); ++i) {
     const GeneratorFunctionHandle *func_handle =
         GeneratorFunctionFactory::Instance().reconstructFromProto(proto.generator_functions(i));
@@ -77,6 +87,33 @@ QueryContext::QueryContext(const serialization::QueryContext &proto,
     join_hash_tables_.emplace_back(
         JoinHashTableFactory::CreateResizableFromProto(proto.join_hash_tables(i),
                                                        storage_manager));
+
+    // Check if there are any build side bloom filter defined on the hash table.
+    if (proto.join_hash_tables(i).build_side_bloom_filter_id_size() > 0) {
+      join_hash_tables_[i]->enableBuildSideBloomFilter();
+      bloom_filter_id build_bloom_filter_id = proto.join_hash_tables(i).build_side_bloom_filter_id(0);
+      join_hash_tables_[i]->setBuildSideBloomFilter(bloom_filters_[build_bloom_filter_id].get());
+    }
+
+    // Check if there are any probe side bloom filters defined on the hash table.
+    if (proto.join_hash_tables(i).probe_side_bloom_filters_size() > 0) {
+      join_hash_tables_[i]->enableProbeSideBloomFilter();
+      // Add as many probe bloom filters as defined by the proto.
+      for (int j = 0; j < proto.join_hash_tables(i).probe_side_bloom_filters_size(); ++j) {
+        // Add the pointer to the probe bloom filter within the list of probe bloom filters to use.
+        const auto probe_side_bloom_filter = proto.join_hash_tables(i).probe_side_bloom_filters(j);
+        const bloom_filter_id probe_bloom_filter_id = probe_side_bloom_filter.probe_side_bloom_filter_id();
+        join_hash_tables_[i]->addProbeSideBloomFilter(bloom_filters_[probe_bloom_filter_id].get());
+
+        // Add the attribute ids corresponding to this probe bloom filter.
+        std::vector<attribute_id> probe_attribute_ids;
+        for (int k = 0; k < probe_side_bloom_filter.probe_side_attr_ids_size(); ++k) {
+          const attribute_id probe_attribute_id = probe_side_bloom_filter.probe_side_attr_ids(k);
+          probe_attribute_ids.push_back(probe_attribute_id);
+        }
+        join_hash_tables_[i]->addProbeSideAttributeIds(std::move(probe_attribute_ids));
+      }
+    }
   }
 
   for (int i = 0; i < proto.insert_destinations_size(); ++i) {
@@ -138,6 +175,12 @@ bool QueryContext::ProtoIsValid(const serialization::QueryContext &proto,
                                 const CatalogDatabaseLite &database) {
   for (int i = 0; i < proto.aggregation_states_size(); ++i) {
     if (!AggregationOperationState::ProtoIsValid(proto.aggregation_states(i), database)) {
+      return false;
+    }
+  }
+
+  for (int i = 0; i < proto.bloom_filters_size(); ++i) {
+    if (!BloomFilter::ProtoIsValid(proto.bloom_filters(i))) {
       return false;
     }
   }
--- a/query_execution/QueryContext.cpp
+++ b/query_execution/QueryContext.cpp
@@ -69,10 +69,7 @@ QueryContext::QueryContext(const serialization::QueryContext &proto,
   }
 
   for (int i = 0; i < proto.bloom_filters_size(); ++i) {
-    const serialization::BloomFilter &bloom_filter_proto = proto.bloom_filters(i);
-    bloom_filters_.emplace_back(new BloomFilter(bloom_filter_proto.bloom_filter_seed(),
-                                                bloom_filter_proto.number_of_hashes(),
-                                                bloom_filter_proto.bloom_filter_size()));
+    bloom_filters_.emplace_back(new BloomFilter(proto.bloom_filters(i)));
   }
 
   for (int i = 0; i < proto.generator_functions_size(); ++i) {
--- a/query_execution/QueryContext.hpp
+++ b/query_execution/QueryContext.hpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015-2016 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -32,6 +34,7 @@
 #include "storage/HashTable.hpp"
 #include "storage/InsertDestination.hpp"
 #include "types/containers/Tuple.hpp"
+#include "utility/BloomFilter.hpp"
 #include "utility/Macros.hpp"
 #include "utility/SortConfiguration.hpp"
 
@@ -61,6 +64,11 @@ class QueryContext {
    * @brief A unique identifier for an AggregationOperationState per query.
    **/
   typedef std::uint32_t aggregation_state_id;
+
+  /**
+   * @brief A unique identifier for a BloomFilter per query.
+   **/
+  typedef std::uint32_t bloom_filter_id;
 
   /**
    * @brief A unique identifier for a GeneratorFunctionHandle per query.
@@ -181,6 +189,52 @@ class QueryContext {
   }
 
   /**
+   * @brief Whether the given BloomFilter id is valid.
+   *
+   * @param id The BloomFilter id.
+   *
+   * @return True if valid, otherwise false.
+   **/
+  bool isValidBloomFilterId(const bloom_filter_id id) const {
+    return id < bloom_filters_.size();
+  }
+
+  /**
+   * @brief Get a mutable reference to the BloomFilter.
+   *
+   * @param id The BloomFilter id.
+   *
+   * @return The BloomFilter, already created in the constructor.
+   **/
+  inline BloomFilter* getBloomFilterMutable(const bloom_filter_id id) {
+    DCHECK_LT(id, bloom_filters_.size());
+    return bloom_filters_[id].get();
+  }
+
+  /**
+   * @brief Get a constant pointer to the BloomFilter.
+   *
+   * @param id The BloomFilter id.
+   *
+   * @return The constant pointer to BloomFilter that is 
+   *         already created in the constructor.
+   **/
+  inline const BloomFilter* getBloomFilter(const bloom_filter_id id) const {
+    DCHECK_LT(id, bloom_filters_.size());
+    return bloom_filters_[id].get();
+  }
+
+  /**
+   * @brief Destory the given BloomFilter.
+   *
+   * @param id The id of the BloomFilter to destroy.
+   **/
+  inline void destroyBloomFilter(const bloom_filter_id id) {
+    DCHECK_LT(id, bloom_filters_.size());
+    bloom_filters_[id].reset();
+  }
+
+  /**
    * @brief Whether the given GeneratorFunctionHandle id is valid.
    *
    * @param id The GeneratorFunctionHandle id.
@@ -257,7 +311,7 @@ class QueryContext {
    *
    * @param id The JoinHashTable id in the query.
    *
-   * @return The JoinHashTable, alreadly created in the constructor.
+   * @return The JoinHashTable, already created in the constructor.
    **/
   inline JoinHashTable* getJoinHashTable(const join_hash_table_id id) {
     DCHECK_LT(id, join_hash_tables_.size());
@@ -408,6 +462,7 @@ class QueryContext {
 
  private:
   std::vector<std::unique_ptr<AggregationOperationState>> aggregation_states_;
+  std::vector<std::unique_ptr<BloomFilter>> bloom_filters_;
   std::vector<std::unique_ptr<const GeneratorFunctionHandle>> generator_functions_;
   std::vector<std::unique_ptr<InsertDestination>> insert_destinations_;
   std::vector<std::unique_ptr<JoinHashTable>> join_hash_tables_;
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -588,8 +588,8 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   std::vector<attribute_id> probe_original_attribute_ids;
   std::vector<attribute_id> build_original_attribute_ids;
 
-  relation_id probe_relation_id;
-  relation_id build_relation_id;
+  const CatalogRelation *referenced_stored_probe_relation;
+  const CatalogRelation *referenced_stored_build_relation;
 
   bool any_probe_attributes_nullable = false;
   bool any_build_attributes_nullable = false;
@@ -600,15 +600,14 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
       physical_plan->left_join_attributes();
   for (const E::AttributeReferencePtr &left_join_attribute : left_join_attributes) {
     // Try to determine the original stored relation referenced in the Hash Join.
-    const CatalogRelation *referenced_stored_catalog_relation =
+    referenced_stored_probe_relation =
         optimizer_context_->catalog_database()->getRelationByName(left_join_attribute->relation_name());
-    if (referenced_stored_catalog_relation == nullptr) {
+    if (referenced_stored_probe_relation == nullptr) {
       // Hash Join optimizations are not possible, if the referenced relation cannot be determined.
       skip_hash_join_optimization = true;
     } else {
-      probe_relation_id = referenced_stored_catalog_relation->getID();
       const attribute_id probe_operator_attribute_id =
-      referenced_stored_catalog_relation->getAttributeByName(left_join_attribute->attribute_name())->getID();
+          referenced_stored_probe_relation->getAttributeByName(left_join_attribute->attribute_name())->getID();
       probe_original_attribute_ids.emplace_back(probe_operator_attribute_id);
     }
 
@@ -625,15 +624,14 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
       physical_plan->right_join_attributes();
   for (const E::AttributeReferencePtr &right_join_attribute : right_join_attributes) {
     // Try to determine the original stored relation referenced in the Hash Join.
-    const CatalogRelation *referenced_stored_catalog_relation =
+    referenced_stored_build_relation =
         optimizer_context_->catalog_database()->getRelationByName(right_join_attribute->relation_name());
-    if (referenced_stored_catalog_relation == nullptr) {
+    if (referenced_stored_build_relation == nullptr) {
       // Hash Join optimizations are not possible, if the referenced relation cannot be determined.
       skip_hash_join_optimization = true;
     } else {
-      build_relation_id = referenced_stored_catalog_relation->getID();
       const attribute_id build_operator_attribute_id =
-      referenced_stored_catalog_relation->getAttributeByName(right_join_attribute->attribute_name())->getID();
+          referenced_stored_build_relation->getAttributeByName(right_join_attribute->attribute_name())->getID();
       build_original_attribute_ids.emplace_back(build_operator_attribute_id);
     }
 
@@ -673,7 +671,7 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
       std::swap(probe_attribute_ids, build_attribute_ids);
       std::swap(any_probe_attributes_nullable, any_build_attributes_nullable);
       std::swap(probe_original_attribute_ids, build_original_attribute_ids);
-      std::swap(probe_relation_id, build_relation_id);
+      std::swap(referenced_stored_probe_relation, referenced_stored_build_relation);
     }
   }
 
@@ -833,10 +831,10 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   if (FLAGS_optimize_joins && !skip_hash_join_optimization) {
     execution_heuristics_->addHashJoinInfo(build_operator_index,
                                            join_operator_index,
-                                           build_relation_id,
-                                           probe_relation_id,
-                                           &build_original_attribute_ids,
-                                           &probe_original_attribute_ids,
+                                           referenced_stored_build_relation,
+                                           referenced_stored_probe_relation,
+                                           std::move(build_original_attribute_ids),
+                                           std::move(probe_original_attribute_ids),
                                            join_hash_table_index);
   }
 }
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -50,6 +50,7 @@
 #include "expressions/scalar/ScalarAttribute.hpp"
 #include "query_execution/QueryContext.hpp"
 #include "query_execution/QueryContext.pb.h"
+#include "query_optimizer/ExecutionHeuristics.hpp"
 #include "query_optimizer/OptimizerContext.hpp"
 #include "query_optimizer/QueryHandle.hpp"
 #include "query_optimizer/QueryPlan.hpp"
@@ -144,6 +145,9 @@ static const volatile bool aggregate_hashtable_type_dummy
 
 DEFINE_bool(parallelize_load, true, "Parallelize loading data files.");
 
+DEFINE_bool(optimize_joins, false,
+            "Enable post execution plan generation optimizations for joins.");
+
 namespace E = ::quickstep::optimizer::expressions;
 namespace P = ::quickstep::optimizer::physical;
 namespace S = ::quickstep::serialization;
@@ -196,6 +200,11 @@ void ExecutionGenerator::generatePlan(const P::PhysicalPtr &physical_plan) {
     execution_plan_->addDependenciesForDropOperator(
         drop_table_index,
         temporary_relation_info.producer_operator_index);
+  }
+
+  // Optimize execution plan based on heuristics captured during execution plan generation, if enabled.
+  if (FLAGS_optimize_joins) {
+    execution_heuristics_->optimizeExecutionPlan(execution_plan_, query_context_proto_);
   }
 
 #ifdef QUICKSTEP_DISTRIBUTED
@@ -576,12 +585,33 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   std::vector<attribute_id> probe_attribute_ids;
   std::vector<attribute_id> build_attribute_ids;
 
+  std::vector<attribute_id> probe_original_attribute_ids;
+  std::vector<attribute_id> build_original_attribute_ids;
+
+  relation_id probe_relation_id;
+  relation_id build_relation_id;
+
   bool any_probe_attributes_nullable = false;
   bool any_build_attributes_nullable = false;
+
+  bool skip_hash_join_optimization = false;
 
   const std::vector<E::AttributeReferencePtr> &left_join_attributes =
       physical_plan->left_join_attributes();
   for (const E::AttributeReferencePtr &left_join_attribute : left_join_attributes) {
+    // Try to determine the original stored relation referenced in the Hash Join.
+    const CatalogRelation *referenced_stored_catalog_relation =
+        optimizer_context_->catalog_database()->getRelationByName(left_join_attribute->relation_name());
+    if (referenced_stored_catalog_relation == nullptr) {
+      // Hash Join optimizations are not possible, if the referenced relation cannot be determined.
+      skip_hash_join_optimization = true;
+    } else {
+      probe_relation_id = referenced_stored_catalog_relation->getID();
+      const attribute_id probe_operator_attribute_id =
+      referenced_stored_catalog_relation->getAttributeByName(left_join_attribute->attribute_name())->getID();
+      probe_original_attribute_ids.emplace_back(probe_operator_attribute_id);
+    }
+
     const CatalogAttribute *probe_catalog_attribute
         = attribute_substitution_map_[left_join_attribute->id()];
     probe_attribute_ids.emplace_back(probe_catalog_attribute->getID());
@@ -594,6 +624,19 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   const std::vector<E::AttributeReferencePtr> &right_join_attributes =
       physical_plan->right_join_attributes();
   for (const E::AttributeReferencePtr &right_join_attribute : right_join_attributes) {
+    // Try to determine the original stored relation referenced in the Hash Join.
+    const CatalogRelation *referenced_stored_catalog_relation =
+        optimizer_context_->catalog_database()->getRelationByName(right_join_attribute->relation_name());
+    if (referenced_stored_catalog_relation == nullptr) {
+      // Hash Join optimizations are not possible, if the referenced relation cannot be determined.
+      skip_hash_join_optimization = true;
+    } else {
+      build_relation_id = referenced_stored_catalog_relation->getID();
+      const attribute_id build_operator_attribute_id =
+      referenced_stored_catalog_relation->getAttributeByName(right_join_attribute->attribute_name())->getID();
+      build_original_attribute_ids.emplace_back(build_operator_attribute_id);
+    }
+
     const CatalogAttribute *build_catalog_attribute
         = attribute_substitution_map_[right_join_attribute->id()];
     build_attribute_ids.emplace_back(build_catalog_attribute->getID());
@@ -629,6 +672,8 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
       std::swap(probe_cardinality, build_cardinality);
       std::swap(probe_attribute_ids, build_attribute_ids);
       std::swap(any_probe_attributes_nullable, any_build_attributes_nullable);
+      std::swap(probe_original_attribute_ids, build_original_attribute_ids);
+      std::swap(probe_relation_id, build_relation_id);
     }
   }
 
@@ -783,6 +828,17 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
       std::forward_as_tuple(join_operator_index,
                             output_relation));
   temporary_relation_info_vec_.emplace_back(join_operator_index, output_relation);
+
+  // Add heuristics for the Hash Join, if enabled.
+  if (FLAGS_optimize_joins && !skip_hash_join_optimization) {
+    execution_heuristics_->addHashJoinInfo(build_operator_index,
+                                           join_operator_index,
+                                           build_relation_id,
+                                           probe_relation_id,
+                                           &build_original_attribute_ids,
+                                           &probe_original_attribute_ids,
+                                           join_hash_table_index);
+  }
 }
 
 void ExecutionGenerator::convertNestedLoopsJoin(
@@ -894,7 +950,6 @@ void ExecutionGenerator::convertCopyFrom(
                                        scan_operator_index,
                                        false /* is_pipeline_breaker */);
 }
-
 
 void ExecutionGenerator::convertCreateIndex(
   const P::CreateIndexPtr &physical_plan) {
--- a/query_optimizer/ExecutionHeuristics.cpp
+++ b/query_optimizer/ExecutionHeuristics.cpp
@@ -19,6 +19,7 @@
 
 #include <cstddef>
 #include <utility>
+#include <unordered_map>
 #include <vector>
 
 #include "catalog/CatalogTypedefs.hpp"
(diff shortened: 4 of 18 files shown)


query_execution/QueryContext.hpp
```diff
@@ -181,6 +189,51 @@ class QueryContext {
...
+  }
+
+  /**
+   * @brief Get a constant reference to the BloomFilter.
+   *
+   * @param id The BloomFilter id.
+   *
+   * @return The BloomFilter, already created in the constructor.
+   **/
+  inline const BloomFilter* getBloomFilter(const bloom_filter_id id) const {
```
hbdeshmukh: This function returns a pointer but the doxygen says const reference.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -563,12 +572,33 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
   bool any_probe_attributes_nullable = false;
   bool any_build_attributes_nullable = false;
 
+  bool skip_hash_join_optimization = false;
+
   const std::vector<E::AttributeReferencePtr> &left_join_attributes =
       physical_plan->left_join_attributes();
   for (const E::AttributeReferencePtr &left_join_attribute : left_join_attributes) {
+    // Try to determine the original relation referenced in the Hash Join.
+    const CatalogRelation *catalogRelation =
```
hbdeshmukh: Variable names are usually ``catalog_relation`` instead of ``catalogRelation``. Also, can you think of a more informative name for this variable?

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,104 @@
...
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ **/
+
+#include "query_optimizer/ExecutionHeuristics.hpp"
+
+#include <cstddef>
+#include <vector>
+#include <utility>
```
hbdeshmukh: Please sort the included files.

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,104 @@
...
+  // Currently this only optimizes left deep joins using bloom filters.
+  // It uses a simple algorithm to discover the left deep joins.
+  // It starts with the first hash join in the plan and keeps on iterating
+  // over the next hash joins, till a probe on a different relation id is found.
+  // The set of hash joins found in this way forms a chain and can be recognized
+  // as a left deep join. It becomes a candidate for optimization.
+
+  // The optimization is done by modifying each of the build operators in the chain
+  // to generate a bloom filter on the build key during their hash table creation.
+  // The origin(bottom-most) probe operator is then modified to query all the bloom
```
hbdeshmukh: You can say leaf-level probe operator.

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,104 @@
...
+  // filters generated from all the build operators in the chain. Bloom filter queries
```
hbdeshmukh: What do you mean by bloom filter query?

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * @brief A simple internal class that holds information about various
+ *        hash joins within the execution plan for a query.
+ **/
+class HashJoinInfo {
+ public:
+  HashJoinInfo(QueryPlan::DAGNodeIndex build_operator_index,
+               QueryPlan::DAGNodeIndex join_operator_index,
+               const relation_id &build_relation_id,
+               const relation_id &probe_relation_id,
```
hbdeshmukh: I am wondering if it's worth using a reference for an integer vs using the integer itself. If int takes up four bytes and a reference is 8 bytes, then we should go for the shorter one, I feel.

relational_operators/CMakeLists.txt
```diff
@@ -92,7 +92,7 @@ target_link_libraries(quickstep_relationaloperators_BuildHashOperator
                       quickstep_storage_StorageManager
                       quickstep_storage_TupleReference
                       quickstep_storage_TupleStorageSubBlock
-                      quickstep_storage_ValueAccessor
+                      quickstep_storage_ValueAccessor
```
hbdeshmukh: Trailing spaces.

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,104 @@
...
+  // are issued just prior to probing the hash table.
+
+  QueryPlan::DAGNodeIndex origin_node = 0;
+  while (origin_node < hash_joins_.size() - 1) {
```
hbdeshmukh: This function is somewhat complicated to understand. I will find you this afternoon, if you are in the office and we can discuss this in person.

query_execution/QueryContext.hpp
saketj: Updated the comment.

query_optimizer/ExecutionGenerator.cpp
saketj: Fixed.

query_optimizer/ExecutionHeuristics.cpp
saketj: Done.

query_optimizer/ExecutionHeuristics.cpp
saketj: Updated.

query_optimizer/ExecutionHeuristics.cpp
saketj: It meant to query the bloom filter for membership test. I have updated the comments to be more precise here. Thanks.

query_optimizer/ExecutionHeuristics.hpp
saketj: Removed the reference usage.

relational_operators/CMakeLists.txt
saketj: Fixed.

saketj:
@hbdeshmukh  Thanks Harshad for the comments. I have addressed them now. Sure, I will be there around in the afternoon at the department. We can discuss it the optimization function.

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,137 @@
...
+   * @param build_relation_id Id of the relation on which hash table is being built.
+   * @param probe_relation_id Id of the relation on which hash table is being probed.
+   * @param build_attributes List of attributes on which hash table is being built.
+   * @param probe_attributes List of attributes on which hash table is being probed.
+   * @param join_hash_table_id Id of the hash table which refers to the actual hash
+   *        table within the query context.
+   **/
+  inline void addHashJoinInfo(QueryPlan::DAGNodeIndex build_operator_index,
+                              QueryPlan::DAGNodeIndex join_operator_index,
+                              const relation_id build_relation_id,
```
hbdeshmukh: Please rename this variable for better clarity.

hbdeshmukh:
I and @saketj had a discussion. Saket will improve the documentation and variable names for better clarity. We also had a discussion on how to write tests for this code.

pateljm:
Sounds good @hbdeshmukh. Thanks! 

Thanks @saketj -- This is an awesome feature!

storage/HashTable.hpp
```diff
@@ -1429,6 +1492,11 @@ HashTablePutResult HashTable<ValueT, resizable, serializable, force_key_copy, al
                                        variable_size,
                                        (*functor)(*accessor),
                                        using_prealloc ? &prealloc_state : nullptr);
+            // Insert into bloom filter, if enabled.
+            if (has_build_side_bloom_filter_) {
+              this->build_bloom_filter_->insert(static_cast<const std::uint8_t *>(key.getDataPtr()),
```
hbdeshmukh: I am not completely sure if this code will be thread safe. I think we should have another round of discussion to discuss the thread-safety of using bloom filters inside the hash table.

saketj: @hbdeshmukh Harshad, thread safety is being guaranteed here from inside the `build_bloom_filter_->insert()`, which implements an exclusive mutex lock to allow only one thread to modify the bloom filter.
For containment test, I guess we will not require a lock, since the assumption is that a bloom filter will never be used, unless it has been built.

hbdeshmukh: I see, thanks for that information. However, we are inserting an entry in the bloom filter for every tuple, and while doing so, acquiring a lock on the bloom filter. Even though the time to insert an entry in the bloom filter may not be high (flipping a single bit), we are essentially serializing all the worker threads building the hash table. 

I feel that there are some relatively easy optimizations possible, and it may be worth discussing them.

saketj: Thanks. That's a good point. I guess inserting into a local bloom filter and doing a union with the shared bloom filter with locking may solve this problem. Sure, I will stop by today in the afternoon to discuss this in detail.

query_execution/QueryContext.cpp
```diff
@@ -65,6 +68,13 @@ QueryContext::QueryContext(const serialization::QueryContext &proto,
                                                         storage_manager));
   }
 
+  for (int i = 0; i < proto.bloom_filters_size(); ++i) {
+    const serialization::BloomFilter &bloom_filter_proto = proto.bloom_filters(i);
+    bloom_filters_.emplace_back(new BloomFilter(bloom_filter_proto.bloom_filter_seed(),
```
zuyu: Could we have a `BloomFilter` constructor that takes its proto as the single argument? Thus, this line could be simplified to

`bloom_filters_.emplace_back(new BloomFilter(bloom_filter_proto));`

query_execution/QueryContext.cpp
```diff
@@ -77,6 +87,33 @@ QueryContext::QueryContext(const serialization::QueryContext &proto,
...
+        join_hash_tables_[i]->addProbeSideBloomFilter(bloom_filters_[probe_bloom_filter_id].get());
+
+        // Add the attribute ids corresponding to this probe bloom filter.
+        std::vector<attribute_id> probe_attribute_ids;
+        for (int k = 0; k < probe_side_bloom_filter.probe_side_attr_ids_size(); ++k) {
+          const attribute_id probe_attribute_id = probe_side_bloom_filter.probe_side_attr_ids(k);
+          probe_attribute_ids.push_back(probe_attribute_id);
+        }
+        join_hash_tables_[i]->addProbeSideAttributeIds(std::move(probe_attribute_ids));
+      }
```
zuyu: Have you considered of moving some of the code here inside `JoinHashTableFactory ::CreateResizableFromProto`? For example, calling `setBuildSideBloomFilter`.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -563,12 +572,33 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+    // Try to determine the original stored relation referenced in the Hash Join.
+    const CatalogRelation *referenced_stored_catalog_relation =
+        optimizer_context_->catalog_database()->getRelationByName(left_join_attribute->relation_name());
+    if (referenced_stored_catalog_relation == nullptr) {
+      // Hash Join optimizations are not possible, if the referenced relation cannot be determined.
+      skip_hash_join_optimization = true;
+    } else {
+      probe_relation_id = referenced_stored_catalog_relation->getID();
+      const attribute_id probe_operator_attribute_id =
+      referenced_stored_catalog_relation->getAttributeByName(left_join_attribute->attribute_name())->getID();
```
zuyu: According to the code style, please add four whitespace indentations.

https://github.com/pivotalsoftware/quickstep/wiki/Quickstep-Style-Guide

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -581,6 +611,19 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+    // Try to determine the original stored relation referenced in the Hash Join.
+    const CatalogRelation *referenced_stored_catalog_relation =
+        optimizer_context_->catalog_database()->getRelationByName(right_join_attribute->relation_name());
+    if (referenced_stored_catalog_relation == nullptr) {
+      // Hash Join optimizations are not possible, if the referenced relation cannot be determined.
+      skip_hash_join_optimization = true;
+    } else {
+      build_relation_id = referenced_stored_catalog_relation->getID();
+      const attribute_id build_operator_attribute_id =
+      referenced_stored_catalog_relation->getAttributeByName(right_join_attribute->attribute_name())->getID();
```
zuyu: Ditto for breaking a line with the four whitespace indentation.

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,137 @@
...
+namespace execution_heuristics_internal {
+
+/**
+ * @brief A simple internal class that holds information about various
+ *        hash joins within the execution plan for a query.
+ **/
+class HashJoinInfo {
+ public:
+  HashJoinInfo(QueryPlan::DAGNodeIndex build_operator_index,
+               QueryPlan::DAGNodeIndex join_operator_index,
```
zuyu: We could also mark two `operator_index`s above `const`.

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,137 @@
...
+               const relation_id &build_relation_id,
+               const relation_id &probe_relation_id,
+               std::vector<attribute_id> &&build_attributes,
+               std::vector<attribute_id> &&probe_attributes,
+               const QueryContext::join_hash_table_id &join_hash_table_id)
```
zuyu: I agree, it does not worth to use a reference for an argument in a plain type.

So, please remove the reference here as well.

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,137 @@
...
+    : build_operator_index_(build_operator_index),
+      join_operator_index_(join_operator_index),
+      build_relation_id_(build_relation_id),
+      probe_relation_id_(probe_relation_id),
+      build_attributes_(build_attributes),
+      probe_attributes_(probe_attributes),
```
zuyu: Please use the correct way for `rvalue`:

`probe_attributes_(std::move(probe_attributes)),`

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+#include "catalog/CatalogTypedefs.hpp"
+#include "query_execution/QueryContext.pb.h"
+#include "query_optimizer/QueryPlan.hpp"
+#include "utility/Macros.hpp"
+
+#include "glog/logging.h"
+
+namespace quickstep {
+
```
zuyu: Please remove this empty line.

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,137 @@
...
+   * @param probe_attributes List of attributes on which hash table is being probed.
+   * @param join_hash_table_id Id of the hash table which refers to the actual hash
+   *        table within the query context.
+   **/
+  inline void addHashJoinInfo(QueryPlan::DAGNodeIndex build_operator_index,
+                              QueryPlan::DAGNodeIndex join_operator_index,
+                              const relation_id build_relation_id,
+                              const relation_id probe_relation_id,
+                              std::vector<attribute_id> *build_attributes,
+                              std::vector<attribute_id> *probe_attributes,
```
zuyu: Please use `rvalue` properly:

`std::vector<attribute_id> &&build_attributes,`

And in the caller site:

`std::move(build_attributes),`

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,105 @@
...
+      // Add node dependencies from chained build nodes to origin node probe.
+      for (std::size_t i = 1; i < chained_nodes.size(); ++i) {  // Note: It starts from index 1.
+        query_plan->addDirectDependency(hash_joins_[origin_node].join_operator_index_,
+                                        hash_joins_[i].build_operator_index_,
+                                        true /* is_pipeline_breaker */);
+      }
+    }
+
+    // Update the origin node.
+    origin_node = chained_nodes.at(chained_nodes.size() - 1) + 1;
```
zuyu: Could we refactor to the following?

`origin_node = chained_nodes.back() + 1;`

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,105 @@
...
+        chained_nodes.push_back(i);
+      } else {
+        break;
+      }
+    }
+
+    // Only chains of length greater than one are suitable candidates for semi-join optimization.
+    if (chained_nodes.size() > 1) {
+      std::vector<std::pair<QueryContext::bloom_filter_id, std::vector<attribute_id>>> probe_bloom_filter_info;
+      for (std::size_t &i : chained_nodes) {
```
zuyu: How about the following refactor:

`for (const std::size_t node : chained_nodes) {`

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,105 @@
...
+      if (checked_relation_id == expected_relation_id) {
+        chained_nodes.push_back(i);
+      } else {
+        break;
+      }
+    }
+
+    // Only chains of length greater than one are suitable candidates for semi-join optimization.
+    if (chained_nodes.size() > 1) {
+      std::vector<std::pair<QueryContext::bloom_filter_id, std::vector<attribute_id>>> probe_bloom_filter_info;
```
zuyu: What if we change to `std::unordered_map<QueryContext::bloom_filter_id, std::vector<attribute_id>>`?

storage/HashTable.hpp
```diff
@@ -990,6 +991,61 @@ class HashTable : public HashTableBase<resizable,
...
+  /**
+   * @brief This function adds a vector of attribute ids corresponding to a
+   *        bloom filter used during the probe phase of this hash table.
+   * @warning Should call enable_probe_side_bloom_filter() first to enable
+   *          bloom filter usage during probe phase.
+   *
+   * @param probe_attribute_ids The vector of attribute ids to use for probing
+   *        the bloom filter.
+   **/
+  inline void addProbeSideAttributeIds(std::vector<attribute_id> &&probe_attribute_ids) {
```
zuyu: I don't understand why using `rvalue` here? Actually, I was thinking if we could initialize these values in the constructor or the factory method.

utility/BloomFilter.hpp
```diff
@@ -47,7 +53,30 @@ class BloomFilter {
...
+   **/
+  BloomFilter(const std::uint64_t random_seed,
+              const std::size_t hash_fn_count,
+              const std::uint64_t bit_array_size_in_bytes)
+      : random_seed_(random_seed),
+        hash_fn_count_(hash_fn_count),
+        array_size_in_bytes_(bit_array_size_in_bytes),
+        is_bit_array_owner_(true) {
+    array_size_ = bit_array_size_in_bytes * kNumBitsPerByte;
+    bit_array_.reset(new std::uint8_t[bit_array_size_in_bytes]);
```
zuyu: We could move both `array_size_` and `bit_array_` in the init list of the constructor as well.

utility/BloomFilter.hpp
```diff
@@ -47,7 +53,30 @@ class BloomFilter {
...
+    reset();
```
zuyu: Do we call `reset` more than once? If no, as the implementation of `reset` is simple, so we could move the code in the constructor and get rid of this method.

utility/BloomFilter.proto
```diff
@@ -0,0 +1,30 @@
+//   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+//    University of Wisconsin—Madison.
```
zuyu: For consistency, please add one more whitespace indentation.

utility/BloomFilter.proto
```diff
@@ -0,0 +1,30 @@
...
+
+message BloomFilter {
+  // The default values were determined from empirical experiments.
+  // These values control the amount of false positivity that
+  // is expected from Bloom Filter.
+  // - Default seed for initializing family of hashes = 0xA5A5A5A55A5A5A5A.
+  // - Default bloom filter size = 10 KB.
+  // - Default number of hash functions used in bloom filter = 5.
+  required fixed64 bloom_filter_seed = 1 [default = 0xA5A5A5A55A5A5A5A];
+  required fixed64 bloom_filter_size = 2 [default = 10000];
```
zuyu: I was wondering why we need a `fixed64` here?

utility/BloomFilter.proto
```diff
@@ -0,0 +1,30 @@
...
+  required int32 number_of_hashes = 3 [default = 5];
```
zuyu: Could we use `uint32` instead?

zuyu:
@saketj I was wondering, as an alternative to add a new `execution_heuristics_`, we just treat this as a new method called `optimizeHashJoin` in the existing `ExecutionGeneerator`.

saketj:
@hbdeshmukh and I had discussion on how we can reduce lock contention between various worker threads during bloom filter insertion. An improved locking mechanism is now implemented in the latest commit.

saketj:
@zuyu Thanks Zuyu for the comments. I am working on incorporating them into this PR. 
As for ExecutionHeuristics class, I can foresee it as something that will grow to incorporate many more complex optimization logic, along with its own data structures. For example, the logic for identification of left-deep trees is currently simple and ignores some complicated scenarios. This will be work of subsequent PRs, I guess. Also, setting bloom filter properties based on the relation statistics is  another component that has been introduced.

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -101,5 +104,23 @@ void ExecutionHeuristics::optimizeExecutionPlan(QueryPlan *query_plan,
   }
 }
 
+void ExecutionHeuristics::setBloomFilterProperties(serialization::BloomFilter *bloom_filter_proto,
+                                                   const CatalogRelation *relation) {
+  const std::size_t cardinality = relation->estimateTupleCardinality();
+  if (cardinality < kOneThousand) {
+    bloom_filter_proto->set_bloom_filter_size(kOneThousand / kCompressionFactor);
+    bloom_filter_proto->set_number_of_hashes(kVeryLowSparsityHash);
+  } else if (cardinality >= kOneThousand && cardinality < kTenThousand) {
```
zuyu: We could simplify to ` else if (cardinality < kTenThousand)`, as `cardinality >= kOneThousand` is always `true`.

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -101,5 +104,23 @@ void ExecutionHeuristics::optimizeExecutionPlan(QueryPlan *query_plan,
...
+    bloom_filter_proto->set_bloom_filter_size(kTenThousand / kCompressionFactor);
+    bloom_filter_proto->set_number_of_hashes(kLowSparsityHash);
+  } else if (cardinality >= kTenThousand && cardinality < kHundredThousand) {
+    bloom_filter_proto->set_bloom_filter_size(kHundredThousand / kCompressionFactor);
+    bloom_filter_proto->set_number_of_hashes(kMediumSparsityHash);
+  } else if (cardinality >= kHundredThousand) {
```
zuyu: How about using `else`, instead of `else if`?

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -101,5 +104,23 @@ void ExecutionHeuristics::optimizeExecutionPlan(QueryPlan *query_plan,
...
+void ExecutionHeuristics::setBloomFilterProperties(serialization::BloomFilter *bloom_filter_proto,
+                                                   const CatalogRelation *relation) {
+  const std::size_t cardinality = relation->estimateTupleCardinality();
+  if (cardinality < kOneThousand) {
+    bloom_filter_proto->set_bloom_filter_size(kOneThousand / kCompressionFactor);
+    bloom_filter_proto->set_number_of_hashes(kVeryLowSparsityHash);
+  } else if (cardinality >= kOneThousand && cardinality < kTenThousand) {
+    bloom_filter_proto->set_bloom_filter_size(kTenThousand / kCompressionFactor);
+    bloom_filter_proto->set_number_of_hashes(kLowSparsityHash);
+  } else if (cardinality >= kTenThousand && cardinality < kHundredThousand) {
```
zuyu: We could remove the always-true `cardinality >= kTenThousand` expression here.

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,161 @@
...
+  static const std::size_t kTenThousand = 10000;
+  static const std::size_t kHundredThousand = 100000;
+  static const std::size_t kMillion = 1000000;
+
+  static const std::size_t kCompressionFactor = 10;
+
+  static const std::size_t kVeryLowSparsityHash = 1;
+  static const std::size_t kLowSparsityHash = 2;
+  static const std::size_t kMediumSparsityHash = 5;
+  static const std::size_t kHighSparsityHash = 10;
```
zuyu: I was wondering if we could set above five constants from `kCompressionFactor` to `kHighSparsityHash` as `gflag` variables with the same default values as you set, so that we could benchmark various combinations without recompiling the code.

query_optimizer/ExecutionHeuristics.cpp
```diff
@@ -0,0 +1,126 @@
...
+    bloom_filter_proto->set_bloom_filter_size(kHundredThousand / kCompressionFactor);
+    bloom_filter_proto->set_number_of_hashes(kMediumSparsityHash);
+  } else if (cardinality >= kHundredThousand) {
+    bloom_filter_proto->set_bloom_filter_size(kMillion / kCompressionFactor);
```
zuyu: Also, you may want to pre-compute the bloom filter size for each case, to avoid do so for each bloom filter, although it is still lightweight.

query_optimizer/ExecutionHeuristics.hpp
```diff
@@ -0,0 +1,161 @@
...
+ *  @{
+ */
+
+namespace execution_heuristics_internal {
+
+/**
+ * @brief A simple internal class that holds information about various
+ *        hash joins within the execution plan for a query.
+ **/
+class HashJoinInfo {
```
zuyu: How about moving this class inside `ExecutionHeuristics`? Also, I notice that you did not mark all data members `private`. If you intend to do so, how about using `struct` instead?

query_execution/QueryContext.cpp
saketj: Added a new constructor to initialize from the bloom filter protobuf.

query_execution/QueryContext.cpp
saketj: @zuyu These functions (e.g. `addProdeSideBloomFilter(), setBuildSideBloomFilter()`) may not always be called and therefore, I thought we can save the cost of passing references to the data structures of Query Context here, when invoking `JoinHashTableFactory ::CreateResizableFromProto()`

query_optimizer/ExecutionGenerator.cpp
saketj: Fixed.

query_optimizer/ExecutionGenerator.cpp
saketj: Fixed.

query_optimizer/ExecutionHeuristics.hpp
saketj: Updated.

query_optimizer/ExecutionHeuristics.hpp
saketj: Removed.

query_optimizer/ExecutionHeuristics.hpp
saketj: Thanks. Updated these.

query_optimizer/ExecutionHeuristics.hpp
saketj: Removed.

query_optimizer/ExecutionHeuristics.hpp
saketj: Thanks. Updated.

query_optimizer/ExecutionHeuristics.cpp
saketj: Refactored.

query_optimizer/ExecutionHeuristics.cpp
saketj: Refactored.

query_optimizer/ExecutionHeuristics.cpp
saketj: Changed.

storage/HashTable.hpp
saketj: @zuyu  These may not be always initialized and hence, I thought to add them as separate functions.

utility/BloomFilter.hpp
saketj: Moved.

utility/BloomFilter.hpp
saketj: @zuyu The reset() function is also used by the BloomFilterStorageSubBlock to clear out the bloom filter, hence added as a separate method.

utility/BloomFilter.proto
saketj: Updated.

utility/BloomFilter.proto
saketj: Changed to `uint32`. Thanks

query_optimizer/ExecutionHeuristics.cpp
saketj: Done.

query_optimizer/ExecutionHeuristics.cpp
saketj: Done.

query_optimizer/ExecutionHeuristics.cpp
saketj: Done.

query_optimizer/ExecutionHeuristics.hpp
saketj: Moved it inside the `ExecutionHeuristics` class and changed to `struct` now.

query_optimizer/ExecutionHeuristics.hpp
saketj: Thanks. This is a good suggestion. I am still in process of figuring out which ones of these will be useful in benchmarking. Will update them to GFLAGS as necessary.

query_execution/QueryContext.cpp
zuyu: Sorry, I did not get that.

The goal is to make `QueryContext` clean with a few reconstruction methods.

All the methods called here are setting up for `join_hash_tables_[i]`, reconstructed by `JoinHashTableFactory::CreateResizableFromProto`, based on `proto.join_hash_tables(i)`. So, I was wondering what additional info we need from `QueryContext`.

storage/HashTable.hpp
zuyu: Sorry, I did not get that. If we do not have probe side attribute ids, we could leave the `vector` as an empty one.

If I understand correctly, this entry is set once, and won't change. Then we could even mark it as `const` during the reconstruction using `JoinHashTableFactory::CreateResizableFromProto`.

query_execution/QueryContext.cpp
saketj: @zuyu The information about bloom filters is some additional info that we need to pass from the `QueryContext`. But, nevermind, I liked your idea of keeping the `QueryContext` cleaner and I have refactored the code now to move all these initializations inside `JoinHashTableFactory::CreateResizableFromProto`.

saketj:
@zuyu @hbdeshmukh  I have addressed all of the comments now. I have also added the unit tests for this PR at `query_optimizer/tests/ExecutionHeuristics_unittest.cpp`. Please let me know if this all looks good and if we can merge.
There is one more incremental feature that builds on top of this PR and it adds the ability to push the bloom filters even further down the tree to a selection operator preceding a probe. I have the changes in my local copy and I am planning to immediately open that PR, once this gets merged. Thanks.

storage/HashTable.hpp
saketj: Moved it now inside `JoinHashTableFactory::CreateResizableFromProto`

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,307 @@
...
+                       const QueryPlan::DAGNodeIndex join_operator_index,
+                       const CatalogRelation *build_relation,
+                       const CatalogRelation *probe_relation,
+                       const attribute_id build_attribute_id,
+                       const attribute_id probe_attribute_id,
+                       const QueryContext::join_hash_table_id join_hash_table_id) {
+    std::vector<attribute_id> build_attribute_ids;
+    build_attribute_ids.push_back(build_attribute_id);
+    std::vector<attribute_id> probe_attribute_ids;
+    probe_attribute_ids.push_back(probe_attribute_id);
```
zuyu: We could refactor to

```
  std::vector<attribute_id> build_attribute_ids(1, build_attribute_id);
  std::vector<attribute_id> probe_attribute_ids(1, probe_attribute_id);
```

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,307 @@
...
+                                          std::move(probe_attribute_ids),
+                                          join_hash_table_id);
+  }
+
+  QueryPlan::DAGNodeIndex createDummyBuildHashOperator(QueryPlan *query_plan,
+                                                       const CatalogRelation *build_relation,
+                                                       const attribute_id build_attribute_id,
+                                                       const QueryContext::join_hash_table_id join_hash_table_index) {
+    std::vector<attribute_id> build_attribute_ids;
+    build_attribute_ids.push_back(build_attribute_id);
```
zuyu: We could refactor to `const std::vector<attribute_id> build_attribute_ids(1, build_attribute_id);`

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,307 @@
...
+    return build_operator_index;
+  }
+
+  QueryPlan::DAGNodeIndex createDummyHashJoinOperator(QueryPlan *query_plan,
+                                                      const CatalogRelation *build_relation,
+                                                      const CatalogRelation *probe_relation,
+                                                      const attribute_id probe_attribute_id,
+                                                      const QueryContext::join_hash_table_id join_hash_table_index) {
+    std::vector<attribute_id> probe_attribute_ids;
+    probe_attribute_ids.push_back(probe_attribute_id);
```
zuyu: We could refactor to `const std::vector<attribute_id> probe_attribute_ids(1, probe_attribute_id);`

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,307 @@
...
+};
+
+TEST_F(ExecutionHeuristicsTest, HashJoinOptimizedTest) {
+  // This test case creates three hash joins, all of which are being probed on the same relation.
+  // Since the probe are being made on the same relation, ExecutionHeuristics should optimize
+  // these hash joins using bloom filters.
+
+  std::unique_ptr<ExecutionHeuristics> execution_heuristics(new ExecutionHeuristics());
+  std::unique_ptr<QueryPlan> query_plan(new QueryPlan());
+  std::unique_ptr<serialization::QueryContext> query_context_proto(new serialization::QueryContext());
```
zuyu: What do you think of moving these three variables in `ExecutionHeuristicsTest`, and then reset them in `SetUp`? Thus, we don't have to duplicate these code in each test fixture.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+  const CatalogRelation *build_relation_1 = createCatalogRelation("build_relation_1");
+  const CatalogRelation *build_relation_2 = createCatalogRelation("build_relation_2");
+  const CatalogRelation *build_relation_3 = createCatalogRelation("build_relation_3");
+  const CatalogRelation *probe_relation_1 = createCatalogRelation("probe_relation_1");
+
+  const attribute_id build_attribute_id_1 = 0;
+  const attribute_id build_attribute_id_2 = 0;
+  const attribute_id build_attribute_id_3 = 0;
```
zuyu: I was wondering why we give three build attributes with the same id `0`?

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+  // Test whether correct number of bloom filters were added.
+  EXPECT_EQ(1, query_context_proto->join_hash_tables(0).build_side_bloom_filter_id_size());
+  EXPECT_EQ(1, query_context_proto->join_hash_tables(1).build_side_bloom_filter_id_size());
+  EXPECT_EQ(1, query_context_proto->join_hash_tables(2).build_side_bloom_filter_id_size());
+  EXPECT_EQ(3, query_context_proto->join_hash_tables(0).probe_side_bloom_filters_size());
+
+  // Test that the DAG was modified correctly or not.
+  // Probe operator 1 should have now build operator 1 and build operator 2 added as dependencies.
+  auto node_dependencies = query_plan->getQueryPlanDAG().getDependencies(probe_operator_index_1);
```
zuyu: Mark `const`.

storage/HashTable.hpp
```diff
@@ -2164,6 +2247,27 @@ void HashTable<ValueT, resizable, serializable, force_key_copy, allow_duplicate_
...
       [&](auto *accessor) -> void {  // NOLINT(build/c++11)
     while (accessor->next()) {
+      // Probe any bloom filters, if enabled.
+      if (has_probe_side_bloom_filter_) {
+        DCHECK_EQ(probe_bloom_filters_.size(), probe_attribute_ids_.size());
+        // Check if the key is contained in the BloomFilters or not.
+        bool bloom_miss = false;
+        for (std::size_t i = 0; i < probe_bloom_filters_.size() && !bloom_miss; ++i) {
+          const BloomFilter *bloom_filter = probe_bloom_filters_.at(i);
+          for (const attribute_id &attr_id : probe_attribute_ids_.at(i)) {
```
zuyu: For above two `at(i)`s, we should use `[i]` for performance reasons.

storage/HashTableFactory.hpp
```diff
@@ -305,10 +309,37 @@ class HashTableFactory {
...
+
+        // Add the attribute ids corresponding to this probe bloom filter.
+        std::vector<attribute_id> probe_attribute_ids;
+        for (int k = 0; k < probe_side_bloom_filter.probe_side_attr_ids_size(); ++k) {
+          const attribute_id probe_attribute_id = probe_side_bloom_filter.probe_side_attr_ids(k);
+          probe_attribute_ids.push_back(probe_attribute_id);
+        }
+        hash_table->addProbeSideAttributeIds(std::move(probe_attribute_ids));
+      }
+    }
```
zuyu: What do you think of moving all these late-initializations into `CreateResizable` and the hash table constructors? With this being said, we don't need additional helper methods in `HashTable.hpp`, and could mark these data members `const` wherever possible.

```
    bool has_build_side_bloom_filter = false;
    BloomFilter *build_bloom_filter = nullptr;
    if (proto.build_side_bloom_filters_size() > 0) {
      has_build_side_bloom_filter = true;
      build_bloom_filter = bloom_filters[proto.build_side_bloom_filter_id(0)].get();
    }

    const bool has_probe_side_bloom_filter = proto.probe_side_bloom_filters_size() > 0;
    std::vector<const BloomFilter*> probe_bloom_filters(proto.probe_side_bloom_filters_size());
    std::vector<std::vector<attribute_id>> probe_attribute_ids_list;
    for (int i = 0; i < proto.probe_side_bloom_filters_size(); ++i) {
      const auto probe_side_bloom_filter = proto.probe_side_bloom_filters(i);
      probe_bloom_filters[i] = bloom_filters[probe_side_bloom_filter.probe_side_bloom_filter_id()].get();

      std::vector<attribute_id> probe_attribute_ids(probe_side_bloom_filter.probe_side_attr_ids_size());
      for (int j = 0; j < probe_side_bloom_filter.probe_side_attr_ids_size(); ++j) {
        probe_attribute_ids[j] = probe_side_bloom_filter.probe_side_attr_ids(j);
      }
      probe_attribute_ids_list.push_back(std::move(probe_attribute_ids));
    }

    return CreateResizable(HashTableImplTypeFromProto(proto.hash_table_impl_type()),
                           key_types,
                           proto.estimated_num_entries(),
                           has_build_side_bloom_filter,
                           build_bloom_filter,
                           has_probe_side_bloom_filter,
                           std::move(probe_bloom_filters),
                           std::move(probe_attribute_ids_list),
                           storage_manager);
```

We also need to modify various hash table constructors, i.e., HashTable below:

```
// storage/HashTable.hpp
  HashTable(...,
            const bool has_build_side_bloom_filter,
            BloomFilter *build_bloom_filter,
            const bool has_probe_side_bloom_filter,
            std::vector<const BloomFilter*> &&probe_bloom_filters,
            std::vector<std::vector<attribute_id>> &&probe_attribute_ids)
      : ...
        has_build_side_bloom_filter_(has_build_side_bloom_filter),
        build_bloom_filter_(build_bloom_filter),
        has_probe_side_bloom_filter_(has_probe_side_bloom_filter),
        probe_bloom_filters_(std::move(probe_bloom_filters)),
        probe_attribute_ids_(std::move(probe_attribute_ids)) {}

...

  const bool has_build_side_bloom_filter_;
  BloomFilter *build_bloom_filter_;

  const bool has_probe_side_bloom_filter_;
  const std::vector<const BloomFilter*> probe_bloom_filters_;
  const std::vector<std::vector<attribute_id>> probe_attribute_ids_;
```

utility/BloomFilter.hpp
```diff
@@ -61,39 +91,168 @@ class BloomFilter {
...
 
   /**
+   * @brief Constructor.
+   * @note When a bloom filter proto is passed as an initializer,
+   *       then the bit_array is owned and managed by this class.
+   *
+   * @param bloom_filter_proto The protobuf representation of a
+   *        bloom filter configuration.
+   **/
+  explicit BloomFilter(const serialization::BloomFilter bloom_filter_proto)
```
zuyu: Please use the reference (`&`) here to avoid a deep copy. Thanks.

utility/BloomFilter.hpp
```diff
@@ -61,39 +91,168 @@ class BloomFilter {
...
+  }
+
+  static bool ProtoIsValid(const serialization::BloomFilter &bloom_filter) {
+    if (bloom_filter.number_of_hashes() <= 0) {
+      return false;
+    }
+    if (bloom_filter.bloom_filter_size() <= 0) {
+      return false;
+    }
+    return true;
```
zuyu: We should refactor as following:

```
  static bool ProtoIsValid(const serialization::BloomFilter &proto) {
    return proto. IsInitialized();
  }
```

Note that we don't need to check whether these two values are negative, as we are using unsigned values for them.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
saketj: Refactored.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
saketj: Refactored.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
saketj: @zuyu Adding the const qualifier prevents the move semantics to be applied later in the function below, so dropping it.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
saketj: Thanks. Moved them inside the Setup() now.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
saketj: The build_attribute ids are inconsequential to the actual tests being performed. They are more of being used as placeholders.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
saketj: Updated.

storage/HashTable.hpp
saketj: Thanks. Updated.

utility/BloomFilter.hpp
saketj: Thanks. Updated.

utility/BloomFilter.hpp
saketj: Refactored.

storage/HashTableFactory.hpp
saketj: @zuyu I also thought about pushing the initialization of these members all the way down to the constructor. However, I realized that such a change will lead to a large number of modifications to the constructor of all the specific hash table implementations, and in a way some code will be duplicated at each constructor of those hash table implementations. So, I thought this may be a more cleaner way to do it at one place, since this is common to all the instances of hash tables created by the factory.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+  // Test whether correct number of bloom filters were added.
+  EXPECT_EQ(1, query_context_proto_->join_hash_tables(0).build_side_bloom_filter_id_size());
+  EXPECT_EQ(1, query_context_proto_->join_hash_tables(1).build_side_bloom_filter_id_size());
+  EXPECT_EQ(1, query_context_proto_->join_hash_tables(2).build_side_bloom_filter_id_size());
+  EXPECT_EQ(3, query_context_proto_->join_hash_tables(0).probe_side_bloom_filters_size());
+
+  // Test that the DAG was modified correctly or not.
+  // Probe operator 1 should have now build operator 1 and build operator 2 added as dependencies.
+  auto const node_dependencies = query_plan_->getQueryPlanDAG().getDependencies(probe_operator_index_1);
```
hbdeshmukh: Please rename to ``probe_node_dependencies`` or something similar for clarity.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,301 @@
...
+  EXPECT_EQ(1u, node_dependencies.count(build_operator_index_2));
```
hbdeshmukh: We should perform a similar check before modifying the DAG.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -0,0 +1,301 @@
...
+  EXPECT_EQ(1u, node_dependencies.count(build_operator_index_3));
+}
+
+TEST_F(ExecutionHeuristicsTest, HashJoinNotOptimizedTest) {
+  // This test case creates three hash joins, all of which are being probed on the different relation.
```
hbdeshmukh: Please make the comment consistent with this test fixture.

storage/HashTableFactory.hpp
zuyu: No, we won't have duplicated implementations at all. Check out how `InsertDestination` handles the same situation.

What we need to do is for various hash table constructors, to pass multiple arguments along the way.

One benefit of this approach is that we could get rid of several set-up helper methods, and mark the data members `const`.

saketj: @zuyu Had a discussion about this in Monday meeting today and we have agreed to pick this code refactoring as a part of a future PR, since it does not affect the current performance benchmarking required for the paper deadline. Marking this as a TODO for now. Thanks.

utility/CMakeLists.txt
```diff
@@ -146,6 +146,10 @@ configure_file (
   "${CMAKE_CURRENT_BINARY_DIR}/UtilityConfig.h"
 )
 
+QS_PROTOBUF_GENERATE_CPP(quickstep_utility_BloomFilter_proto_srcs
+                         quickstep_utility_BloomFilter_proto_hdrs
+                         BloomFilter.proto)
+
```
zuyu: We don't need this extra empty line. Please remove it.

hbdeshmukh:
FYI, I cancelled the Travis build.

hbdeshmukh:
LGTM, merging.

PR #196, opened 2016-04-28
Fixed the deadlock when loading while evicting.

zuyu:
This PR improved #183 by counting all referenced shards for `makeRoomForBlock` to prevent deadlocks mentioned in #70 when loading while evicting.


pateljm:
LGTM. Merging.

PR #197, opened 2016-04-28
Fixes the lost alias problem with PullUpProjectExpressions

jianqiao:
This PR is a small fix to `PullUpProjectExpressions` to retain column alias during the "pull up selection" optimization.

For example, before this fix, the query
```
SELECT *
FROM (
  SELECT i, SUM(i) AS sum
  FROM generate_series(1, 2) AS gs(i)
  GROUP BY i
) t1 JOIN (
  SELECT j, AVG(j) avg
  FROM generate_series(1, 2) AS gs(j)
  GROUP BY j
) t2 ON i = j;
```
generates the following output:
```
+-----------+--------------------+-----------+------------------------+
|i          |$aggregate0         |j          |$aggregate0             |
+-----------+--------------------+-----------+------------------------+
|          1|                   1|          1|                       1|
|          2|                   2|          2|                       2|
+-----------+--------------------+-----------+------------------------+
```
where the specified alias names `sum` and `avg` are lost.

After the fix, the output becomes:
```
+-----------+--------------------+-----------+------------------------+
|i          |sum                 |j          |avg                     |
+-----------+--------------------+-----------+------------------------+
|          1|                   1|          1|                       1|
|          2|                   2|          2|                       2|
+-----------+--------------------+-----------+------------------------+
```

### DIFF
--- a/query_optimizer/rules/RuleHelper.cpp
+++ b/query_optimizer/rules/RuleHelper.cpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


pateljm:
LGTM. Merging.

PR #198, opened 2016-04-28
Increase the size of the sharded lock manager

pateljm:
Having small number of entries in the sharded lock manager introduces artificial conflict causing the buffer pool to grow. The size of an entry in the lock_manager_ is small, so don't be so stingy with the size of the lock manager. By default, we want to run well on large memory boxes.


zuyu:
LGTM! Could you please also describe what you have observed regarding actual memory usage after this change? Thanks!

PR #199, opened 2016-04-28
Improved shared subplan support.

jianqiao:
This PR improves the shared subplan (aka. common table expression) support to handle TPC-H query 15. There are two main updates:
1. Augmented `SimpleCostModel` to handle  `SharedSubplanReference` physical plan node.
2. Revised `SharedSubplanReference` logical/physical plan nodes to avoid `ExprId` resolution collision when a same shared plan is referenced multiple times in a query.


pateljm:
LGTM. Merging.

PR #202, opened 2016-04-29
Set block slots correctly.

zuyu:
This PR fixed some issues merged in #41.


pateljm:
LGTM. Merging (ctests were run locally).

PR #204, opened 2016-04-30
Fixes creation of BITWEAVING_H indices.

cramja:
Fixes bug which disallows the creation of BITWEAVING_H indices. Adds a unit test for the parser portion.


pateljm:
LGTM. Merging.

PR #205, opened 2016-05-02
Fixes a bug with NUMA settings.

cramja:
- NUMA lib was not being compiled in
- Bug in method which calculates the number of NUMA nodes.

### DIFF
--- a/cli/DefaultsConfigurator.hpp
+++ b/cli/DefaultsConfigurator.hpp
@@ -22,6 +22,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "storage/StorageConfig.h"
 #include "utility/Macros.hpp"
 
 #ifdef QUICKSTEP_HAVE_LIBNUMA
@@ -48,6 +49,21 @@ class DefaultsConfigurator {
    **/
   static std::size_t GetNumHardwareThreads() {
     return std::thread::hardware_concurrency();
+  }
+
+  /**
+   * @brief Get the number of available numa sockets.
+   *
+   * @return Number of available numa sockets. Always 1 if the system doesn't
+   *         have libnuma.
+   **/
+  static std::size_t GetNumNUMANodes() {
+  #ifdef QUICKSTEP_HAVE_LIBNUMA
+    // Id of the maximum node.
+    return numa_max_node() + 1;
+  #else
+    return 1;
+  #endif
   }
 
   /**
--- a/cli/InputParserUtil.cpp
+++ b/cli/InputParserUtil.cpp
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "catalog/CatalogConfig.h"
+#include "storage/StorageConfig.h"
 #include "utility/StringUtil.hpp"
 
 #include "glog/logging.h"
--- a/cli/QuickstepCli.cpp
+++ b/cli/QuickstepCli.cpp
@@ -258,8 +258,7 @@ int main(int argc, char* argv[]) {
       InputParserUtil::ParseWorkerAffinities(real_num_workers,
                                              quickstep::FLAGS_worker_affinities);
 
-  const std::size_t num_numa_nodes_covered =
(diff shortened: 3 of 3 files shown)


cli/DefaultsConfigurator.hpp
```diff
@@ -22,6 +22,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "cli/CliConfig.h"
```
zuyu: Include `storage/StorageConfig.h` instead, and remove `QUICKSTEP_HAVE_LIBNUMA`.

cramja:
I am still solving issues related to this bug. Please do not merge until finished.

cli/QuickstepCli.cpp
```diff
@@ -258,8 +258,7 @@ int main(int argc, char* argv[]) {
       InputParserUtil::ParseWorkerAffinities(real_num_workers,
                                              quickstep::FLAGS_worker_affinities);
 
-  const std::size_t num_numa_nodes_covered =
-      DefaultsConfigurator::GetNumNUMANodesCoveredByWorkers(worker_cpu_affinities);
```
cramja: There's no issue with the old method, it just isn't used anymore.

cli/QuickstepCli.cpp
```diff
@@ -280,7 +279,8 @@ int main(int argc, char* argv[]) {
   Foreman foreman(&bus,
                   query_processor->getDefaultDatabase(),
                   query_processor->getStorageManager(),
-                  num_numa_nodes_covered);
+                  -1, /* CPU id to bind foreman. -1 is unbound. */
```
cramja: @hbdeshmukh Is there a need to bind the foreman to a node? I set it to unbound because it seems like a good default.

cli/DefaultsConfigurator.hpp
cramja: We could. Though seems to me like this is a style preference.

hbdeshmukh: I agree with @zuyu because ``QUICKSTEP_HAVE_LIBNUMA`` is a result of whether NUMA libraries are present on the system or not (Ref: https://github.com/pivotalsoftware/quickstep/blob/master/storage/CMakeLists.txt#L41). Either we repeat the same check in cli/CMakeLists.txt or just use whatever is already done in storage module.

cli/QuickstepCli.cpp
hbdeshmukh: Sure, you can leave that unbound.

cli/DefaultsConfigurator.hpp
cramja: Okay, will change.

cli/InputParserUtil.cpp
```diff
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "catalog/CatalogConfig.h"
+#include "cli/CliConfig.h"
```
hbdeshmukh: This is not needed, right?

cramja: oops, should be storage/Cliconfig

hbdeshmukh:
LGTM. Thanks @cramja for spotting and fixing the bug.

cli/DefaultsConfigurator.hpp
```diff
@@ -51,6 +52,21 @@ class DefaultsConfigurator {
   }
 
   /**
+   * @brief Get the number of available numa sockets.
+   *
+   * @return Number of available numa sockets. Always 1 if the system doesn't
+   *         have libnuma.
+   **/
+  static std::size_t GetNumNUMANodes() {
+  #ifdef QUICKSTEP_HAVE_LIBNUMA
```
zuyu: @cramja For preprocessors, there is no indentations.

https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives

PR #207, opened 2016-05-04
Change default aggregate_hashtable_type from LinearOpenAddressing to SeparateChaining

jianqiao:
This PR changes the default hashtable for aggregation from `LinearOpenAddressingHashTable` to `SeparateChainingHashTable`.

The reason is that the performance of  `LinearOpenAddressingHashTable` drops radically if the future hash table size is underestimated on its initialization, while `SeparateChainingHashTable` is more resilient to any initial size. This situation gets reflected if we run TPC-H query 18.

The long-term solution will be to develop a more accurate estimation of the hash table size during query optimization.

### DIFF
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -135,7 +135,7 @@ static const volatile bool join_hashtable_type_dummy
     = gflags::RegisterFlagValidator(&FLAGS_join_hashtable_type,
                                     &ValidateHashTableImplTypeString);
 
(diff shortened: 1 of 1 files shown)


pateljm:
LGTM. Merging.

jianqiao:
The test failures are due to use of `SimpleScalarSeparateChainingHashTable` for `DISTINCT` aggregation that involve one group-by attribute -- but the _distinctify_ hash table should have two attributes as composite keys: one from group by and one is the argument. Will fix this problem and create a new PR.

PR #206, opened 2016-05-03
Support for NUMA aware preloading

hbdeshmukh:
Stored relations which have been NUMA-partitioned, can now be preloaded. The preloading can maintain the same NUMA placement of storage blocks, as specified while populating the relation.

PR #209, opened 2016-05-05
Cleanup in Preloader class

hbdeshmukh:
Added missing include and more informational log messages.

### DIFF
--- a/storage/PreloaderThread.cpp
+++ b/storage/PreloaderThread.cpp
@@ -77,7 +77,7 @@ void PreloaderThread::run() {
                 << " completely preloaded in buffer pool";
     }
   }
-  printf("Loaded %lu blocks", blocks_loaded);
+  printf(" Loaded %lu blocks ", blocks_loaded);
 }
 
 std::size_t PreloaderThread::preloadNUMAAware(
--- a/storage/PreloaderThread.cpp
+++ b/storage/PreloaderThread.cpp
@@ -17,6 +17,8 @@
 
 #include "storage/PreloaderThread.hpp"
 
+#include "storage/StorageConfig.h"
+
 #include <cstddef>
 #include <vector>
 
@@ -29,7 +31,6 @@
 #include "catalog/CatalogTypedefs.hpp"
 #include "storage/StorageBlock.hpp"
 #include "storage/StorageBlockInfo.hpp"
-#include "storage/StorageConfig.h"
 #include "storage/StorageManager.hpp"
 #include "threading/ThreadUtil.hpp"
 
@@ -59,7 +60,7 @@ void PreloaderThread::run() {
           BlockReference current_block =
               storage_manager_->getBlock(current_block_id, relation);
         } catch (...) {
-          LOG(ERROR) << "Error after loading " << blocks_loaded << "blocks\n";
+          LOG(ERROR) << "Error after loading " << blocks_loaded << "blocks";
           throw;
         }
         ++blocks_loaded;
@@ -73,7 +74,7 @@ void PreloaderThread::run() {
         }
       }
       LOG(INFO) << "Relation " << relation.getName()
-                << " completely preloaded in buffer pool\n";
+                << " completely preloaded in buffer pool";
     }
   }
   printf("Loaded %lu blocks", blocks_loaded);
@@ -106,7 +107,7 @@ std::size_t PreloaderThread::preloadNUMAAware(
         LOG(ERROR) << "Error while preloading: " << " After loading total "
                    << blocks_loaded + num_previously_loaded_blocks
                    << " blocks and " << blocks_loaded
-                   << " blocks of relation " << relation.getName() << "\n";
+                   << " blocks of relation " << relation.getName();
         throw;
       }
       ++blocks_loaded;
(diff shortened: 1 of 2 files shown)


hbdeshmukh:
Hi @pateljm 

Please review this PR. It has mostly simple changes. These changes can be potentially useful in cases when we decide not to preload the complete database (e.g. for experiments involving data sets larger than buffer pool), to check which relations have been preloaded to the buffer pool. 

A useful feature (TODO) will be to specify which relations can be preloaded to the buffer pool.

storage/PreloaderThread.cpp
```diff
@@ -29,6 +29,7 @@
 #include "catalog/CatalogTypedefs.hpp"
 #include "storage/StorageBlock.hpp"
 #include "storage/StorageBlockInfo.hpp"
+#include "storage/StorageConfig.h"
```
zuyu: We should include it before the first occurrence of `QUICKSTEP_HAVE_LIBNUMA`.

storage/PreloaderThread.cpp
```diff
@@ -55,22 +56,27 @@ void PreloaderThread::run() {
...
+          // The buffer pool has filled up. But, some database blocks are not
+          // loaded.
+          printf(
+              " The database is larger than the buffer pool. Only %lu blocks "
+              "were loaded", blocks_loaded);
           return;
         }
       }
+      LOG(INFO) << "Relation " << relation.getName()
+                << " completely preloaded in buffer pool\n";
```
zuyu: No need for `\n`, as `glog` prints one statement per line.

storage/PreloaderThread.cpp
```diff
@@ -55,22 +56,27 @@ void PreloaderThread::run() {
...
     }
   }
-  printf(" Loaded %lu blocks ", blocks_loaded);
+  printf("Loaded %lu blocks", blocks_loaded);
```
zuyu: Some whitespaces are intended to add.

https://github.com/pivotalsoftware/quickstep/pull/172#discussion_r60004795

storage/PreloaderThread.cpp
```diff
@@ -96,15 +104,17 @@ std::size_t PreloaderThread::preloadNUMAAware(
         BlockReference current_block = storage_manager_->getBlock(
             curr_block_id, relation, partition_numa_node_id);
       } catch (...) {
-        LOG(ERROR) << "Error after loading "
+        LOG(ERROR) << "Error while preloading: " << " After loading total "
```
zuyu: We could remove `" << "` to combine two strings.

storage/PreloaderThread.cpp
```diff
@@ -79,7 +86,8 @@ std::size_t PreloaderThread::preloadNUMAAware(
     const std::size_t num_slots) {
 #ifdef QUICKSTEP_HAVE_LIBNUMA
   std::size_t blocks_loaded = 0;
-  const NUMAPlacementScheme *placement_scheme = relation.getNUMAPlacementSchemePtr();
+  const NUMAPlacementScheme *placement_scheme =
+      relation.getNUMAPlacementSchemePtr();
```
zuyu: This is a late comment replied to #206.

I would suggest to define the method `preloadNUMAAware` under `QUICKSTEP_HAVE_LIBNUMA`.

To me, it does not make sense to build this method on a machine without `libnuma`.

pateljm: Good point @zuyu 
@hbdeshmukh  Want to roll that here or a separate PR?

hbdeshmukh: The intention was to be able to call the function regardless of the NUMA availability or not. If NUMA is not present the function body just becomes ``{return 0;}``, which is not really a big compilation overhead. 

Nonetheless, I have put the function in NUMA guards now and therefore the call site is also protected by NUMA guards.

hbdeshmukh:
Addressed all the comments.

storage/PreloaderThread.cpp
```diff
@@ -49,37 +49,46 @@ void PreloaderThread::run() {
   std::size_t blocks_loaded = 0;
 
   for (const CatalogRelation &relation : database_) {
-    if (relation.hasPartitionScheme()) {
+    if (relation.hasPartitionScheme() && relation.hasNUMAPlacementScheme()) {
```
jianqiao: It seems that when `QUICKSTEP_HAVE_LIBNUMA` is off, `CatalogRelation` will have no definition of `hasNUMAPlacementScheme()` and will incur compile error.

hbdeshmukh: Thanks for pointing out @jianqiao. I can remove the check in another PR. However I wonder, why the function hasNUMAPlacementScheme() should be unavailable when NUMA support is not present. It should just return false then. cc @adalbertgerald

jianqiao: Because in `CatalogRelation` (https://github.com/pivotalsoftware/quickstep/blob/master/catalog/CatalogRelation.hpp#L144), it is:
```
#ifdef QUICKSTEP_HAVE_LIBNUMA
  ...
  bool hasNUMAPlacementScheme() const {
    return placement_scheme_ != nullptr;
  }
  ...
#endif
```
Perhaps either move that `hasNUMAPlacementScheme()` definition out of the `#ifdef` block, or move this  `hasNUMAPlacementScheme()` call into the `#ifdef` block.

zuyu: I agree. Just remove `hasNUMAPlacementScheme` outside of the `QUICKSTEP_HAVE_LIBNUMA` scope.

zuyu: Checkout #215.

PR #210, opened 2016-05-05
Change default aggregate_hashtable_type from LinearOpenAddressing to SeparateChaining 

jianqiao:
This PR changes the default aggregation hash table implementation from LinearOpenAddressingHashTable to SeparateChainingHashTable.

The reason is that the performance of LinearOpenAddressingHashTable drops radically if the future hash table size is underestimated on its initialization, while SeparateChainingHashTable is more resilient to any initial size. This situation gets reflected if we run TPC-H query 18.

The long-term solution will be to develop a more accurate estimation of the hash table size during query optimization.

Meanwhile, this PR also fixes a bug for `DISTINCT` aggregation that occurs with the hash table impl type change. That is, the _distinctify_ hash tables (used for getting distinct arguments within each group) may need to have different implementation types from the _aggregation_ hash table. Typically, consider the following query:
```
SELECT SUM(DISTINCT x)
FROM a
GROUP BY y;
```
Here `x` and `y` have `int` types.
For this query, `ExecutionGenerator` will decide to use `SimpleScalarSeparateChainingHashTable` for the _aggregation_ hash table. But `SimpleScalarSeparateChainingHashTable` cannot be used for the _distinctify_ hash table because the keys to be distinctified are `{x, y}` which are composite.


storage/AggregationOperationState.proto
```diff
@@ -37,4 +39,7 @@ message AggregationOperationState {
   // NOTE(chasseur): 'hash_table_impl_type' is marked optional, but it is
   // needed if 'group_by_expressions' is non-empty, and ignored otherwise.
   optional HashTableImplType hash_table_impl_type = 6;
+
+  // Each DISTINCT aggregation has its distinctify hash table impl type.
+  repeated HashTableImplType distinctify_hash_table_impl_types = 7;
```
zuyu: I would suggest to add this field in `Aggregate` as following:

```
  message Aggregate {
    ...
    required bool is_distinct = 3;
    required HashTableImplType distinctify_hash_table_impl_type = 4;
  }
```

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -135,7 +135,7 @@ static const volatile bool join_hashtable_type_dummy
     = gflags::RegisterFlagValidator(&FLAGS_join_hashtable_type,
                                     &ValidateHashTableImplTypeString);
 
-DEFINE_string(aggregate_hashtable_type, "LinearOpenAddressing",
+DEFINE_string(aggregate_hashtable_type, "SeparateChaining",
```
zuyu: @jianqiao I understand that the performance of a particular query (Q18) is poor when using the default `LinearOpenAddressing`, but I was wondering if we do need to change the default value.

In general, `LinearOpenAddressing` would have a better performance over `SeparateChaining` version, as we avoid a look up in a long chain of hashed keys.

Alternatively, for some slow queries, we could just set a different hash table implementation to `SeparateChaining` instead of using the default option.

jianqiao: Currently changing from `LinearOpenAddressing` to `SeparateChaining` appears to improve the performance of all TPC-H queries that involve aggregation, but I'm not sure about SSB @pateljm.

pateljm:
LGTM. Merging.

PR #211, opened 2016-05-05
Add support for the substring function.

jianqiao:
This PR adds end-to-end support for the SUBSTRING function. Tests are also added.

TODO: In this PR, `SubstringOperation` is defined as a subclass of `UnaryOperation`. However it has two constant arguments `start_position` and `substring_length` so that it actually does not fit well into the singleton pattern as a `SubstringOperation` subclass. It stills remains to design a better long-term solution. For example, implement `ScalarTernaryOperation` or even `ScalarNaryOperation` class with specialization when some arguments are constants.


pateljm:
LGTM. Note about is good and agree that a longer term fix is needed. Merging. 

Thanks @jianqiao!

PR #213, opened 2016-05-07
Add print_query flag for printing input queries

jianqiao:
This PR adds a gflag `print_query` for printing each input query before printing the query's result. This is useful when we run a large number of queries in a batch. That is, we can know easily which query that each result is corresponding to.


pateljm:
Nice feature. Thanks @jianqiao!

PR #218, opened 2016-05-08
Leave 10% more memory for the OS.

pateljm:
The previous allocation is too aggressive when the database is far larger than the main memory -- in this case, the OS needs more memory to manage the IOs.


zuyu:
LGTM!

PR #219, opened 2016-05-10
Print number of rows in the output

hbdeshmukh:
- Provide an API to calculate the number of rows in a table.
- Print output time in milliseconds, upto 3 decimal places, similar to what Postgres does.
- Simplify output messages.

Sample output on a table with only one column and two rows: 
```
quickstep> select * from r;
+-----------+
|id         |
+-----------+
|          3|
|          5|
+-----------+
( 2 rows )
Time: 2.315 ms
quickstep> select count(*) from r;
+--------------------+
|count(*)            |
+--------------------+
|                   2|
+--------------------+
( 1 rows )
Time: 0.879 ms
quickstep> select count(*) from r where r.id < 4;
+--------------------+
|count(*)            |
+--------------------+
|                   1|
+--------------------+
( 1 rows )
Time: 0.879 ms
quickstep> select count(*) from r where r.id < 2;
+--------------------+
|count(*)            |
+--------------------+
|                   0|
+--------------------+
( 1 rows )
Time: 0.748 ms
```

I also tried running queries on larger tables and it worked correctly.

### DIFF
--- a/cli/PrintToScreen.cpp
+++ b/cli/PrintToScreen.cpp
@@ -162,9 +162,11 @@ std::size_t PrintToScreen::GetNumTuplesInRelation(
 void PrintToScreen::PrintOutputSize(const CatalogRelation &relation,
                                     StorageManager *storage_manager,
                                     FILE *out) {
+  const std::size_t num_rows = GetNumTuplesInRelation(relation, storage_manager);
   fprintf(out,
-          "( %lu rows )\n",
-          GetNumTuplesInRelation(relation, storage_manager));
+          "(%lu %s)\n",
+          num_rows,
+          (num_rows == 1) ? "row" : "rows");
 }
 
 }  // namespace quickstep
--- a/cli/PrintToScreen.cpp
+++ b/cli/PrintToScreen.cpp
@@ -29,6 +29,7 @@
 #include "storage/StorageManager.hpp"
 #include "storage/TupleIdSequence.hpp"
 #include "storage/TupleStorageSubBlock.hpp"
+#include "types/IntType.hpp"
 #include "types/Type.hpp"
 #include "types/TypedValue.hpp"
 #include "utility/Macros.hpp"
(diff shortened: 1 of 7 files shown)


hbdeshmukh:
Cancelled the travis build for another commit.

hbdeshmukh:
Minor change in the output formatting: 
```
quickstep> select count(*) from r;
+--------------------+
|count(*)            |
+--------------------+
|                   2|
+--------------------+
(1 row)
Time: 2.281 ms
quickstep> select * from r;
+-----------+
|id         |
+-----------+
|          3|
|          5|
+-----------+
(2 rows)
Time: 0.719 ms
```

cli/PrintToScreen.cpp
```diff
@@ -147,4 +148,25 @@ void PrintToScreen::printTuple(const TupleStorageSubBlock &tuple_store,
...
 
+std::size_t PrintToScreen::GetNumTuplesInRelation(
+    const CatalogRelation &relation, StorageManager *storage_manager) {
+  const std::vector<block_id> &blocks = relation.getBlocksSnapshot();
+  std::size_t total_num_tuples = 0;
+  for (block_id block : blocks) {
+    total_num_tuples +=
+        storage_manager->getBlock(block, relation)->getNumTuples();
+  }
+  return total_num_tuples;
```
zuyu: @hbdeshmukh This feature is great, but I suggest to merge this method as a part of `PrintToScreen::PrintRelation` around https://github.com/pivotalsoftware/quickstep/blob/master/cli/PrintToScreen.cpp#L94.

And thus the implementation of `PrintToScreen::PrintOutputSize` could just append to `PrintToScreen::PrintRelation` as the last line.

storage/StorageBlock.cpp
```diff
@@ -1361,4 +1361,9 @@ void StorageBlock::invalidateAllIndexes() {
   }
 }
 
+const std::size_t StorageBlock::getNumTuples() const {
```
zuyu: If we adopt my suggestions, we don't need this new method. We would have access to `tuple_store`.

cli/PrintToScreen.cpp
pateljm: Hi @zuyu: Good suggestions and one can go either way. In this case, it may be good to still print the number of rows when printing is disabled, just as we also print the time when printing is disabled.

pateljm:
LGTM. Merging.

cli/PrintToScreen.cpp
zuyu: @pateljm The new method suffers if we evict some blocks from the query result relation from the buffer pool. We have to load them back just to count the number of tuples again.

hbdeshmukh: @zuyu I had considered implementing printing the output size inside the ``PrintRelation()`` but didn't go ahead with the approach for following reasons: 

- The number of rows is not the part of the relation per se. 
- A lot of tests in optimizer and some in cli modules depend on the PrintRelation for correctness. Modifying all of them is a fair bit of work. 
- As @pateljm mentioned, separating these functions help to control whether we want to print both the output and its size, or just the size of the output. 
- This particular function can be reusable, e.g. for the describe table and describe database commands implemented by @rogersjeffreyl , we can print the number of tuples in the tables as a part of the output. 
-  The eviction issue you had mentioned is valid, but for the common case (e.g. TPC-H, SSB workloads) the query output size is typically small i.e. few number of blocks. So eviction won't be an issue in the common case. A permanent fix will be to cache the number of tuples for each relation, but that's for the future.

Issue #212, opened 2016-05-06
CCOUNT(*) bug


pateljm:
The following sequence of queries give surprising results: 
```
CREATE TABLE Foo (a INT, b INT);
CREATE TABLE Bar (a INT, b INT);
INSERT INTO Foo SELECT i, i%2 FROM generate_series(0, 1) AS gs(i);
INSERT INTO Bar SELECT i, i%2 FROM generate_series(0, 1) AS gs(i);

SELECT COUNT(a) FROM (SELECT Foo.a FROM Foo, Bar WHERE Foo.a = Bar.b) AS A(a);
SELECT COUNT(*) FROM (SELECT Foo.a FROM Foo, Bar WHERE Foo.a = Bar.b) AS A(a);
```
The second `SELECT` query returns 0 instead of 2.

hbdeshmukh:
I guess this is the same issue as [this](https://github.com/Pivotal-DataFabric/quickstep/issues/9)

pateljm:
@hbdeshmukh Yup -- but need it here in this repo too.

PR #220, opened 2016-05-10
Quick fix to the count(*) problem

jianqiao:
This PR is a quick fix to #212 that we retain at least one column from an `Aggregate`'s child plan to survive the `PruneColumns` physical optimization.

### DIFF
--- a/query_optimizer/physical/Aggregate.cpp
+++ b/query_optimizer/physical/Aggregate.cpp
@@ -1,6 +1,8 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
  *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


query_optimizer/physical/Aggregate.cpp
```diff
@@ -70,6 +72,18 @@ std::vector<E::AttributeReferencePtr> Aggregate::getReferencedAttributes()
                                  referenced_attributes_in_predicate.begin(),
                                  referenced_attributes_in_predicate.end());
   }
+
+  // TODO(jianqiao): This is a quick fix to the COUNT(*) problem that we retain
```
zuyu: We would emphasis `COUNT(*) on a temporary relation`.

pateljm:
Thanks @jianqiao for the quick fix and the TODO note.

query_optimizer/physical/Aggregate.cpp
```diff
@@ -70,6 +72,18 @@ std::vector<E::AttributeReferencePtr> Aggregate::getReferencedAttributes()
...
+  // at least one column from the child plan to survive the PruneColumns physical
+  // optimization. The comprehensive approach should also try to optimize the
+  // child plan with regard to the knowledge that the aggregation actually don't
+  // care about the values of the child plan's output but needs only the number
+  // of rows of the output.
+  if (referenced_attributes.size() == 0) {
+    DCHECK_GT(input_->getOutputAttributes().size(), static_cast<std::size_t>(0));
+    referenced_attributes.emplace_back(input_->getOutputAttributes()[0]);
```
zuyu: Question: what if we just create a dummy attribute for the `int` type?

pateljm: Good question @zuyu! Would it be possible to create a referenced_attribute that is simply an expression that evaluate to 1. If that is hard, let's just note it in the discussion here (no change required to the code), and worry about it in a separate PR.

jianqiao: The dummy attribute approach can work but we need to modify most of the `Physical` subclass implementations for `maybeCopyWithPrunedExpressions()`. There is a cascading effect on `Aggregate`'s descendants' output attributes. I can do this after finishing the join ordering PR.

pateljm: Good point. I think this will work for now. Merging!

PR #223, opened 2016-05-16
Fix bug in the SMA code

pateljm:
Fix bug in the SMA code so that the SMA predicate evaluation is only applied if at least one of the operands in the predicate is a static value. Thus, queries like: `select * from foo where foo.int1 < foo.int2;`


jianqiao:
LGTM. Merging.

PR #225, opened 2016-05-17
Quickstep gen stats

rogersjeffreyl:
This code adds an operator  for getting the number of rows in the relation. A new field is added to the CatalogRelation to store the tuple. This is currently invoked after the TextScanOperator. This will be used in the optimizer for cardinality estimates.

### DIFF
--- a/catalog/Catalog.proto
+++ b/catalog/Catalog.proto
@@ -1,7 +1,7 @@
 //   Copyright 2011-2015 Quickstep Technologies LLC.
 //   Copyright 2015-2016 Pivotal Software, Inc.
 //   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
-//    University of Wisconsin—Madison.
+//     University of Wisconsin—Madison.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -80,6 +80,16 @@ message IndexScheme {
   repeated IndexEntry index_entries = 1;
 }
 
+message CatalogRelationStatistics {
+  optional fixed64 num_tuples = 1;
+
+  message NumDistinctValuesEntry {
+    required int32 attr_id = 1;
+    required fixed64 num_distinct_values = 2;
+  }
+  repeated NumDistinctValuesEntry num_distinct_values_map = 2;
+}
+
 message CatalogRelationSchema {
   required int32 relation_id = 1;
   required string name = 2;
@@ -99,7 +109,7 @@ message CatalogRelation {
     optional IndexScheme index_scheme = 18;
     optional PartitionScheme partition_scheme = 19;
     optional NUMAPlacementScheme placement_scheme = 20;
-    optional uint32 num_tuples = 21 [default = 0];
+    optional CatalogRelationStatistics statistics = 21;
   }
 }
 
--- a/catalog/Catalog.proto
+++ b/catalog/Catalog.proto
@@ -99,7 +99,7 @@ message CatalogRelation {
     optional IndexScheme index_scheme = 18;
     optional PartitionScheme partition_scheme = 19;
     optional NUMAPlacementScheme placement_scheme = 20;
-    optional uint32 num_tuples = 21;
+    optional uint32 num_tuples = 21 [default = 0];
   }
 }
 
--- a/catalog/Catalog.proto
+++ b/catalog/Catalog.proto
@@ -99,6 +99,7 @@ message CatalogRelation {
     optional IndexScheme index_scheme = 18;
     optional PartitionScheme partition_scheme = 19;
     optional NUMAPlacementScheme placement_scheme = 20;
+    optional uint32 num_tuples = 21;
   }
 }
 
--- a/catalog/CatalogRelation.cpp
+++ b/catalog/CatalogRelation.cpp
@@ -132,6 +132,7 @@ CatalogRelation::CatalogRelation(const serialization::CatalogRelationSchema &pro
   }
 
   default_layout_.reset(new StorageBlockLayout(*this, proto_default_layout));
+  num_tuples_ = proto.GetExtension(serialization::CatalogRelation::num_tuples);
 }
 
 serialization::CatalogRelationSchema CatalogRelation::getProto() const {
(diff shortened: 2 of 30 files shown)


pateljm:
Nice feature @rogersjeffreyl ! Thanks.

pateljm:
Quick question: Looks like this runs the new operator (which is nice to have to run to gather stats). Would it be possible to not re-scan the data (which this seems to be doing), and to have the TextScanOperator return the tuples loaded, which can then be used to increment the current catalog stats. 

We can simply make a note of this and do a separate PR later. This is a nice addition! 

@jianqiao @hbdeshmukh Comments?

hbdeshmukh:
@pateljm That's a good point. Alternative could be to introduce the new operator in between TextScan and SaveBlocks (which is the operator chain before this PR), so that stats on each block can be computed via pipelined blocks from TextScan to GetStats operator, which subsequently get streamed to SaveBlocks and saved to disk.

catalog/CatalogRelation.cpp
```diff
@@ -180,6 +180,10 @@ serialization::CatalogRelationSchema CatalogRelation::getProto() const {
   return proto;
 }
 
+void CatalogRelation::setNumTuples(std::size_t num_tuples) const {
```
jianqiao: Remove this `const`.

catalog/CatalogRelation.hpp
```diff
@@ -190,6 +191,22 @@ class CatalogRelation : public CatalogRelationSchema {
...
+  const std::size_t getNumTuples() const {
+    return num_tuples_;
+  }
+
+  /**
+   * @brief Set the number of tuples for catalog relation.
+   *
+   * @param num_tuples The number of tuples in the catalog relation.
+   **/
+  void setNumTuples(std::size_t num_tuples) const;
```
jianqiao: Remove `const` since this method is actually updating the state of `CatalogRelation`.

catalog/CatalogRelation.hpp
```diff
@@ -397,6 +414,8 @@ class CatalogRelation : public CatalogRelationSchema {
   // Mutex for locking the index scheme.
   alignas(kCacheLineBytes) mutable SpinSharedMutex<false> index_scheme_mutex_;
 
+  mutable std::size_t num_tuples_;
```
jianqiao: Remove `mutable` here.
The combination of a `mutable` field and its related `const` method are for some special situations, e.g. when the `mutable` field is an auxiliary state (mutex, value cache) and the related method _logically_ does not change the state of the object.

This StackOverflow thread gives more discussion: http://stackoverflow.com/questions/105014/does-the-mutable-keyword-have-any-purpose-other-than-allowing-the-variable-to

catalog/CatalogRelation.hpp
```diff
@@ -80,6 +80,7 @@ class CatalogRelation : public CatalogRelationSchema {
                   bool temporary = false)
       : CatalogRelationSchema(parent, name, id, temporary),
         default_layout_(nullptr) {
+        num_tuples_ = 0;
```
jianqiao: Move the initialization of `num_tuples_` into the member initializer list, i.e. 
```
      : CatalogRelationSchema(parent, name, id, temporary),
        default_layout_(nullptr),
        num_tuples_(0) {
  }
```

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -947,6 +948,14 @@ void ExecutionGenerator::convertCopyFrom(
   execution_plan_->addDirectDependency(save_blocks_operator_index,
                                        scan_operator_index,
                                        false /* is_pipeline_breaker */);
+
+  const QueryPlan::DAGNodeIndex num_rows_operator_index =
+      execution_plan_->addRelationalOperator(
+          new GenerateNumRowsStatsOperator( output_relation));
```
jianqiao: Pass as parameter the mutable pointer to the output relation, i.e.
```
optimizer_context_->catalog_database()->getRelationByIdMutable(output_relation->getID());
```

relational_operators/GenerateNumRowsStatsOperator.hpp
```diff
@@ -0,0 +1,79 @@
...
+class GenerateNumRowsStatsOperator : public RelationalOperator {
+ public:
+  /**
+   * @brief Constructor.
+   *
+   * @param relation The relation to get the number of rows from.
+   *                 This GenNumRowStatsOperator owns relation until
+   *                 the WorkOrder it produces is successfully executed.
+   **/
+  explicit GenerateNumRowsStatsOperator(const CatalogRelation *relation)
```
jianqiao: Remove `const` since this operator will update the state of `relation`.

relational_operators/GenerateNumRowsStatsOperator.hpp
```diff
@@ -0,0 +1,79 @@
...
+   * @note no WorkOrder is generated for this operator.
+   **/
+  bool getAllWorkOrders(WorkOrdersContainer *container,
+                        QueryContext *query_context,
+                        StorageManager *storage_manager,
+                        const tmb::client_id scheduler_client_id,
+                        tmb::MessageBus *bus) override;
+
+ private:
+  const CatalogRelation *relation_;
```
jianqiao: Remove `const`.

catalog/Catalog.proto
```diff
@@ -99,6 +99,7 @@ message CatalogRelation {
     optional IndexScheme index_scheme = 18;
     optional PartitionScheme partition_scheme = 19;
     optional NUMAPlacementScheme placement_scheme = 20;
+    optional uint32 num_tuples = 21;
```
jianqiao: Also needs to update `CatalogRelation::getProto()` and `CatalogRelation(const serialization::CatalogRelationSchema &proto)` to serialize / deserialize this field.

relational_operators/GenerateNumRowsStatsOperator.cpp
```diff
@@ -0,0 +1,42 @@
...
+namespace quickstep {
+
+bool GenerateNumRowsStatsOperator::getAllWorkOrders(
+    WorkOrdersContainer *container,
+    QueryContext *query_context,
+    StorageManager *storage_manager,
+    const tmb::client_id scheduler_client_id,
+    tmb::MessageBus *bus) {
+  std::size_t num_tuples =
+      PrintToScreen::GetNumTuplesInRelation(*relation_, storage_manager);
```
jianqiao: Better not have this dependency from a `RelationalOperator` to a static method of `PrintToScreen` which is from the cli module.

One simple solution is to just re-implement the logic here (i.e. that small `for` loop in `PrintToScreen::GetNumTuplesInRelation()`).

jianqiao:
LGTM. Merging.

PR #226, opened 2016-05-18
Change the default value of Joined Tuple Collector

pateljm:
On NUMA boxes, the vector based joined tuple collector slows down queries.


zuyu:
LGTM.

PR #227, opened 2016-05-19
Initial support for collecting table statistics: number of distinct values

jianqiao:
This PR adds an initial support for collecting table statistics for cost-based query optimization. Specifically, it implements a command `\analyze` that counts the number of distinct values for each column of each relation. This information will be used to estimate predicate's selectivity in a following PR that performs join order optimization.

TODO: will add unit tests later for the `\analyze` feature.


pateljm:
LGTM. Merging.

PR #229, opened 2016-05-19
Add hash join order optimization for star schema queries.

jianqiao:
This PR implements a cost-based optimizer for ordering hash joins. The optimizer uses a greedy algorithm that is suitable for queries on star-schema tables (e.g. the SSB benchmark).

**Note:** To obtain good query plans, run `\analyze` command once after loading (i.e. COPY) the tables to gather statistics of the tables.

TODO: more comments may be added to explain the optimization algorithm.


pateljm:
Super! LGTM. Merging.

Issue #228, opened 2016-05-19
Non-deterministic failure of test case 82 (SMAIndexSubBlock_unittest) in Build #829


jianqiao:
https://travis-ci.org/pivotalsoftware/quickstep/builds/131457246

The failure is non-deterministic because the test can be passed by restarting the test job.

pateljm:
Fixed in PR#230. Closing this issue.

PR #230, opened 2016-05-19
Initialize the allocated memory bytes to 0 in ScopedBuffer

jianqiao:
This PR initializes by default the newly allocated memory bytes to 0 in `ScopedBuffer`. This will fix the nondeterministic `free(): invalid pointer: 0x00007fcbe4877458` error as addressed in issue #228. The reason is that in the original code `SMAIndexSubBlockTest::createIndex()` passed an uninitialized memory block allocated from `ScopedBuffer` to `SMAIndexSubBlock` which was then mis-interpreted.


pateljm:
LGTM. Merging.

PR #234, opened 2016-05-22
Resolves SMA bugs revealed by TPCH.

cramja:
Running the TPCH queries revealed some bugs in SMA.
**Bugfix 1**
This resolves the issue where predicates which are not of the form 'Attribute comparison literal' (Ex. `lo_orderkey > 100000`) cause a GLOG(FATAL) when really, they can just be ignored by the SMA.

The commit `eb6fecd` resolves this.

**Bugfix 2**
There is another issue with SMA to do with conversion of TypedValues. For example on a simplified version of TPCH query 6:
```SQL
SELECT sum(l_extendedprice * l_discount) AS revenue 
FROM lineitem
WHERE l_quantity < 24;
```
The literal value `24` is integer type compared to the double type which l_quantity is stored as.

The commit `824efaa` fixes this.

### DIFF
--- a/storage/SMAIndexSubBlock.cpp
+++ b/storage/SMAIndexSubBlock.cpp
@@ -181,19 +181,22 @@ Selectivity getSelectivity(const TypedValue &literal,
 }
 
 SMAPredicate* SMAPredicate::ExtractSMAPredicate(const ComparisonPredicate &predicate) {
-  if (predicate.getLeftOperand().hasStaticValue()) {
-    DLOG_IF(FATAL, predicate.getRightOperand().getDataSource() != Scalar::kAttribute);
+  // Check that the predicate is contains an attribute and a literal value.
+  if (predicate.getLeftOperand().hasStaticValue() &&
+      predicate.getRightOperand().getDataSource() == Scalar::kAttribute) {
     return new SMAPredicate(
          static_cast<const ScalarAttribute&>(predicate.getRightOperand()).getAttribute().getID(),
          flipComparisonID(predicate.getComparison().getComparisonID()),
          predicate.getLeftOperand().getStaticValue().makeReferenceToThis());
-  } else {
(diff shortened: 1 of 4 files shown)


d:
Instead of aborting the process we now return a `Selectivity::kUnknown`. Was the code under test? Are we seeing a missing test for the new behavior now?

cramja:
@d The previous code was not under test in as far as the unit test does not cover the method `SMAIndexSubBlock::getSelectivityForPredicate()` though the helper methods inside of it were tested. 

This PR will include a test for `getSelectivityForPredicate` once finished, as the type-coercion commit will need a unit test to cover it and in that same unit test I'll add tests for the `kUnknown` case.

d:
ah thx for the clarification marc

zuyu:
@cramja I was wondering if we could upcast `int` to `double` when doing such comparisons. I believe we support so in `TypedValue` comparisons.

pateljm:
LGTM. Thanks @marc! Merging.

storage/SMAIndexSubBlock.cpp
```diff
@@ -617,6 +617,36 @@ Selectivity SMAIndexSubBlock::getSelectivityForPredicate(const ComparisonPredica
...
+        TypeFactory::GetType(sma_predicate->literal.getTypeID(), static_cast<std::size_t>(literal_length), false);
+    const Type &attribute_type = literal_length == -1 ?
+        TypeFactory::GetType(entry->type_id, false) :
+        TypeFactory::GetType(entry->type_id, static_cast<std::size_t>(literal_length), false);
+    if (attribute_type.isSafelyCoercibleFrom(literal_type)) {
+      // Convert the literal's type inside the predicate.
+      SMAPredicate *replacement = new SMAPredicate(
+          sma_predicate->attribute,
+          sma_predicate->comparison,
+          attribute_type.coerceValue(sma_predicate->literal, literal_type));
```
cramja: @zuyu The call to `coerce` should do upcasts safely.

PR #235, opened 2016-05-23
Explicitly specify where tcmalloc comes from

d:
This change adds a declaration that `libtcmalloc_minimal.a` is generated
by the external project. This is needed by the Ninja generator to
properly build.

To reproduce:
run `cmake -GNinja path/to/quickstep`
then `cmake --build path/to/build/dir` and see it fails.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -490,6 +490,7 @@ if(USE_TCMALLOC)
         CXXFLAGS=${THIRD_PARTY_CXX_FLAGS}
     BUILD_COMMAND make
     BUILD_IN_SOURCE 0
+    BUILD_BYPRODUCTS <INSTALL_DIR>/lib/libtcmalloc_minimal.a
   )
   # Static libtcmalloc_minimal.a
   add_library(libtcmalloc_minimal STATIC IMPORTED)


CMakeLists.txt
```diff
@@ -490,6 +490,7 @@ if(USE_TCMALLOC)
         CXXFLAGS=${THIRD_PARTY_CXX_FLAGS}
     BUILD_COMMAND make
     BUILD_IN_SOURCE 0
+    BUILD_BYPRODUCTS <INSTALL_DIR>/lib/libtcmalloc_minimal.a
```
pateljm: Quick question: Does INSTALL_DIR refer to where libtcmalloc_minimal.a is installed? If so can we name this variable TCM_INSTALL_DIR? 

Thanks @d for getting involved!

d: Yes and no.
Yes `<INSTALL_DIR>` refers to the value of `PREFIX`. No we don't have the liberty to rename it, it's a special string that gets replaced by `ExternalProject` module

pateljm: Thanks @d. LGTM. Merging.

pateljm:
@d. Ok -- I'm rethinking this PR as it builds now in the third_party/gperftools rather than under the build directory. Would you be ok if we revered this PR, by commenting out the change, so that if you need you can build. I think the common case is now broken.

d:
Yes @pateljm please revert if the change blocked everybody else or broke the build. But can you show me a travis link where it was broken?

d:
Just came back from manual testing, and I can confirm that the Makefile generator will **not** pollute your source tree with `tcmalloc` static library files

PR #236, opened 2016-05-23
Groupby hashtable pool

hbdeshmukh:
This PR creates a pooling mechanism for group by hash tables. Earlier, all the worker threads involved in the aggregation phase shared a hash table per aggregation handle. This caused degradation in the performance when the number of groups in the output are low. 

In the new mechanism, each thread can work on a private hash table and we finally merge the individual hash tables in the pool. 

The performance is pretty much improved for all the SSB queries and for queries which don't see any improvement, the degradation is < 10%. TPC-H Q1 SF100 saw a 4x performance improvement with these changes (based on @pateljm 's evaluation for TPC-H).

### DIFF
--- a/expressions/aggregation/AggregationConcreteHandle.cpp
+++ b/expressions/aggregation/AggregationConcreteHandle.cpp
@@ -68,10 +68,4 @@ void AggregationConcreteHandle::insertValueAccessorIntoDistinctifyHashTable(
   }
 }
 
-/*void AggregationConcreteHandle::mergeHashTables(
-    const AggregationStateHashTableBase &source_hash_table,
-    AggregationStateHashTableBase *destination_hash_table) {
-
-}*/
-
 }  // namespace quickstep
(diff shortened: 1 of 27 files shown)


pateljm:
We should in a separate PR allow the optimizer to turn on/off this option. Nice work @hbdeshmukh!

I'd personally love to see a few more comments around the `HashTableMerger` class (ok -- you knew I was going to say that :-) But, ok to do that in a separate PR. 

It it looks good to @jianqiao, we should close.

hbdeshmukh:
Killed the travis build, as GCC didn't like one of the ``DCHECK``s.

expressions/aggregation/AggregationConcreteHandle.cpp
```diff
@@ -68,4 +68,10 @@ void AggregationConcreteHandle::insertValueAccessorIntoDistinctifyHashTable(
   }
 }
 
+/*void AggregationConcreteHandle::mergeHashTables(
+    const AggregationStateHashTableBase &source_hash_table,
+    AggregationStateHashTableBase *destination_hash_table) {
+
+}*/
+
```
hbdeshmukh: Remove the dead code.

hbdeshmukh:
@jianqiao Based on our discussion, I am merging this PR as some of us are waiting for the merge in order to run the experiments. Thanks for your review comments!

jianqiao:
LGTM! Sorry that I was out for a while and didn't merge in time..

